Issue #464, opened 2014-12-21
Refactor towards ROP


theimowski:
What do you guys think about employing the [Railway Oriented Programming](http://fsharpforfunandprofit.com/rop/) in Paket codebase?

One obvious benefit would be that in case of any failure, no file would get modified. (yeah I know one could simply ``git reset``, but that is probably not the best UX)

A command-line tool seems like a good candidate for the ROP technique.
As a volunteer I could do an experiment with the commands I'm most familiar with, i.e. ``convert-from-nuget`` and ``simplify`` to see where we can get to.

Edit:
List of paket commands that use ROP:
- [x] Init
- [ ] Add
- [ ] Remove
- [ ] Install
- [ ] Restore
- [ ] Update
- [x] Outdated
- [x] ConvertFromNuget
- [x] InitAutoRestore
- [x] Simplify
- [x] FindRefs
- [x] Config

forki:
are you going to refactor other commands as well?

theimowski:
I plan to, but surely it can take more time. 
Applying ROP gradually (i.e. not in a single PR) to the commands should be less error-prone

forki:
:+1

forki:
I wonder what Scott Wlashin is thinking about this!? It's becoming probably the biggest OSS project on ROP.

theimowski:
yeah, it'd be great if he could review the stuff one day when more commands use ROP, and tell whether that's what he meant :+1:

PR #476, opened 2014-12-27
WIP - Rop experiment

theimowski:
#464 

Submitting to get some feedback. 
I've applied (at least tried to) Railway Oriented Programming to two commands: Convert from nuget and Simplify. The results / errors of a command are handled by call-site in PublicApi.
Pros
 - I find the code a bit easier to reason about 
 - No modification in file system is performed until the command returns success result

Cons
 - code overhead for the infrastructure
 - in order to fully benefit from the technique, probably all commands ought to be transformed to use it

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -56,47 +56,67 @@ type Dependencies(dependenciesFileName: string) =
         Dependencies(dependenciesFileName)
      
     /// Converts the solution from NuGet to Paket.
-    static member ConvertFromNugetRRRRR(force: bool,installAfter: bool,initAutoRestore: bool,credsMigrationMode: string option) : unit =
+    static member ConvertFromNuget(force: bool,installAfter: bool,initAutoRestore: bool,credsMigrationMode: string option) : unit =
         let dependencies = 
             try Some <| Dependencies.Locate()
             with _ -> None 
         
-        let existingDependenciesFile = 
-            dependencies |> Option.map (fun d -> d.DependenciesFile)
-        
         let dependenciesFileName = 
-            match existingDependenciesFile with
-            | Some file -> file
+            match dependencies  with
+            | Some dependencies -> dependencies.DependenciesFile
             | None -> Path.Combine(Environment.CurrentDirectory, Constants.DependenciesFileName)
         
         let result = 
             Utils.RunInLockedAccessMode(
                 Path.GetDirectoryName(dependenciesFileName),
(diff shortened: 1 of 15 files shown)


theimowski:
There are some new changes here.
I introduced inline operators for ``Rop`` module - hopefully they're not to mystifying.
There's the new concept of ``PaketEnv`` - hopefully this can be used by other commands in future as well
@forki do you think we can merge it, or shall I do some more cleanups?

forki:
balls

src/Paket.Core/PublicAPI.fs
```diff
@@ -53,27 +55,38 @@ type Dependencies(dependenciesFileName: string) =
...
+    static member Simplify(interactive : bool) =
+        match PaketEnv.locatePaketRootDirectory(DirectoryInfo(Environment.CurrentDirectory)) with
+        | Some rootDirectory ->
+            Utils.RunInLockedAccessMode(
+                rootDirectory.FullName,
+                fun () -> 
+                    PaketEnv.fromRootDirectory rootDirectory
+                    >>= Simplifier.ensureNotInStrictMode
+                    >>= Simplifier.simplify interactive
+                    |> either (Simplifier.updateEnvironment) (List.iter (string >> Logging.traceError))
```
forki: so this doesn't fail anymore? The exit code will not be set to -1?

theimowski: yeah, my bad. There can be multiple failures now, so maybe we should raise exn with concatenated failures?

forki: I created a naive version in order to allow easier transition for semi-converted functions. see https://github.com/fsprojects/Paket/commit/ef01f31a4e53120eca7f4a12b55c62f6f14b85b6#diff-e4187eca86b045ab98ddafa4e6c44075R14

theimowski: OK, then for now we can do

    |> returnOrFail |> Simplifier.updateEnvironment

instead of 

    |> either (Simplifier.updateEnvironment) (List.iter (string >> Logging.traceError))

forki: yep. maybe you could do this with your next PR ;-)

2015-01-06 11:13 GMT+01:00 Tomasz Heimowski <notifications@github.com>:

theimowski: Will do, hopefully tonight - also will prepere a list of commands with status of what needs to be still converted to ROP.
Prefer project milestone or a plain list in an issue?

forki: Cool.
Use whatever works best for you.
I think plain issue is probably enough.

Issue #386, opened 2014-11-23
Implement "paket add url http://..." for github dependencies


forki:
We should a github/url version for paket add.

PR #511, opened 2015-01-04
WIP - paket add support for HTTP, gist and github dependencies

kerams:
- [X] `paket add url` (as per #386)
- [ ] `paket add gist`
- [ ] `paket add github`
- [ ] web documentation

It would be nice if someone went through the commits to make sure I am not messing up something, because I'm new to Paket's codebase.

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -49,6 +49,7 @@ type CLIArguments =
     | [<AltCommandLine("-f")>] Force
     | Hard
     | [<CustomCommandLine("nuget")>] Nuget of string
+    | [<CustomCommandLine("url")>] Url of string * string
     | [<CustomCommandLine("version")>] Version of string
     | [<CustomCommandLine("add-credentials")>] AddCredentials of string
     | [<Rest>]References_Files of string
@@ -84,6 +85,7 @@ with
             | Include_Prereleases -> "includes prereleases when searching for outdated packages."
             | No_Auto_Restore -> "omits init-auto-restore after convert-from-nuget."
             | Nuget _ -> "allows to specify a nuget package."
+            | Url (_, _) -> "allows to specify an HTTP resource."
             | Version _ -> "allows to specify a package version."
             | Creds_Migration _ -> "allows to specify credentials migration mode for convert-from-nuget."
             | Log_File _ -> "allows to specify a log file."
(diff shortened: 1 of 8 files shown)


src/Paket/Program.fs
```diff
@@ -156,14 +162,18 @@ try
...
-                Dependencies.Locate().Add(packageName, version, force, hard, interactive, noInstall |> not)
+            | Command.Add ->
+                match results.TryGetResult <@ CLIArguments.Nuget @> with
+                | Some packageName ->
+                    let version = match results.TryGetResult <@ CLIArguments.Version @> with
+                                  | Some x -> x
+                                  | _ -> ""
+
+                    Dependencies.Locate().Add(packageName, version, force, hard, interactive, noInstall |> not)
+                | None -> match results.TryGetResult <@ CLIArguments.Url @> with
```
kerams: I am not sure this is the right way to branch on paket add nuget/http/gist/github -- I'd end up with 4 nested matches. Any tips?

forki: I think that's ok for now. We can refactor later (and put this into the Add module)

kerams:
I've edited `paket-add.md` to update the docs, but it seems the file is in .gitignore. What gives?

forki:
Oh you need to look into the paket console project. The help texts are
reused and generated into the md file

forki:
https://github.com/fsprojects/Paket/blob/master/src/Paket/HelpTexts.fs#L170

src/Paket/Program.fs
```diff
@@ -49,6 +49,9 @@ type CLIArguments =
     | [<AltCommandLine("-f")>] Force
     | Hard
     | [<CustomCommandLine("nuget")>] Nuget of string
+//    | [<CustomCommandLine("gist")>] Gist of string * string
+//    | [<CustomCommandLine("github")>] Github of string * string
+    | [<CustomCommandLine("url")>] Url of string * string
     | [<CustomCommandLine("version")>] Version of string
     | [<CustomCommandLine("add-credentials")>] AddCredentials of string
```
kerams: I'm gonna need help with this one, guys. As you know, it is possible to reference a single file on Github or an entire repo. Similarly, a single gist file or a multi-file gist. This means that I'd ideally want to use something like `[<CustomCommandLine("gist")>] Gist of string * string option` for parsing `paket add gist`, but I'm getting an exception that Option types are not supported by the UnionArgParser. Any idea how to tackle this problem?

kerams: | [<CustomCommandLine("gist")>] MultiGist of string * string
    | [<CustomCommandLine("gist")>] SingleGist of string
Does not work either. `paket add gist http://..` works, but if I provide a file name, it throws saying that the parameter is not recognized. The parser seems to work with SingleGist all the time. Don't know that else to try. @forki could you look into this when you have a moment?

forki: /cc @ eiriktsarpalis any ideas? thanks

kerams: You made a little typo, fixing ;)
@eiriktsarpalis

forki: I'm not sure, but maybe it will be easier after https://github.com/fsprojects/Paket/pull/551 - can you try this?

forki: You might also want to look at https://github.com/fsprojects/Paket/blob/master/src/Paket/Program.fs#L45 and https://github.com/fsprojects/Paket/blob/master/src/Paket/Program.fs#L189

with this trick you don't get string * string option but a list of strings. That might be enough. You could then just throw if you have more than 2 strings

eiriktsarpalis: @kerams @forki Just came across this mention, apologies for the delay.

Separate parameters need to have separate CLI identifiers. The fact that the library selects an arbitrary parameter rather than failing at initialization is a bug. You should rename either of the two parameter identifiers.

It would be interesting to see if optional parameters could be supported in a reasonable manner. Why not make an issue under [UnionArgParser](https://github.com/nessos/UnionArgParser)?

kerams: Eirik, I've created the said issue.

Forki, I am not sure if using more than one `[<Rest>]` is possible. Maybe I don't fully understand that attribute, but I don't see how using 2 of them would make sense.
I'll take a look at what's being done in #551 tomorrow.

kerams:
Been a little busy lately, but will hopefully get back to this the next week.

theimowski:
there've been some breaking changes with command args - let me know when you need some help

kerams:
Gonna close this since the main repo has changed significantly and start over.

PR #520, opened 2015-01-06
Fixes in Convert-from-nuget after ROP

forki:


theimowski:
looks good, I forgot we had the ``FindReferencesFile`` function

forki:
also it seems https://github.com/fsprojects/Paket/pull/520/files#diff-8ae98470383c2c7af865300976ee41feR263 got lost.

theimowski:
yeah, I assumed the default source should be in the APPDATA nuget.config; but that assumption may not always be correct

PR #522, opened 2015-01-06
Raise exception when a train arrives on Failure rails - references #464

theimowski:


src/Paket.Core/Utils.fs
```diff
@@ -230,7 +230,7 @@ let RunInLockedAccessMode(rootFolder,action) =
     with
     | exn ->
             releaseLock()
-            raise exn
+            reraise()
```
theimowski: ``raise exn`` resulted in losing the stacktrace, hence verbose mode in case of exception always showed a newly generated stacktrace

PR #524, opened 2015-01-07
Rop builder

mexx:
I added a simple builder for ROP and converted some simple functions in `Simpilier` module to use the builder.

@forki @theimowski Please review
related to #464

### DIFF
--- a/src/Paket.Core/Simplifier.fs
+++ b/src/Paket.Core/Simplifier.fs
@@ -34,32 +34,28 @@ let removePackage(packageName, indirectPackages, fileName, interactive) =
     else
         false
 
-let simplifyDependenciesFile (dependenciesFile : DependenciesFile, flatLookup, interactive) =
-    let create (d : DependenciesFile) indirect =
-        let newPackages = 
-            dependenciesFile.Packages
-            |> List.filter (fun package -> not <| removePackage(package.Name, indirect, dependenciesFile.FileName, interactive))
-        DependenciesFile(d.FileName, d.Options, d.Sources, newPackages, d.RemoteFiles)
-
+let simplifyDependenciesFile (dependenciesFile : DependenciesFile, flatLookup, interactive) = rop {
     let packages = dependenciesFile.Packages |> List.map (fun p -> p.Name)
-    let indirect = findIndirect(packages, flatLookup, DependencyNotFoundInLockFile)
-        
-    create dependenciesFile
-    <!> indirect
+    let! indirect = findIndirect(packages, flatLookup, DependencyNotFoundInLockFile)
 
-let simplifyReferencesFile (refFile, flatLookup, interactive) =
-    let create refFile indirect =
-        let newPackages = 
-            refFile.NugetPackages 
-            |> List.filter (fun p -> not <| removePackage(p, indirect, refFile.FileName, interactive))
-        { refFile with NugetPackages = newPackages }
+    let newPackages = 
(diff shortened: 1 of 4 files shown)


src/Paket.Core/Simplifier.fs
```diff
@@ -34,60 +34,50 @@ let removePackage(packageName, indirectPackages, fileName, interactive) =
...
-            referencesFiles
-            |> List.map (fun refFile -> simplifyReferencesFile(refFile, flatLookup, interactive))
-            |> Rop.collect
+    let referencesFiles' =
+        referencesFiles
+        |> List.map (fun refFile -> simplifyReferencesFile(refFile, flatLookup, interactive))
+        |> Rop.collect
 
-        let projects = List.zip projectFiles <!> referencesFiles'
+    let! projects = List.zip projectFiles <!> referencesFiles'
```
theimowski: ``referencesFiles'`` could be also bound with ``let!``, so that the ``<!>`` operator is not necessary here

theimowski:
Looks good to me
Is it a good practice (in general) to maintain both custom operators and a custom builder for the same concept?

src/Paket.Core/Simplifier.fs
mexx: you're right, I didn't notice that one.

mexx:
I would go only for builder as for me it's more convenient to read the code.

theimowski:
Ok I think I can live with only builder, @forki U have a opinion?

forki:
then we should opt for builder...

2015-01-07 11:40 GMT+01:00 Tomasz Heimowski <notifications@github.com>:

theimowski:
Ok I'll switch to the builder with new PRs

isaacabraham:
I think we should have both a builder and custom operators options out there - oftentimes the builder will be more useful but occasionally e.g. if you just have one or two items in the pipeline - then operators are useful.

vasily-kirichenko:
Why have you guys invented this "rop" thing?.. Isn't it the Either monad which is already implemented in FSharpx and ExtCore (+ lots of combinators and combined monads like AsyncChoice, ProtectedState etc.)?

mexx:
@vasily-kirichenko You're right, it's the Either monad.
For me it has some learn effect, as I can dive into creating own computational expression builders.
Nevertheless we could replace the own implementation by referencing either of given alternatives.
@forki and @theimowski  Thoughts?

forki:
I'm pretty happy with "domain specific" wording. Reusing either might be a
good idea, but only as implementation detail.

theimowski:
I think the key point here is that we apply the technique, and whether we use custom or depend on another infrastructure is not that important

mexx:
@theimowski well spoken

vasily-kirichenko:
Two things: 1. code reuse 2. well known monads/names.

isaacabraham:
@vasily-kirichenko makes a good point. To be fair I went through the pain of creating my own Either monad a while back, it was a good learning experience, but I think it might be worth using a tried-and-tested one that will be more "standard".

forki:
The point of Rop is that it's more specific than Either monad. Of course it's an instance of an Either (and we can reuse implementation), but it gives meaning to really abstract Choice1Of2 and Choice2of2. It also comes with a nice metaphor which can make it easier to grasp the concept (we will see if this is true). 
So in one sense it's not the same as either - it's ab application of either.

PR #525, opened 2015-01-07
Rop builder

mexx:
continues work of #524, related to #464

### DIFF
--- a/src/Paket.Core/Environment.fs
+++ b/src/Paket.Core/Environment.fs
@@ -63,6 +63,18 @@ module PaketEnv =
             |> Seq.tryFind File.Exists
             |> Option.map (fun f -> DirectoryInfo(Path.GetDirectoryName(f)))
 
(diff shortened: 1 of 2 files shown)


src/Paket.Core/Environment.fs
```diff
@@ -63,6 +63,18 @@ module PaketEnv =
             |> Seq.tryFind File.Exists
             |> Option.map (fun f -> DirectoryInfo(Path.GetDirectoryName(f)))
 
+    let ensureNotExists (directory : DirectoryInfo) =
+        match fromRootDirectory directory with
```
mexx: I only would check that directory is present and dependencies file is not.
Actually for NuGet conversion you even set the lock file to `None` programmatically.

theimowski: I'm ok with that, but then maybe move back the ``ensureNotExists`` to use for nugetConvert module only

PR #526, opened 2015-01-07
Rop builder

mexx:
continues work of #524, related to #464


forki:
it's so cool seeing this taking shape

insideout

PR #529, opened 2015-01-07
Convert FindRefs command to ROP

mexx:
related to #464

### DIFF
--- a/src/Paket.Core/LockFile.fs
+++ b/src/Paket.Core/LockFile.fs
@@ -284,6 +284,27 @@ module LockFileParser =
 /// Allows to parse and analyze paket.lock files.
 type LockFile(fileName:string,options,resolution:PackageResolution,remoteFiles:ResolvedSourceFile list) =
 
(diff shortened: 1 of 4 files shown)


forki:
LGTM

src/Paket.Core/LockFile.fs
```diff
@@ -284,6 +284,27 @@ module LockFileParser =
 /// Allows to parse and analyze paket.lock files.
 type LockFile(fileName:string,options,resolution:PackageResolution,remoteFiles:ResolvedSourceFile list) =
 
+    let dependenciesByPackageLazy = lazy (
```
theimowski: Why lazy? Can't be just a ``unit -> map`` function?

theimowski:
nice :+1:

src/Paket.Core/LockFile.fs
mexx: Memorization, it only needs to be calculated once.

PR #533, opened 2015-01-08
Rop init & init auto restore command

theimowski:
actually ``init`` is so simple, there's no failure track for now 
references #464

### DIFF
--- a/src/Paket.Core/Environment.fs
+++ b/src/Paket.Core/Environment.fs
@@ -85,9 +85,10 @@ module PaketEnv =
         match locatePaketRootDirectory directory with
         | Some rootDirectory -> 
             Logging.tracefn "Paket is already initialized in %s" rootDirectory.FullName
+            succeed ()
         | None -> 
             let dependenciesFile = 
                 DependenciesFile(
                     Path.Combine(directory.FullName, Constants.DependenciesFileName), 
                     InstallOptions.Default, [], [], [])
-            dependenciesFile.Save()
+            dependenciesFile.ToString() |> saveFile dependenciesFile.FileName
--- a/src/Paket.Core/Environment.fs
+++ b/src/Paket.Core/Environment.fs
@@ -85,38 +85,10 @@ module PaketEnv =
     let init (directory : DirectoryInfo) =
         match locatePaketRootDirectory directory with
         | Some rootDirectory -> 
-            fromRootDirectory rootDirectory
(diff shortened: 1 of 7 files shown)


src/Paket.Core/VSIntegration.fs
```diff
@@ -3,39 +3,50 @@
...
+        let releasesUrl = "https://api.github.com/repos/fsprojects/Paket/releases";
+        let! data = client |> downloadStringSync releasesUrl
+        let! latestVersion = getLatestVersionFromJson data
+
+        let! downloads = 
+            ["paket.targets"; "paket.bootstrapper.exe"] 
+            |> List.map (download latestVersion)
+            |> collect
+
+        let projectsUnderPaket =
```
mexx: aren't the projects already in `environment`?

theimowski: good spot! I totally forgot about it

src/Paket.Core/Environment.fs
```diff
@@ -80,3 +80,14 @@ module PaketEnv =
...
+    let init (directory : DirectoryInfo) =
+        match locatePaketRootDirectory directory with
+        | Some rootDirectory -> 
+            Logging.tracefn "Paket is already initialized in %s" rootDirectory.FullName
+        | None -> 
+            let dependenciesFile = 
+                DependenciesFile(
+                    Path.Combine(directory.FullName, Constants.DependenciesFileName), 
+                    InstallOptions.Default, [], [], [])
+            dependenciesFile.Save()
```
mexx: Actually this `Save` call can fail. For example insufficient rights to create a file.

src/Paket.Core/VSIntegration.fs
```diff
@@ -3,39 +3,46 @@
...
+    try 
+        let start = data.IndexOf("tag_name") + 11
+        let end' = data.IndexOf("\"", start)
+        (data.Substring(start, end' - start)) |> SemVer.Parse |> succeed
+    with _ ->
+        fail ReleasesJsonParseError
+
+let InitAutoRestore environment =
+    let exeDir = Path.Combine(environment.RootDirectory.FullName, ".paket")
+    CreateDir(exeDir)
```
mexx: This one can also fail due to permissions.

mexx: Btw. I don't like to have this one explicitly called, for me it should be part of `download` function.
No one needs this directory to be present besides the download, as it would fail.

theimowski:
:shipit: !

forki:
thanks

Issue #536, opened 2015-01-09
Paket restore doesn't download http dependencies


tpluscode:
As per the title please try doing `paket restore` in repository https://github.com/wikibus/JsonLD.Entities. The one http dependency will not be downloaded. It's defined as follows

```
http http://www.w3.org/2013/N-QuadsTests TESTS.zip
```

forki:
![image](https://cloud.githubusercontent.com/assets/57396/5685401/28a5cd10-983a-11e4-8a37-8ac50526d03a.png)

forki:
can't reproduce. works as expected

tpluscode:
You're showing me github files. There should be two folders in paket-files

forki:
ah I see

forki:
after changing the link (see #537) and running update we get a change in the lockfile:

![image](https://cloud.githubusercontent.com/assets/57396/5686196/419ef8f8-9841-11e4-94ee-d4ef82b393d6.png)

I can now confirm: restore fails to parse this.

tpluscode:
Thank you for acting so quickly again :+1:

So is full URI the correct syntax for http dependency?

forki:
yep. it would be awesome if you could try to reread the docs and improve it if it is wrong or misleading.

tpluscode:
sure thing. will do first thing tomorrow. maybe after breakfast ;)

tpluscode:
Hm, I wanted to create a pull request for gh-pages but I see that you keep entire HTML documents. Are they generated from somewhere with Jekyll or whatever?

theimowski:
We use [FSharp.Formatting](https://github.com/tpetricek/FSharp.Formatting) to generate the docs. Just edit one of the markdown files: https://github.com/fsprojects/Paket/blob/master/docs/content/http-dependencies.md and run ``build GenerateHelp`` - it should generate the htmls to /docs/output

forki:
There is also
https://github.com/fsprojects/Paket/blob/master/src/Paket/HelpTexts.fs -
this file contains the sources of the command texts

PR #541, opened 2015-01-10
fix http dependency documentation

tpluscode:
Reflects change which fixed issue #536. Should be enough for now


forki:
Thanks. Very cool

PR #551, opened 2015-01-14
WIP: Separate args for each command

theimowski:
Because Paket.exe is always to be invoked with a given command, I thought it would be nice to separate arguments for each of the commands.

 - clearer view of which command uses what arguments
 - easier to express the syntax of commands (see f.e issue in #511)
 - good way to maintain the command line help (will need to fill in the usage for arguments)
 - potential for automated (build process) command syntax in markdown docs

Tell me what you think, if you give green light I'll proceed with filling up the usages, and what else gives.

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -119,13 +119,85 @@ let filterGlobalArgs args =
 
     verbose, logFile, rest
 
+type AddArgs =
+    | [<First>][<CustomCommandLine("nuget")>][<Mandatory>] Nuget of string
+    | [<CustomCommandLine("version")>] Version of string
+    | [<AltCommandLine("-f")>] Force
+    | [<AltCommandLine("-i")>] Interactive
+    | Hard
+    | No_Install
+with 
+    interface IArgParserTemplate with
+        member __.Usage = ""
+
+type ConvertFromNugetArgs =
+    | [<AltCommandLine("-f")>] Force
+    | No_Install
+    | No_Auto_Restore
+    | Creds_Migration of string
+with 
+    interface IArgParserTemplate with
+        member __.Usage = ""
+
+type FindRefsArgs =
+    | [<Rest>][<CustomCommandLine("nuget")>][<Mandatory>] Packages of string
+with 
+    interface IArgParserTemplate with
+        member __.Usage = ""
+
+type InitArgs =
+    | [<Hidden>] NoArg
+with 
+    interface IArgParserTemplate with
+        member __.Usage = ""
+
+type InstallArgs =
+    | [<AltCommandLine("-f")>] Force
+    | Hard
+    | Redirects
+with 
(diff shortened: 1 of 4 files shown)


src/Paket/Program.fs
```diff
@@ -90,106 +79,68 @@ with
...
-            match command with
-            | Command.Init -> Dependencies.Init()
-            | Command.Add -> 
-                let packageName = results.GetResult <@ CLIArguments.Nuget @>
-                let version = 
-                    match results.TryGetResult <@ CLIArguments.Version @> with
-                    | Some x -> x
-                    | _ -> ""
+    match args with
+        | "add" :: args ->
```
forki: why did we loose the nice discriminated union case here?
@mexx what do you think?

forki:
Yes I think making the args more explicit per command is a good thing.

src/Paket/Program.fs
mexx: Actually neither version looks good to me.
The old one do not support our use case for having separate argument for each command well.
The new one tries to provide the desired behavior but introduces redundancy and magic values.

I have to look further into `Nessos.UnionArgParser` to see how we can do better here.
I'll try to look at it on this weekend.

theimowski: Hm I'll see what I can do to bring the Commands DU back. @mexx what type of redundancy? You mean arguments that are applicable to more than one command?

theimowski:
I added the Command DU - does it look any better now?

src/Paket/Program.fs
```diff
@@ -15,199 +18,218 @@ let assembly = Assembly.GetExecutingAssembly()
...
+            let hard = results.Contains <@ AddArgs.Hard @>
+            let interactive = results.Contains <@ AddArgs.Interactive @>
+            let noInstall = results.Contains <@ AddArgs.No_Install @>
+            Dependencies.Locate().Add(packageName, version, force, hard, interactive, noInstall |> not)
+        
+    | Command(Config, args) ->
+        let results = commandArgs<ConfigArgs> args
 
+        if results.IsUsageRequested then
+            trace <| results.Usage("paket config")
```
mexx: Why not `showHelp HelpTexts.commands.["config"]`?

mexx:
Looks a little bit better now. With this version of `UnionArgParser` it seems to be the best achievable result, albeit the `filterGlobalArgs` looks like a hack with this magic strings in it.

src/Paket/Program.fs
theimowski: There's no "config" in this dictionary

mexx: It was the actual implementation. Congratulations, you found a bug ;)

theimowski: Oh, indeed, didn't notice that one ;)

PR #552, opened 2015-01-15
Fix proxy script use

gluck:
When using a proxy configuration script (aka wpad pac script), the actual
proxy settings you get depend on the url you target.
It's used for load balancing / web proxy routing in enterprises.

If you were unlikely enough (like me) to have a split between google urls
and the rest of the www, the current code would use the google proxy
to target github urls, which was failing (405 in my case but I guess
that could depend on the actual setup).

(this code is part 1/2 to fix the bootstrapper, paket code needs a similar fix as well)

### DIFF
--- a/src/Paket.Bootstrapper/Program.cs
+++ b/src/Paket.Bootstrapper/Program.cs
@@ -8,13 +8,13 @@ namespace Paket.Bootstrapper
 {
     class Program
     {
-        static IWebProxy GetDefaultWebProxy()
+        static IWebProxy GetDefaultWebProxyFor(String url)
         {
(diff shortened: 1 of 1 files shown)


vasily-kirichenko:
It should work. "google.com" was chosen in a random basis. Please merge it and release so I can test it.

forki:
@gluck thanks for contributing. Could you please review 9cc8d5

It's both released as 0.23 - let's see how it goes.

vasily-kirichenko:
Works OK for me.

gluck:
Today won't be the day where I make my first F# contribution :sob: 
j/k thanks for finishing up the work, change looks good.

forki:
Sorry about that.

PR #555, opened 2015-01-19
Use ILRepack 1.25 from nuget.org

forki:
Still WIP since ILRepack is throwing an error:

image


forki:
Tracking issue at https://github.com/gluck/il-repack/issues/72

gluck:
You can use 1.22.2 from NuGet: https://www.nuget.org/packages/ILRepack/1.22.2

It'll likely work and prevent you from having a local copy.
That being said I know about this issue (nested generic parameters referencing) and have found to 100% fix so far.

(though it seems the 1.22.2 code generates less frustration than the latest version).

Sorry about that, I'll try to give another go at this issue sometime.

Issue #542, opened 2015-01-12
Create new command to automatically remove paket.targets from all projects in solution


vasily-kirichenko:
It calls `paket install` for each project in solution and it often fails (exit code = 1) on TeamCity. 
1. Do I understand properly that I can remove `paket.targets` from all projects if `paket install` is always launched by hand or build.cmd?
2. If above statement is true, how I can remove it automatically?

forki:
0. It should only call paket restore. 
1. Yes
2. We need the inverse of http://fsprojects.github.io/Paket/paket-init-auto-restore.html

vasily-kirichenko:
yes, it calls `paket restore`, sorry.

vasily-kirichenko:
I'm right that it's enough to remove 
```xml
<Import Project="..\..\.paket\paket.targets" />
```
from all projects?

forki:
yes.
or you could set your https://github.com/fsprojects/Paket/blob/master/.paket/paket.targets#L5 to false (maybe only in TC)

forki:
if you manage to set the RestorePackages flag to false when inside a TC (FAKE?) build then this would probably be the best solutions. Then Restore from inside VS would still work.

vasily-kirichenko:
I just set it to `false` permanently.

theimowski:
maybe we could make a toggle command like ``paket auto-restore on`` \ ``paket auto-restore off``

forki:
Yes.

PR #557, opened 2015-01-19
Auto restore on off

theimowski:
references #542

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -130,13 +130,13 @@ try
         else
             Dependencies.Init()
 
-    | Command(InitAutoRestore, args) ->
-        let results = commandArgs<InitAutoRestoreArgs> args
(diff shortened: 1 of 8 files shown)


src/Paket/Program.fs
```diff
@@ -130,13 +130,13 @@ try
         else
             Dependencies.Init()
 
-    | Command(InitAutoRestore, args) ->
-        let results = commandArgs<InitAutoRestoreArgs> args
+    | Command(AutoRestore, args) ->
+        let results = commandArgs<AutoRestoreArgs> args
 
-        if results.IsUsageRequested then
```
mexx: Shouldn't this remain in new version?
I mean if help is requested we shouldn't perform any action even when all parameters for the action are provided.

theimowski: yes, you're right. I'll send a correction. thanks

PR #568, opened 2015-01-25
ROP for config command

theimowski:
Some IO checks for the add-credentials command

#464

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -93,13 +93,16 @@ try
             trace <| results.Usage("paket config")
         else
             let args = results.GetResults <@ ConfigArgs.AddCredentials @> 
-            let source = args.Item 0
-            let username = 
-                if(args.Length > 1) then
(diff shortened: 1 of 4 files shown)


forki:
cool thanks

src/Paket/Program.fs
```diff
@@ -93,13 +93,16 @@ try
...
+            if args.Length = 0 then
+                trace <| results.Usage("paket config")
+            else
+                let source = args.Item 0
+                let username = 
+                    if(args.Length > 1) then
+                        args.Item 1
+                    else
+                        ""
+                Dependencies.Locate().AddCredentials(source, username)
```
mexx: This changed the behavior, former to this change you could run `paket config` without having a `paket.dependencies` file, now `Dependencies.Locate()` will complain about not having it.

theimowski: good point, do you think we should bring this thing back?

Issue #101, opened 2014-09-15
Handle NuGet Build `.props` `.targets`


bartelink:
Unless I'm mistaken, Paket does not yet hook in [`.props` and `.targets` files as NuGet has done since 2.5](http://docs.nuget.org/docs/release-notes/nuget-2.5); I believe it should.

The primary reason is that more often than not, it presents an appropriate way for packages to manage [adjustments to the buld process such as this](https://github.com/fsprojects/Paket/issues/65#issuecomment-55561378)

Key differences between such `.props` (imported at top of `.*proj`)  / `.targets` (imported at bottom of `.*proj`) and the `.ps1` are:

- no reliance on VS
- can deal with `TargetPlatform` etc. without silent problems and/or need to reinstall packages
- doesnt induce diffs when swithcing build output type

Note that `.props` is wired into VS with an editor for editing MSBuild Properties entries.

forki:
conversation is happening at #516

PR #578, opened 2015-02-02
Install props and targets files to the project

forki:
this implements #516 and #101 

It's prototype and needs review/testing


forki:
this is now an alpha release

PR #580, opened 2015-02-03
[convert-from-nuget] ignore case when looking for nuget.targets - closes #577

### LINKED ISSUE #577: convert-from-nuget does not remove import for NuGet.targets
@theimowski 
Current implementation only removes lower-cased `nuget.targets`:
The following line does not get removed:
```
<Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
```

theimowski:


forki:
thanks for fixing. the red build was my fault

theimowski:
ok I didn't even notice build failure

PR #581, opened 2015-02-03
[paket add] remove FirstAttribute from Nuget switch for add command - closes #573

### LINKED ISSUE #573: Command line switches can't be positioned right after the action.
Following command line `paket add -i nuget FSharp.Data` is now illegal.
Also `paket add nuget -i FSharp.Data` isn't working.
Caused by #551.
However `paket add nuget FSharp.Data -i` function properly.

theimowski:


forki:
sorry. forgot about that one.

PR #588, opened 2015-02-08
Replace illegal <Import> attr 'Paket' with Label="Paket"

damageboy:
This replaces the Paket attribute used to identify paket-controlled `<Import>` elements a `Label` attribute that doesn't make VS b0rk


forki:
lbspd6y


Thanks a lot.

Issue #600, opened 2015-02-13
paket.version file not generated (0.27.11)


cdrnet:
The first run of build.sh of ProjectScaffold fails with

    Paket failed with:
        Could not find file 'c:\ [...] \fsharp\FAKE\modules\Octokit\paket.version'.

Indeed, the file is not there. This seems to be a recent regression as it works fine in 0.27.10 but fails with 0.27.11.

forki:
Ok. Will look at this. /cc @sergey-tihon

cdrnet:
Looks like [here](https://github.com/sergey-tihon/Paket/commit/1b76f1ee5705cdcd789c1dbae8f7c3522e31b2a0#diff-f49621e896c9b89745f9d4aa9e06cb6fL168) we should check whether the file exists, if not write it as before.

forki:
Could you please send a pull request?

cdrnet:
Yes, will do.

cdrnet:
Hm, that does not seem to have been enough, will check

sergey-tihon:
Ouch, sorry guys
Thank you @cdrnet

cdrnet:
Argh, I missed the operator precedence ... will submit a new PR, sorry for the extra work

PR #601, opened 2015-02-13
Fix for fsprojects/Paket#600 - paket.version file not generated

cdrnet:
Check paket.version file actually exists before reading it.

See #600.


forki:
Thanks for solving this. :heart:

Issue #583, opened 2015-02-07
Add package to specified project


ascjones:
When dealing with solutions with many projects (not F# solutions :-)), and you want to install a package to only a single project it can be annoying to go through all the projects using the interactive option with paket add.

So how about an option to add the package to a specified project? Similar to the -ProjectName parameter in the Install-Package nuget cmdlet. Something like this perhaps:

```paket add nuget SomePackage project MyProject```

Is this desirable, and is the above syntax the correct style? I based it on the 'version' option of paket update.

I've had a quick stab at this: https://github.com/betgenius/Paket/commit/14ce3a2632ed052bce0e8b9acb13ee7924988ae2

Let me know what you think and I will create a pull request if you agree it's a good idea. I certainly would use this feature!

isaacabraham:
linked to #352

forki:
Sounds good.

ascjones:
Ah thanks @isaacabraham - didn't know about that way of doing it. Now I read the docs for ```paket install``` properly it does mention that behaviour.

So is it still worth adding the command line 'project' option I've done, or not? 

I suppose the benefit would be just running one command ```add``` instead of ```add```; edit ```paket.references```; then ```install```. 

If so I will just test it properly before sending a PR.

isaacabraham:
I think an add command would be a good feature - particularly if you already have a package and want to add it to (or remove from)  a project.
________________________________

PR #610, opened 2015-02-16
option to add/remove package for a single project

ascjones:
As per https://github.com/fsprojects/Paket/issues/583

Added a `project` option to both add and remove commands so you can specify an individual project to add/remove a package. This is especially useful in some of our large solutions where using the interactive option is a bit unwieldy.


forki:
yeah

Issue #521, opened 2015-01-06
NuGet pack on projects with paket dependencies


mavnn:
If you NuGet pack a proj file with an associated nuspec file (same filename in same directory), NuGet automatically populates a lot of the nuspec fields from the proj file.

Most of these work fine, regardless of whether the proj file has NuGet or Paket dependencies - but the automatic dependency resolution obviously doesn't!

I'm looking at writing something that updates the nuspec dependency nodes; the logic NuGet uses is documented here: http://docs.nuget.org/docs/reference/command-line-reference#Pack_Command and is in general quite sane.

All of this to say: would you like this to be (an optional) part of the Paket code, or shall I write it as something independent?

And any pointers on whether we do anything around project references at the moment?

forki:
1) https://github.com/tpetricek/FSharp.Formatting/blob/master/build.fsx#L153 works with Nuget and Paket
2) Paket.Core allows you to query the lock file

mavnn:
1) is only really convenient for solutions with a single packaged output. We regularly have solutions that generate tens of dlls which need packaging, and I don't think it's that unusual. Also - those dependencies had to be added manually, but could have been calculated automatically from Paket data and project references. NuGet does this already, and we leverage it heavily in larger solutions.
2) It does indeed, I've now discovered! Which leads to...

A proposal. @vasily-kirichenko , I know you're using paket in larger solutions as well, I'd love your input.

Packaging Paket based solutions
==========================

For single dll projects, project scaffold gives a good example of how a nuget package can be created. However, it does not scale well for solutions with many output NuGet packages.

To avoid confusion for people moving from NuGet, and frankly also because it seems a pretty sensible idea, I'd recommend a Paket based option that closely follows the [NuGet "From a Project"](http://docs.nuget.org/docs/creating-packages/creating-and-publishing-a-package) flow.

This relies on a nuspec file in the same directory as a *proj file; when you pack, you pack based on the project file which then loads up the nuspec, and substitutes in token values based on the build output of the project. It will also optionally create a dependencies node in the nuspec with required nuget dependencies.

To achieve the same functionality in a Paket based solution, we will need the following steps (steps in brackets are only required when generating the dependencies node):

* Check if the project has an associated nuspec file
  * skip if not
  * check which tokens have been used - we might be able to skip some of the other steps
* Scan the project file to determine:
  * The output file
  * (Project references where the target project has a nuspec file)
* Scan the output file (with reflection) to load its assembly info
[...]

PR #613, opened 2015-02-17
Pack command

mavnn:
The first (and most important) part of #521 . Still missing a convert from nuget option, but I'm slightly unsure how to proceed on that one as I have not supported (and do not plan to support) all possible nuspec options. Still, this should be usable in it's own right.

This is still very much alpha quality, but it does work for most simple cases. I'm not sure if you want to merge now with warnings or leave this pull request open for a while. It's already pretty chunky though!


forki:
tpys3sx

mavnn:
I'll try and get the test passing in mono as well...

docs/content/paket-init-auto-restore.md
```diff
@@ -0,0 +1,11 @@
+# paket init-auto-restore
```
forki: This file is generated from https://github.com/fsprojects/Paket/blob/master/src/Paket/HelpTexts.fs please don't commit it.

docs/content/paket-pack.md
```diff
@@ -0,0 +1,11 @@
...
+Packs all paket.template files within this repository
+
+    [lang=batchfile]
+    $ paket pack output outputDirectory buildconfig Debug
+
+Options:
+
+  `output`: Output directory to put nupkgs
+
+  `buildconfig`: Optionally specify build configuration that should be packaged (defaults to Release).
```
forki: $ paket pack output outputDirectory buildconfig Debug

should be 

  $ paket pack output outputDirectory [buildconfig Debug]

to indicate optionality

forki:
I think we should merge this and release it as prerelease in order to give people a chance to try it out.
Then we should change Paket's own build process to dogfood this. (and maybe add a push command to get rid of Nuget.CommandLine)

docs/content/paket-pack.md
mavnn: Should this file be auto created as well?

forki: yes. It should also go into helptexts. This will allow us to call `paket pack --help` and get the text in the command line

mavnn:
@forki the last travis build failed due to mono not being installed. Can we trigger a rerun to test the path changes? Either that or a helpful volunteer running some tests on *nix would be great!

mavnn:
Agreed it would be good to dog food this as a pre-release. We'll be doing the same thing internally here, so between the two we should iron out the bugs in a couple of different use cases.

forki:
an easy way to trigger the build is to force push a small change. Maybe
it's a good opportunity to rebase on current master ;-)

2015-02-17 17:50 GMT+01:00 Michael Newton <notifications@github.com>:

mavnn:
You and your fancy rebasing. Squashed a few commits for neatness as well.

mavnn:
...and messed up the rebase :(. I'll try again shortly.

forki:
yay we have a green build

mavnn:
Yeah, actual tests detecting actual (potential) bugs and everything. I feel all engineery, although I could have spotted the root cause a little quicker :)

Just spotted that it doesn't create the output directory if it doesn't already exist. I'll add that, and then it'd be good to merge to start the dog fooding process.

mavnn:
Ship it!

Basic sanity checks and build both pass.

mavnn:
:+1:

Issue #604, opened 2015-02-13
bcl.targets is not compatible with Paket


isaacabraham:
If you reference any of the BCL nuget packages, they add a target (Microsoft.Bcl.Build.targets) to the project file. This in turn tries to validate package references by looking for, you guessed it - packages.config.

Not sure what the solution to this is but it leads to a build error and associated warnings.

forki:
what about backlisting these?

mavnn:
I have never had anything but pain from bcl.targets, and I've never seen a failure caused by removing them.

Even with Nuget they tend to break if you configure a non-standard package directory.

forki:
would love to see a pull request which blacklists these targets files.

isaacabraham:
@mavnn - they confuse the hell out of me too.
@forki - I assume that we're seeing them now that targets support is in paket. Yes a blacklist a good idea. Maybe the only idea aside from manually rewriting the targets file to not look for packages.config. But I don't think that we want to go down that route!

A bit concerned about where this path will lead us down though.

forki:
I made this up-for-grabs, so if anyone wants to take it....

Blacklist is found here: https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/InstallModel.fs#L228 

Sample for a test case: https://github.com/fsprojects/Paket/blob/master/tests/Paket.Tests/InstallModel/Xml/LibGit2Sharp.fs

dedale:
I cannot reproduce the build failures under VS. The Bcl targets tells me that adding the following property in the project would skip nuget validation:

    <SkipValidatePackageReferences>true</SkipValidatePackageReferences>

damageboy:
@forki I'll take a stab at it.
I assume some sort of and embedded resource like a `blacklist.txt` file that is part of paket is acceptable?

ctaggart:
Can I get an option to turn off this feature? Like NuGet content support, this is a feature that I would like to turn off. I was very excited about it when came out in NuGet 2.5, but it doesn't work very well. The `Microsoft.Bcl.Build.targets are preventing me from upgrading paket.

ctaggart:
How about a feature to opt-out of automatic import of targets with syntax from #574.
```
nuget Foobar 1.2.3 framework: net20 net35, autoimport: false
```

damageboy:
I leave it to @forki to decide.

[...]

PR #618, opened 2015-02-18
blacklisted Microsoft.Bcl.Build.targets

ascjones:
As discussed https://github.com/fsprojects/Paket/issues/604#issuecomment-74762408

Wasn't 100% sure where to put the actual blacklist itself so just added as a separate file `Blacklist.fs` like @isaacabraham suggested.


forki:
tree

very cool thanks.

PR #623, opened 2015-02-19
paket.template should not require to be prefixed with project file name

### LINKED ISSUE #622: paket.template should not require to be prefixed with project file name
we should use the same file detecting algorithm as for paket.references.

/cc @mavnn

forki:
this fixes #622

@mavnn what do you think?


mavnn:
Looks much more DRY (and more usable). I like it.

PR #624, opened 2015-02-19
Pack all template files in one go

forki:
I think all template files should be packed in one go.
@mavnn  what do you think?


mavnn:
I split it purely because it allows a background thread to start packaging file only packages while we're still reading project metadata.

I suspect this would make a significant difference in packaging time in some of our larger solutions with both a lot of deployment packages (not project based) and project based dll packages.

mavnn:
Splitting it does simplify the code however, which is never a bad thing. I'll leave it to your call unless we start hitting performance issues.

forki:
In that case we could wrap the file replacement also in an async. At the moment I want to be able to debug everything in one go.

PR #629, opened 2015-02-19
Dog fooding "pack" and "push"

### LINKED ISSUE #619: Dogfood pack command
We now have the pack command available in the latest alpha version (see #613 and [docs](http://fsprojects.github.io/Paket/template-files.html)).
We should start by dogfooding it.

/cc @mavnn

mavnn:
Closes #619 

Bit ugly until we get a fake helper.

I'm also wondering if we want to expand the number of fields we'll infer from projects beyond what NuGet allows. Later, I think...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -201,3 +201,4 @@ docs/content/paket-convert-from-nuget.md
 docs/content/paket-pack.md
 docs/content/paket-push.md
 *.bak
+paket.template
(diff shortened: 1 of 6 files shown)


mavnn:
This uses both a "type project" (for Paket.Core) and a "type file" (the ilmerged paket.exe) template, so if you can publish using this I would say we have a decent sanity check of the core functionality.

forki:
happy dogfooding
i-have-no-idea-what-im-doing-dog

forki:
image

;-)

It's trying to push packages from the `packages` folder. We need to make sure to only push our own pckages. Ideas?

mavnn:
Hmm. Yes, looks like we need to take a root directory parameter like for do for the packing. I was just scanning from dependency file down.

Give me a bit, I might get grabbed for lunch in the middle.

PR #630, opened 2015-02-19
Don't try and push other peoples nupkg files

mavnn:


mavnn:
Just realised that this build.fsx relies on the version that hasn't been published yet. Might be worth repointing at the bin directory instead

forki:
I'm already trying to solve the hen and egg problem ;-)

forki:
ok it now pushed only paket.core. It seems to miss paket.nupkg

mavnn:
That's very odd. Did it create paket.nupkg?

forki:
nope.

forki:
wait a minute. could still be related to hen and egg

mavnn:
No, we've changed template file detection. It now only looks for templates from projects.

mavnn:
I've commented on the commit that changed it, I misunderstood the new logic originally.

forki:
ok it's "fixed", but I need to cleanup my mess

PR #634, opened 2015-02-20
Use relative path references more consistently in docs

tpetricek:
This should solve the issue here: https://github.com/tpetricek/FSharp.Formatting/issues/263#issuecomment-75116353

In ProjectScaffold (and lots of other projects), you can build the documentation in two ways:

 - Running the build, which uses the absolute website path for all references
 - Running code from `generate.fsx` in F# Interactive, which uses absolute local path

See the `#if` I deleted in this commit. This means that doing `build.sh` will generate docs with reference to the live web site.

This change deletes all of this and just uses relative references. This works for the current docs, but it is fragile - in particular, this would break if you started adding new tutorial documents into sub-folders. But then we'd probably have to track the relative paths in F# Formatting (in some clever way).

### DIFF
--- a/docs/tools/generate.fsx
+++ b/docs/tools/generate.fsx
@@ -12,8 +12,6 @@ Paket.HelpTexts.commands
 
 // Binaries that have XML documentation (in a corresponding generated XML file)
 let referenceBinaries = [ "Paket.Core.dll" ]
-// Web site location for the generated documentation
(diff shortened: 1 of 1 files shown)


forki:
yes that seems to work locally and on the server. Could you please apply the same thing to ProjectScaffold.

PR #644, opened 2015-02-23
Minor packaging fixes

mavnn:
* Prioritize values from template file over values from dll attributes
* Ignore attributes in dlls which exist but have no valid value set

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -15,12 +15,34 @@ let (|CompleteTemplate|IncompleteTemplate|) templateFile =
 
 let (|Title|Description|Version|InformationalVersion|Company|Ignore|) (attribute : obj) = 
     match attribute with
-    | :? AssemblyTitleAttribute as title -> Title title.Title
-    | :? AssemblyDescriptionAttribute as description -> Description description.Description
-    | :? AssemblyVersionAttribute as version -> Version(SemVer.Parse version.Version)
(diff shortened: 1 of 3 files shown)


src/Paket.Core/PackageMetaData.fs
```diff
@@ -93,10 +115,10 @@ let merge templateFile metaData =
     match templateFile with
     | { Contents = ProjectInfo(md, opt) } -> 
         let merged = 
-            { Id = md.Id ++ metaData.Id
-              Version = md.Version ++ metaData.Version
-              Authors = md.Authors ++ metaData.Authors
-              Description = md.Description ++ metaData.Description }
+            { Id = metaData.Id ++ md.Id
```
forki: so you changed the usage and not the operator.

forki:
thanks again

src/Paket.Core/PackageMetaData.fs
mavnn: Yes, as I wasn't sure where else it might be used (and what the desired
effect would be)

forki: but originally you intended it to be the other way around, right?
So let me switch all of these ;-)

mavnn: Go for it :+1: . The operator makes more sense the other way round to me - it takes the first value unless it's ``None``

forki: done. and released

PR #647, opened 2015-02-24
Detect 4.5.1

JonCanning:


src/Paket.Core/FrameworkHandling.fs
```diff
@@ -48,7 +48,8 @@ module KnownAliases =
          "2.0", "20" 
          "3.5", "35" 
          "4.0", "40" 
-         "4.5", "45" 
+         "4.5.1", "451"
```
forki: do we need 4.5.2, 4.5.3?

Can we do something like Replace(".","")?

JonCanning:
Like that?

forki:
Lgtm. Thanks

PR #649, opened 2015-02-24
Handle transient dependencies better in pack command

mavnn:


src/Paket.Core/PackageMetaData.fs
```diff
@@ -171,13 +171,27 @@ let findDependencies (dependencies : DependenciesFile) config (template : Templa
...
         ProjectFile.FindReferencesFile <| FileInfo project.FileName |> Option.map (ReferencesFile.FromFile)
     match referenceFile with
     | Some r -> 
         r.NugetPackages
-        |> List.map (fun np -> np.Name.Id, dependencies.DirectDependencies.[np.Name])
+        |> List.map (fun np ->
+                let dep =
+                    match Map.tryFind np.Name dependencies.DirectDependencies with
+                    | Some direct -> direct
+                    // If it's a transiant dependency set
```
forki: typo?

forki:
which case are we fixing here?

src/Paket.Core/PackageMetaData.fs
mavnn: Yes :)

mavnn:
This fixes a "not found" exception when a package listed in ``paket.references`` for a project is not listed in the ``paket.dependencies`` file.

It came up for me in a solution where one project relied on package A, which depends on package B - and another project relied directly on package B. Package B was removed from ``paket.dependencies`` by simplify and :boom:

src/Paket.Core/PackageMetaData.fs
```diff
@@ -171,13 +171,27 @@ let findDependencies (dependencies : DependenciesFile) config (template : Templa
...
     // Add any paket references
     let referenceFile = 
         ProjectFile.FindReferencesFile <| FileInfo project.FileName |> Option.map (ReferencesFile.FromFile)
     match referenceFile with
     | Some r -> 
         r.NugetPackages
-        |> List.map (fun np -> np.Name.Id, dependencies.DirectDependencies.[np.Name])
+        |> List.map (fun np ->
+                let dep =
+                    match Map.tryFind np.Name dependencies.DirectDependencies with
```
forki: do we need to go via NormalizedPackageName or did we get both sides in correct casing?

mavnn: Ah... honestly not sure? I suppose we'd have to normalize both sides at that point; I'd take your advice as to whether it's needed.

forki: I will check

PR #650, opened 2015-02-24
fix paket add/remove with just project name

ascjones:
Fix for adding by project name (without .*sproj extenstion) for projects with names containing periods (e.g. Paket.Bootstrapper.csproj). My bad.

### DIFF
--- a/src/Paket.Core/AddProcess.fs
+++ b/src/Paket.Core/AddProcess.fs
@@ -37,7 +37,7 @@ let AddToProject(dependenciesFileName, package, version, force, hard, projectNam
         let project = 
             projects
             |> Seq.tryFind (fun p -> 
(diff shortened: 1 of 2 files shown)


forki:
thanks

src/Paket.Core/AddProcess.fs
```diff
@@ -37,7 +37,7 @@ let AddToProject(dependenciesFileName, package, version, force, hard, projectNam
         let project = 
             projects
             |> Seq.tryFind (fun p -> 
-                let name = p.Name.Split('.').[0]
+                let name = Path.GetFileNameWithoutExtension p.Name
```
forki: wow didn't know that API exists.
Probably a couple of other places where replace Extension manually.

PR #651, opened 2015-02-24
Uploading packages as multiform content type

### LINKED ISSUE #646: Paket push fails pushing to an Inedo ProGet feed
Trying to use the new Paket Push to push to our internal Inedo Proget (http://inedo.com/proget/overview) server.

Fails with status 400, invalid request.

Now, this might be due to ProGet not following the standard properly. 
It may be that Paket is not following the standard properly
Or it may be that ProGet uses some redirects that are not properly handled by Paket push.

I do know that using nuget push works as expected though.

Are there any good reasons not to piggyback on nuget for this feature. Is paket gaining anything (except 1 less dependency) from re-implementing this?

Vidarls:
Tested against our ProGet feed, and it seems to fix #646 

Implementation is rather crude, let me know if you see anything that needs to be polished.

### DIFF
--- a/src/Paket.Core/RemoteUpload.fs
+++ b/src/Paket.Core/RemoteUpload.fs
@@ -1,22 +1,43 @@
 ﻿module Paket.RemoteUpload
 
 open System
+open System.Globalization
 open System.IO
+open System.Net
+open System.Text
 open Paket
 open Paket.Logging
 
+type System.Net.WebClient with
+        member x.UploadFileAsMultipart (url:Uri) filename =
+            let fileTemplate = "--{0}\r\nContent-Disposition: form-data; name=\"{1}\"; filename=\"{2}\"\r\nContent-Type: {3}\r\n\r\n"
+            let boundary = "---------------------------" + DateTime.Now.Ticks.ToString("x", CultureInfo.InvariantCulture)
+            let fileInfo = (new FileInfo(Path.GetFullPath(filename)))
+            let fileHeaderBytes = String.Format(CultureInfo.InvariantCulture, fileTemplate, boundary, "package", "package", "application/octet-stream")
+                                  |> Encoding.UTF8.GetBytes
+            let newlineBytes = Environment.NewLine |> Encoding.UTF8.GetBytes
+            let trailerbytes = String.Format(CultureInfo.InvariantCulture, "--{0}--", boundary) |> Encoding.UTF8.GetBytes
+            x.Headers.Add(HttpRequestHeader.ContentType, "multipart/form-data; boundary=" + boundary);
+            use stream = x.OpenWrite(url, "PUT")
+            stream.Write(fileHeaderBytes,0,fileHeaderBytes.Length)
+            use fileStream = File.OpenRead fileInfo.FullName
+            fileStream.CopyTo(stream, (4*1024))
+            stream.Write(newlineBytes, 0, newlineBytes.Length)
+            stream.Write(trailerbytes, 0, trailerbytes.Length)
+            ()
+            
 let Push maxTrials url apiKey packageFileName =
     let rec push trial =
         tracefn "Pushing package %s to %s - trial %d" packageFileName url trial
         try
             let client = Utils.createWebClient(url, None)
             client.Headers.Add("X-NuGet-ApiKey", apiKey)
-            client.UploadFile(Uri(url + "/api/v2/package"), "PUT", packageFileName)
+            client.UploadFileAsMultipart (new Uri(url)) packageFileName
             |> ignore
             tracefn "Pushing %s complete." packageFileName
         with
         | exn when trial < maxTrials ->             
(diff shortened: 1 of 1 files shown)


forki:
thanks

src/Paket.Core/RemoteUpload.fs
```diff
@@ -1,22 +1,43 @@
...
+            stream.Write(trailerbytes, 0, trailerbytes.Length)
+            ()
+            
 let Push maxTrials url apiKey packageFileName =
     let rec push trial =
         tracefn "Pushing package %s to %s - trial %d" packageFileName url trial
         try
             let client = Utils.createWebClient(url, None)
             client.Headers.Add("X-NuGet-ApiKey", apiKey)
-            client.UploadFile(Uri(url + "/api/v2/package"), "PUT", packageFileName)
```
Vidarls: Removing the explicit `/api/v2/package` seems to be important in order to make it work for the proget feed.
I don't know how to test against nuget.

forki: we test nuget by dogfooding in paket. seems to work

mavnn: A pity, that should really be standard. Still, specifying it for nuget.org is a small price to pay for it working everywhere else...

forki:
ok it doesn't work for nuget.org. - the build doesn't complain, but the package is not listed https://www.nuget.org/packages/Paket/

mavnn:
@Vidarls did you list the proget server with a trailing '/'? @forki I wouldn't expect nuget.org to take it without the full url, I'm afraide

forki:
@mavnn what hack do I have to commit to make it work? we already have special cases for the "download" part. I have no issues with special casing nuget.org

mavnn:
If push url is nuget.org, add "/api/v2/package" to the end. Thing which makes me reluctant is that this should be the right endpoint for any server implementing the nuget rest interface.

mavnn:
Also, we'll need to check why the push didn't fail when it didn't work.

forki:
https://github.com/fsprojects/Paket/commit/afd8e52423ba724cdea4cd3782dfe4fd92f724db actually works. 

summoning our special nuget hero @maartenba

maartenba:
MyGet needs the api/v2/package if it's not there too :) It should be standard ideally.

forki:
so only proget doesn't want it? sounds strange

maartenba:
Convention rather than standard...

Vidarls:
The reason for ProGet to be picky is that they have a weird way of creating their URLs:

The short form is `http://host.domain/nuget/feedname` 
Which I believe will be redirected to: `http://host.domain/api/v2/package/feedname` 

Force appending `/api/v2/package` to any of these valid url will fail, usually by a `400 invalid request`

Now we could try to get ProGet to adhere to the conventions, but fact is that the nuget client have no problems with this. And as such, neither should Paket (following the principle of least surprise).

As far as I can see, Nuget will add `/api/v2/package` to a url if only the hostname is given:

(from http://nuget.codeplex.com/SourceControl/latest#src/Core/Server/PackageServer.cs) 

```c#
private const string ServiceEndpoint = "/api/v2/package";

internal static Uri GetServiceEndpointUrl(Uri baseUrl, string path)
        {
            Uri requestUri;
            if (String.IsNullOrEmpty(baseUrl.AbsolutePath.TrimStart('/')))
            {
                // If there's no host portion specified, append the url to the client.
                requestUri = new Uri(baseUrl, ServiceEndpoint + '/' + path);
            }
            else
            {
                requestUri = new Uri(baseUrl, path);
            }
            return requestUri;
        }
```

And also the baseurl is redirect-expanded (is that even a word?):

```c#
private Uri ResolveBaseUrl()
        {
            Uri uri;

            try
            {
                var client = new RedirectedHttpClient(new Uri(Source));
                uri = client.Uri;
            }
            catch (WebException ex)
            {
                var response = (HttpWebResponse)ex.Response;
                if (response == null)
                {
                    throw;
                }

                uri = response.ResponseUri;
            }

            return EnsureTrailingSlash(uri);
        }
```

I don't have time to keep shaving this particular yak, but this should be a good starting point should someone feel the urge..

mavnn:
I'll add api end point as an optional parameter later today (defaulting to the nuget convention). I'd rather not have too much "magic", but nuget.org should also just work™. I think the optional override makes the best compromise between the two.

mavnn:
At some point we should also look at following redirects, but that's a bigger yak than I have time to shave today!

Vidarls:
My research yesterday indicated that `WebClient` will use `webrequests` where `AllowAutoRedirect` is set to true, so in general, redirects should not be an issue, except if we need the redirected URL to do something with it. 
__Note:__ I did not find any _official msdn sources_ on this, but at least it seems to be the general consensus understanding that  WebClient will follow redirects automatically by default

mavnn:
@Vidarls that's good news.

PR #656, opened 2015-02-25
Extract markdown files

theimowski:


forki:
cool.

could you please check https://github.com/fsprojects/Paket/blob/master/.gitignore#L185 as well?

docs/content/commands/update.md
```diff
@@ -0,0 +1,12 @@
+## Updating a single package
+
+It's also possible to update only a single package and to keep all other dependencies fixed:
```
forki: where ist the `paket update` case which updates everything? did we lose it?

forki:
also the contributin guide in the readme needs to be updated.

docs/content/commands/update.md
theimowski: it's covered by the auto-gen syntax

theimowski: auto-gen gets rendered just above the optional additional markdown

forki: Awesome

theimowski:
re .gitignore - what I did is for each command generate auto syntax, and then if a corresponding markdown exists in `content\commands` then append the additional markdown, and finally write the result markdown to `content\paket-<command>.md` before FSharp.Formatting comes in. 
So I guess the ignore for `content\paket-*.md` should remain

forki:
ok. what about the readme?

theimowski:
there it is

forki:
:heart:

forki:
this is really awesome. the command line help feels so much better now. Thanks so much.

/cc @agross please give it a try

Issue #659, opened 2015-02-26
Authors node is sometimes left empty even if authors are specified in template


mavnn:
See https://github.com/fsprojects/FSharpx.Async/tree/212645ee810cc8defe9496501c158d979e03233e

Packaging claims to complete successfully, but there are no authors in the resulting nuspec despite the fact they are in the template.

This should:
a) work
b) if the authors weren't there or created an empty list, packing should fail

forki:
I think a + b ;-)

mavnn:
Agreed :)

mavnn:
Ok: this was a general problem with the parser. If you use the multi line syntax for setting a value

```
description
    is a description that works
```
but have white space at the end of the first line it thinks you're explicitly setting the value to an empty string (and then ignores the indented line underneath)

```
description 
    This one is broken. See how obvious the difference is?
```

I'll close this bug for now as it's really part of #625 - I should actually write a proper parser soon.

PR #660, opened 2015-02-26
Rewrite template file parsing

### LINKED ISSUE #625: Template file parser is fragile and very inefficient
But has unit tests. If someone wants to put the time into making it more efficient, feel free - I'm not going to get there for a while.

mavnn:
Closes #625 and tests for #659 

Before:

before

After:

after

It would be easier to smug about the two order of magnitude improvement if I hadn't written the original :)

### DIFF
--- a/src/Paket.Core/TemplateFile.fs
+++ b/src/Paket.Core/TemplateFile.fs
@@ -6,6 +6,80 @@ open System.IO
 open System.Text.RegularExpressions
 open Paket.Rop
 open Paket.Domain
+
+module private TemplateParser =
+    type private ParserState =
+        {
+            Remaining : string list
+            Map : Map<string, string>
+            Line : int
+        }
+
+    let private single = Regex("^(\S+)\s*$", RegexOptions.Compiled)
+    let private multi = Regex("^(\S+)\s+(\S.*)", RegexOptions.Compiled)
+
+    let private (!!) (i : int) (m : Match) =
+        m.Groups.[i].Value.Trim().ToLowerInvariant()
+
+    let private (|SingleToken|_|) line =
+        let m = single.Match line
+        match m.Success with
+        | true -> Some (!! 1 m)
+        | false -> None
+
+    let private (|MultiToken|_|) line =
+        let m = multi.Match line
+        match m.Success with
+        | true ->
+            Some (!! 1 m, m.Groups.[2].Value.Trim())
+        | false -> None
+
+    let private indented = Regex("^\s+(.*)", RegexOptions.Compiled)
+    let private (|Indented|_|) line =
+        let i = indented.Match line
+        match i.Success with
+        | true -> i.Groups.[1].Value.Trim() |> Some
+        | false -> None
+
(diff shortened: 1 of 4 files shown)


tpetricek:
It runs so fast that the colors don't even start fading out :-)

mavnn:
@tpetricek ha :). I rebuilt before realising I wanted a benchmark!

theimowski:
Nice!

src/Paket.Core/TemplateFile.fs
```diff
@@ -86,60 +160,29 @@ module internal TemplateFile =
             | ProjectInfo(core, optional) -> ProjectInfo(core, { optional with ReleaseNotes = Some releaseNotes })
         { templateFile with Contents = contents }
     
-    let private (!<) prefix lines =
```
forki: yay one custome operator less. :heart:

src/Paket.Core/TemplateFile.fs
```diff
@@ -6,6 +6,80 @@ open System.IO
...
+
+module private TemplateParser =
+    type private ParserState =
+        {
+            Remaining : string list
+            Map : Map<string, string>
+            Line : int
+        }
+
+    let private single = Regex("^(\S+)\s*$", RegexOptions.Compiled)
```
forki: the module is already private. So I think we don't need the private here

mavnn: Different levels of private. Just stops namespace pollution when you're in the rest of TemplateFile.

src/Paket.Core/TemplateFile.fs
mavnn: I added a new one to ensure balance in the force.

src/Paket.Core/TemplateFile.fs
mariusschulz: Out of curiosity, why isn't the regex pattern a verbatim string? Shouldn't the backslashes be escaped? I'm no F# expert, so I could be totally wrong here.

forki: I think normal quotes work like in C#, so yes that might be a bug!?

mariusschulz: In C#, it should definitely be `Regex(@"^(\S+)\s*$")` because `\S` should be passed to the regex engine, therefore you'd need `"\\S"` — or `@"\S"`, of course.

forki: care to send a fix?

mavnn: It's not a bug - it works correctly. It is pretty bad practice though :) . The ``@"..."`` version would be a much better idea, my bad.

mavnn: I can tidy that up.

mariusschulz: I'm surprised that it actually does work like this, but you're right, it's likely cleaner to properly escape the backslashes. Anyway, you got it covered. ;)

PR #663, opened 2015-02-26
Better url and endpoint handling

Vidarls:
Addresses fsharp/FAKE#667


forki:
apple

very cool. thanks for helping to make the product ready to use.

Issue #670, opened 2015-03-02
Using `pack`  - include project reference as dependency even if the id doesn't match?


ascjones:
For a project template file, the rule for determining dependencies is (as per the [docs](http://fsprojects.github.io/Paket/template-files.html)):

`the output assemblies of any project references which do not have a matching template file`

However, in most of our projects the package id doesn't match the project name since we use a company prefix on the package names. This means that inter-project dependencies that also publish their own packages will be included in the lib folder but not as dependencies.

So, how can I make it so that our inter project dependencies are added as dependencues even if the id doesn't match?

/cc @mavnn

mavnn:
It's not done on id at all: it's based on whether the linked proj file has
an associated paket.template.

Is it not working?

ascjones:
Oh right, let me try again

ascjones:
It is the behaviour I'm getting (v0.31.6.0):

paket.template id in referenced project == project name: project reference included as dependency
paket.tempate id in referenced project != project name: project reference included in lib folder

Quick look at the code - is this it:

https://github.com/fsprojects/Paket/blob/016da709ba8a74c7244bc9fbba3bac4be05adf8f/src/Paket.Core/PackageMetaData.fs#L145

Would run the debugger but gotta run now. Will check back tomorrow.

mavnn:
Ah, no: it's here https://github.com/fsprojects/Paket/blob/016da709ba8a74c7244bc9fbba3bac4be05adf8f/src/Paket.Core/PackageProcess.fs#L83 . id should be the project name (which is how we look it up later) not the nupkg id (which is what it's being set to currently).

forki:
So it's a bug?

mavnn:
Yes

ascjones:
Okay I need this so I'll fix it now, unless @mavnn is already on it

mavnn:
Go for it - I'm on other stuff today

PR #671, opened 2015-03-03
Fix for #670, inter project dependencies without matching package id

ascjones:
Fix for https://github.com/fsprojects/Paket/issues/670

Use the project name (without extension) to uniquely identify projects when determining inter project dependencies.

For this I added a new property on the ProjectFile for getting the project name without extension, so I've also used this in a couple of other places from some previous work I did.


forki:
thank you!

mavnn:
Thanks. Does it do what you want now?

ascjones:
Indeed it does, project references with corresponding paket.templates files are included as dependencies

mavnn:
Awesome :).

PR #675, opened 2015-03-04
Add support for WinExe output type

### LINKED ISSUE #674: Paket pack fails if output type is executable.
This is probably a bit of an edge case, but: I have a project that produces an `.exe` file, but this `.exe` file is also used as a library, so I want to pack it as a nuget package.

Leaving aside the discussion on where in the package folder this `.exe` file should go, paket .pack assumes it should be looking for a dll file, and failes packing when the `projectname.dll` file is nowhere to be found. (It should be looking for `projectname.exe`).

Vidarls:
Fixes #674 

From https://msdn.microsoft.com/en-us/library/bb629394.aspx 

>OutputType
 
>Specifies the file format of the output file. This parameter can have one of the following values: 

> * Library. Creates a code library. (Default value.)
> * Exe. Creates a console application.
> * Module. Creates a module.
> * Winexe. Creates a Windows-based program.
 
Now only the `module` output type is not handled by Paket, it will be treated as library output

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -456,8 +456,9 @@ type ProjectFile =
     member this.OutputType =
         seq {for outputType in this.Document |> getDescendants "OutputType" ->
                 match outputType.InnerText with
-                | "Exe" -> ProjectOutputType.Exe
-                | _     -> ProjectOutputType.Library }
+                | "Exe"    -> ProjectOutputType.Exe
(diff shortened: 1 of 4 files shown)


forki:
thanks for fixing this. :+1:

Vidarls:
Purely a self serving act :stuck_out_tongue:

Vidarls:
I on the other hand much appreciate the insanely quick turnaround time from PR to release :+1:

PR #681, opened 2015-03-05
remove hard path to bash

qbit:
On systems like OpenBSD bash is installed to /usr/local/bin/bash - this diff removes the hard coded /bin/bash in favor of pulling the path from the users environment.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 if test "$OS" = "Windows_NT"
 then
   # use .Net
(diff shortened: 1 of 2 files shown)


forki:
@ilkerde @dnauck I'd love to hear an opinion from someone else. You know I'm noob in such things...

ilkerde:
@qbit Thanks for fixing this. @forki Good stuff over here. It embraces crossplatform compatibility.

dnauck:
@forki this is a good change

cdrnet:
@qbit could you also propose this upstream in [ProjectScaffold](https://github.com/fsprojects/ProjectScaffold)? Thanks

forki:
Yes I was going to suggest this too.

qbit:
@cdrnet @forki @ilkerde will do! Thanks!

PR #695, opened 2015-03-11
show upload progress for paket push

kfrie:
this should work. Not so sure about the dependency on`FSharp.Control.Reactive` but a sampling operator is nice for this


src/Paket.Core/Paket.Core.fsproj
```diff
@@ -67,11 +67,11 @@
   -->
   <Import Project="$(SolutionDir)\.paket\paket.targets" />
   <ItemGroup>
-    <Compile Include="..\..\paket-files\fsprojects\Chessie\src\Chessie\ErrorHandling.fs">
+    <Compile Include="..\..\.paket\paket-files\fsprojects\Chessie\src\Chessie\ErrorHandling.fs">
```
forki: woups? what happened here?

paket.lock
```diff
@@ -4,6 +4,11 @@ NUGET
     DotNetZip (1.9.3)
     FAKE (3.21.0)
     FSharp.Compiler.Service (0.0.86)
+    FSharp.Control.Reactive (3.1.1)
```
forki: that's a huge dependency for a "progress bar" ;-(

src/Paket.Core/RemoteUpload.fs
```diff
@@ -7,6 +7,7 @@ open System.Net
 open System.Text
 open Paket
 open Paket.Logging
+open FSharp.Control.Reactive
 
 type System.Net.WebClient with
         member x.UploadFileAsMultipart (url:Uri) filename =
```
forki: If we have an Async version we don't need this one any more.

src/Paket.Core/RemoteUpload.fs
```diff
@@ -49,17 +91,24 @@ let GetUrlWithEndpoint (url: string option) (endPoint: string option) =
...
         tracefn "Pushing package %s to %s - trial %d" packageFileName url trial
         try
             let client = Utils.createWebClient(url, None)
             client.Headers.Add("X-NuGet-ApiKey", apiKey)
-            client.UploadFileAsMultipart (new Uri(url)) packageFileName
-            |> ignore
+            let uploadAsync, progressReport = client.UploadFileAsMultipartAsync (new Uri(url)) packageFileName
+            use progressSubscription = 
+                progressReport 
+                |> Observable.sample (TimeSpan.FromSeconds(1.))
```
forki: Is this the reason for the dependency?

forki: instead of updating every second. How about updating when things changed? Like every 10%?

kfrie:
Yes that is at least what it should do conceptually, since not every buffer update should be reported. I can try to achieve the same without the dependency, it will be just a bit hairy.

forki:
just dump the hairy parts into Utils.fs like we all do ;-)

kfrie:
ok :-)

forki:
is this done?

kfrie:
it's done now, I just added upload speed and completion percentage

forki:
wow cool. could you please remove the ol UploadFileAsMultipart?

forki:
mhm. I just merged it and tried it. It didn't show any progress at all.

image

forki:
@kfrie  any ideas?

kfrie:
@forki  the sample operator was blocked by `push |> Async.RunSynchronously`. I fixed this, can you reopen please?

forki:
I can't reopen. But you can send a new pull request. And many thanks

PR #713, opened 2015-03-20
[NuGetV2] verbose logging on exn

haf:
This PR implements verbose logging that's very helpful in determining what API is causing the problems.


forki:
thans.

PR #717, opened 2015-03-23
Minor fix for when the Paket project directory is not directly off the root filesystem

bobpace:
Example:
If I clone the repo here -> /home/devuser/data/Paket
and then run ./build.sh it will generate the files
/home/devuser/data/Paket/src/home/devuser/data/Paket/src/Paket.Core/AssemblyInfo.fs
/home/devuser/data/Paket/src/home/devuser/data/Paket/src/Paket/AssemblyInfo.fs

Because the Path.GetDirectoryName method returns the entire path:
'/home/devuser/data/Paket/src/Paket.Core' instead of just 'Paket.Core'

This fix wraps that piece in a Path.GetFileName call so it will cut it down to that 'Paket.Core' string that looks like was intended.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -78,7 +78,7 @@ let stable =
 
 let genFSAssemblyInfo (projectPath) =
     let projectName = System.IO.Path.GetFileNameWithoutExtension(projectPath)
-    let folderName = System.IO.Path.GetDirectoryName(projectPath)
+    let folderName = System.IO.Path.GetFileName(System.IO.Path.GetDirectoryName(projectPath))
     let basePath = "src" @@ folderName
     let fileName = basePath @@ "AssemblyInfo.fs"
     CreateFSharpAssemblyInfo fileName


bobpace:
Should also note this was while using mono 3.12.0 on linux (ubuntu 14.04)

forki:
Wow. That looks strange, but I assume it's a good fix

Vidarls:
Not sure,and to be honest when I look at this code now, I think we could simplify it.

If I'm not completely off this early in the morning, AssemblyInfo.fs is always in the same directory as the `.fsproj`file?

So we don't need to fool around with src folder..

```f#
let filename = (System.IO.Path.GetDirectoryName(projectPath) @@ "AssemblyInfo.fs")
```
should be enough, and also always correct?

Vidarls:
But the same fix should also be applied in ProjectScaffold as it uses the same kind of code

Issue #719, opened 2015-03-24
Convert-From-Nuget should not copy disabled Nuget feeds


colinbull:
I was just upgrading one of my older projects to paket and I noticed something strange, in the `paket.dependencies` file it has added a second source which doesn't seem to exist anywhere on my machine. 

The dependency file generated looks like this 

     source https://www.nuget.org/api/v2
     source http://hp20006551:8082/guestAuth/app/nuget/v1/FeedService.svc

     nuget DocumentFormat.OpenXml 2.5
     nuget FSharp.Compiler.Service 0.0.36
     nuget FSharp.Formatting 2.4.1
     nuget FSharp.TypeProviders.StarterPack 1.1.3.48
     nuget Microsoft.AspNet.Razor 2.0.30506.0
     nuget NUnit 2.6.3
     nuget NUnit.Runners 2.6.3
     nuget RazorEngine 3.3.0
     nuget NuGet.CommandLine 

What is really strange is the second feed only seems to be used on the `NuGet.CommandLine` package and I'm not sure even where the second feed has come from, looking at my nuget feed options in VS I don't have that listed. 

![2015-03-24 07_30_24-options](https://cloud.githubusercontent.com/assets/585546/6797266/b055c2be-d1f7-11e4-9635-6df70653ab73.png)

I'm guessing this is down to a nuget **feature** (ehemm), or some weird config on my machine rather than a paket problem. Since deleting the second source things work fine. So I have two questions

1. Where does paket read the feeds from? 
2. Why is it using the first source for all of the packages apart from `NuGet.CommandLine`?

theimowski:
here's the logic for extracting nuget feeds: https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NugetConvert.fs#L144-L159
basically it first looks in the %APPDATA%, and then it looks recursively for all `nuget.config` files from the root drive down to the `.nuget` dir in project's root dir.

What's possible is that you have this unwanted feed somewhere in the `nuget.config` file, but it's not marked as `enabled` - during nuget convert we take all feeds and ignore the `enabled` / `disabled` stuff

colinbull:
Ahhh, perfect.. the feed is indeed sitting as disabled, in the root `NuGet.Config`. I have no idea how that got there as the machine doesn't exist as far as I can tell. But removing that makes things run much smoother. 

Thanks.

forki:
@colinbull any chance you could retry with the latest Paket?

[...]

PR #720, opened 2015-03-24
Convert ignore disabled feeds

theimowski:
#719


forki:
cool

Awesome support Tomasz. As always.

theimowski:
I do that for the GIFs ;)

PR #722, opened 2015-03-25
Fixes re-ordering problem with files

colinbull:
I have had problems in a number of projects where I have re-ordered the files in the project, for one reason or another. However when I run `paket update` there are certain situations where this order isn't maintained. For example,

Before running `paket update`
original-ordering-officeprovider-microsoft visual studio

After running `paket update`

afterupdate-ordering-officeprovider-microsoft visual studio

This fix prevents this re-ordering from happening by retaining the paket Compile and Content elements in place when `ProjectFile.UpdateFileItems`. I'm not a 100% sure this is the correct fix, but it was the most simply thing I could see to do with out changing `UpdateFiles` significantly.

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -177,6 +177,19 @@ type ProjectFile =
                 | None  ->
                     let firstNode = fileItemsInSameDir |> Seq.head 
                     firstNode.ParentNode.InsertBefore(libReferenceNode, firstNode) |> ignore
+        
+        let paketNodes = 
+            this.FindPaketNodes("Compile")
+            @ this.FindPaketNodes("Content")
+           
+        //remove uneeded files
+        for paketNode in paketNodes do
+            match getAttribute "Include" paketNode with
+            | Some path ->
+                if not(fileItems |> List.exists (fun fi -> fi.Include = path))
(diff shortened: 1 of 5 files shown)


src/Paket.Core/ProjectFile.fs
```diff
@@ -129,8 +129,6 @@ type ProjectFile =
             node.ParentNode.RemoveChild(node) |> ignore
 
     member this.UpdateFileItems(fileItems : FileItem list, hard) = 
-        this.DeletePaketNodes("Compile")
```
forki: we would never deinstall thing if we don't delete.

colinbull: Humm, So the `InstallProcess.InstallIntoProjects` also removes stuff? I guess by virtue of the fact that if it isn't included in that file list then it shouldn't be there? Correct?

forki: the idea was: remove all paket stuff from a fsproj file and add it again. since we can't know if someone remove a package.

forki:
mhm. I think it will need a bit of testing from my side. Will try to look into it tomorrow.

forki:
so this only applies to the content file items, right?

forki:
thanks. let's see if it works.

Issue #729, opened 2015-03-27
Paket creates corrupt package (zip)


Vidarls:
With a very specific file (that I am not allowed to put here) Paket.pack creates nupkg that give crc error when unpacked. Replacing that file after packing manually with somethign like 7 zip fixes the archive.

Simplest case I have been able to reproduce is file type template with single file content.

Are there any way I can get debug data to help figuring out this without publishing stuff I am not allowed to make public?

forki:
I assume the only thing is to trust me and use email or something.

That said: maybe it's time to upgrade paket to newer .NET framework and use
the framework stuff to zip files.

Vidarls:
Sent you an e-mail with the simplest possible repro. (and now everyone can create wonder what the super-secret file is..)

![mystery](http://1.bp.blogspot.com/-4N2wCeBk-6I/Tq948gW8gFI/AAAAAAAABBo/ZOCbxuocXdw/s1600/mystery.gif)

forki:
IT IS A MYSTERY INDEED!
I got no mail.

forki:
I can reproduce this bug with latest Paket, but #732 seems to fix it.
At least I don't get the CRC error any more. But #732 still has the issue that the package explorer can't find the metadata and I don't know why.

forki:
should be fixed in 0.39 alpha

Vidarls:
I can confirm that my original problem is now fixed

PR #732, opened 2015-03-28
Use .NET 4.5's version of ZipArchive - references #729

forki:
This chance has possibly big implications.

Please review it and then I can create an prerelease version for broader testing.

/cc @theimowski @mexx @mavnn @Vidarls

### DIFF
--- a/src/Paket.Core/NuGetV2.fs
+++ b/src/Paket.Core/NuGetV2.fs
@@ -5,7 +5,7 @@ open System
 open System.IO
 open System.Net
 open Newtonsoft.Json
-open Ionic.Zip
+open System.IO.Compression
 open System.Xml
 open System.Text.RegularExpressions
 open Paket.Logging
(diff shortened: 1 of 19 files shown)


src/Paket.Core/NuGetV2.fs
```diff
@@ -331,15 +330,8 @@ let ExtractPackage(fileName:string, targetFolder, name, version:SemVerInfo) =
...
-            use zip = ZipFile.Read(fileName)
             Directory.CreateDirectory(targetFolder) |> ignore
-            for e in zip do
-                try
-                    e.Extract(targetFolder, ExtractExistingFileAction.OverwriteSilently)
-                with
-                | :? Ionic.Zip.BadCrcException as exn -> 
-                    traceWarnfn "Bad Crc during unzipping %s in %s %A: %s" e.FileName name version exn.Message 
-                | exn -> failwithf "Error during unzipping %s in %s %A: %s" e.FileName name version exn.Message 
+            ZipFile.ExtractToDirectory(fileName, targetFolder)
```
theimowski: Is overwrite silently the default behavior? And how about Crc exn? Not thrown anymore?

theimowski:
Lgtm

src/Paket.Core/NuGetV2.fs
forki: not sure.

forki:
@mavnn most things work, but nuget package explorer tells me about missing manifest in generated package. any ideas?

forki:
I tried to diff the packages, but I don't see any differences. I'm stuck.

PR #733, opened 2015-03-28
implement alternative fallback download of paket when github is failing

Stift:
With the ongoing outage/difficult connectivity of github we need a solution to run our builds. The current situation is that the most builds only contain the paket.bootstrapper.exe. When this fails downloading from github, like it happened the last days from time to time the build stops.
Therefore I introduced a fallback that downloads the nuget.exe and make a `nuget install paket` to extract the paket.exe and copy it to the desired target location.

### DIFF
--- a/src/Paket.Bootstrapper/Program.cs
+++ b/src/Paket.Bootstrapper/Program.cs
@@ -66,10 +66,8 @@ namespace Paket.Bootstrapper
                     using (WebClient client = new WebClient())
                     {
                         var releasesUrl = "https://github.com/fsprojects/Paket/releases";
-						
-                        client.Headers.Add("user-agent", "Paket.Bootstrapper");
-                        client.UseDefaultCredentials = true;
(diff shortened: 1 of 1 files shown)


forki:
Wow. that sounds like a good idea

forki:
I assume this will also fix the issue with broken builds on Travis when we hit the github API limit

forki:
Thanks for helping to do this:

truck

Stift:
No problem. It saves my life as well.
When we switch to .NET 4.5 in the future we can directly download and unzip it without using nuget.exe.

forki:
Oh. If you are interested you can try to do this. I plan to merge #732 soon, so upgrading the bootstrapper would be ok.

Stift:
Ok. I make it a separate branch.

Stift:
Have it ready. To what branch should the pr go? I saw that you have the `zip` branch for that.

forki:
please send it to master

PR #734, opened 2015-03-28
Boostrapper download without nuget exe and switch bootstrapper to .net 4.5

Stift:
The previous approach for the nuget fallback for the bootstrapper used the nuget.exe to accomplish it tasks.
I changed the fallback behaviour for the bootstrapper with the following new features
- change project version to .Net 4.5 (to unzip via Framework)
- do not use nuget.exe anymore
- download package directly via nuget API
- use evaluated parameters for prerelease or specified version

### DIFF
--- a/docs/content/bootstrapper.md
+++ b/docs/content/bootstrapper.md
@@ -3,10 +3,12 @@
 Downloads the latest stable paket.exe from github.com.
 
     [lang=batchfile]
-    $ paket.bootstrapper.exe [prerelease|version]
+    $ paket.bootstrapper.exe [prerelease|version] [--prefer-nuget]
 
 Options:
 
   `prerelease`: Downloads the latest paket.exe from github.com and includes prerelease versions.
 
   `version`: Downloads the given version of paket.exe from github.com.
+
+  `--prefer-nuget`: Downloads the given version of paket.exe from nuget.org instead of github.com.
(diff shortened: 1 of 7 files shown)


forki:
very cool.
could you please add a --prefer-nuget switch, which changes the order?
I'm sure @vasily-kirichenko would like it, since sometime github download is slow.

Stift:
Very nice idea. Will do that.

Stift:
Ok. I introduced now the parameter `--prefer-nuget` which prefers nuget over github. 
Default behaviour: github (fallback: nuget)
`--prefer-nuget` behaviour: nuget (fallback: github)

I hope this pleases all the needs.
igbe3a8tavn0q

Stift:
Just found a bug in version parsing from nuget. Will provide a fix in a minute

Stift:
There was an issue with the SemVer stuff. Should be fixed by now

forki:
cool thanks. I just have to wait a bit until we have the other .NET 4.5 stuff ready.

Stift:
Ok. I'll wait until then with other improvements, I'll have in my pipeline.

forki:
Thanks. It's now part of 0.39 alpha.

Stift:
Very nice. You are welcome.

PR #742, opened 2015-03-31
Fixing #729 and #732

### LINKED ISSUE #729: Paket creates corrupt package (zip)

Vidarls:
It seems moving to the .net 4.5 zip packaging handles paths differently. Having a leading `/` on paths added a "no-name" folder in the archive (at least it looked like that in 7 zip archive browser). Effectively putting metadata in a separate sub folder from the rest of package content.

Removing leading `/` in paths did at least produce valid package in windows, and no tests are failing on my machine.

Hopefully this will fix #729 and #732

### DIFF
--- a/src/Paket.Core/NupkgWriter.fs
+++ b/src/Paket.Core/NupkgWriter.fs
@@ -10,7 +10,7 @@ open System.Xml
 
 let nuspecId = "nuspec"
 let corePropsId = "coreProp"
-let contentTypePath = "/[Content_Types].xml"
+let contentTypePath = "[Content_Types].xml"
 
 let contentTypeDoc fileList = 
     let declaration = XDeclaration("1.0", "UTF-8", "yes")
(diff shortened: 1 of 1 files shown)


Vidarls:
@forki can you check if this fix works for you too?

forki:
of course. will do this evening.

mavnn:
Good catch :)

forki:
image

it seems the package explorer is happy now, but Nuget.org is not accepting the package

forki:
it doesn't show any contents

Vidarls:
Strange. I successfully tested yesterday by running nuget.exe install setting source to local folder.

Easter holidays are starting for me now, so I doubt I'll find the time to investigate further until next week. (That should not discourage others from trying their luck though..)

src/Paket.Core/NupkgWriter.fs
```diff
@@ -117,7 +117,7 @@ let nuspecDoc (info:CompleteInfo) =
     optional.Dependencies |> buildDependenciesNode
     XDocument(declaration, box root)
 
-let corePropsPath = sprintf "/package/services/metadata/core-properties/%s.psmdcp" corePropsId
+let corePropsPath = sprintf "package/services/metadata/core-properties/%s.psmdcp" corePropsId
```
forki: I wonder if this is correct.

forki: ok that doesn't change anything a5fdfde still not working

forki:
image

so it seems the Paket.Core package has content. Only Paket package (which only contains exe) is empty.

PR #752, opened 2015-04-07
Only pick up project output files that exactly match assembly filename. Fixes #750

### LINKED ISSUE #750: An item with the same key has already been added: Project with a project dependency that shares the same assembly name prefix
We receive the following error when packing certain solutions:
```
Paket failed with:
        An item with the same key has already been added.
```
This can be reproduced minimally by packing the following solution:

* Project FooBar
    * Builds FooBar.dll
    * Has basic paket.template file (type is project, id is FooBar, description is "blah")
    * Has project dependency on FooBar.Utils (below)
* Project FooBar.Utils
    * Builds FooBar.Utils.dll
    * No paket.template file

We have debugged the latest code and believe this is because of a bug in the way output files are gathered for a project in PackageMetaData.fs, around line 155:
```
fi.Directory.GetFiles(fi.Name.Replace(fi.Extension,"") + ".*")
```
This will incorrectly pick up FooBar.Utils.dll as an output of project FooBar as it matches the wildcard expression 'FooBar.*'. We then add FooBar.Utils.dll again later to the files for this package as it is a project dependency.

The error we see is thrown when we try to create the zip package file and add the second FooBar.Utils.dll to the archive at the same path as the previous one. The error is thrown from line 206 of NupkgWriter.fs:
```
zipFile.AddFile(source, targetFileName.Replace(" ", "%20")) |> ignore
```
We would suggest that this problem will occur whenever a project with a paket.template file references another project and they share an assembly name prefix (FooBar. in this case).

The full error and stacktrace is below:
```
Paket failed with:
        An item with the same key has already been added.
StackTrace:
     at Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](Result`1 res)
   at Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronously[a](CancellationToken token, FSharpAsync`1 computation, FSharpOption`1 timeout)
   at Microsoft.FSharp.Control.FSharpAsync.RunSynchronously[T](FSharpAsync`1 computation, FSharpOption`1 timeout, FSharpOption`1 cancellationToken)
   at Paket.PackageProcess.Pack(DependenciesFile dependencies, String packageOutputPath, FSharpOption`1 buildConfig, FSharpOption`1 version, FSharpOption`1 releaseNotes)
   at Paket.Program.clo@198-21.Invoke(ArgParseResults`1 results)
   at Paket.Program.processWithValidation[T](FSharpFunc`2 validateF, FSharpFunc`2 commandF, Command command, String[] args)
   at Paket.Program.processCommand@59-1.Invoke(Command command, String[] args)
   at <StartupCode$paket>.$Paket.Program.main@()
```

Ian2020:
Fix for #750. Ideally a unit test around this could be added but it's a bit beyond my F# skill at the moment. PackageProcessSpecs.fs looks like a likely spot to add something...

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -159,8 +159,10 @@ let findDependencies (dependencies : DependenciesFile) config (template : Templa
 
         let additionalFiles =
             fi.Directory.GetFiles(fi.Name.Replace(fi.Extension,"") + ".*")
-            |> Array.filter (fun f -> [".xml"; ".dll"; ".exe"; ".pdb"; ".mdb"] |> List.exists ((=) (f.Extension.ToLower())))
-        
+            |> Array.filter (fun f -> 
+                let sameFileName = (Path.GetFileNameWithoutExtension f.Name) = (Path.GetFileNameWithoutExtension fi.Name)
+                let validExtension = [".xml"; ".dll"; ".exe"; ".pdb"; ".mdb"] |> List.exists ((=) (f.Extension.ToLower()))
+                sameFileName && validExtension)        
         additionalFiles
         |> Array.fold (fun template file -> addFile file.FullName targetDir template) template
     


forki:
thanks. please try 0.38.6

Ian2020:
Works for me, cheers!

PR #755, opened 2015-04-08
Fix for missing content using net4.5 zip

Vidarls:
The reason the project type template worked (`Paket.Core`) and the file type template (`Paket tool`) failed was that project type template yielded target paths (`lib/net45/`) with _trailing slash_. Trailing slashes from file type templates depends on template content.

For reasons beyond my understanding, treating the nupkg as a zip file had no issues, regardless of trailing slash, but nuget would fail to extract a package if trailing slash was not present.

This PR will hopefully fix the missing content issue in #742, I have tested it on Paket and som custom templates, and it seems to work


forki:
thanks. that worked

kermit

Vidarls:
Finally :)

Issue #759, opened 2015-04-08
Enterprise Paket: Build speed and target framework?


Thorium:
My current question is about converting an enterprise application to use Paket:
- Will Paket affect to the build speed? I have over 1000 csproj-files. I can measure if you don't know... But when I build in Visual Studio, Output-window says Paket is holding lock to some files and retry and then success...what is that?
- And target frameworks: it seems that paket adds many conditions to the .csproj-files even if csproj-files already have a specific targetframework. I didn't have NuGet in use in the current enterprise application (but the packages-folder where there), so I created a script that creates all the packages.configs (based on dll-refetences HintPaths) that paket's convert-from-nuget -parameter needs: https://gist.github.com/Thorium/c155e8404dedc7b5610b
But does paket use targetframework-attributes of the package.config file?

forki:
> But when I build in Visual Studio, Output-window says Paket is holding lock to some files and retry and then success...what is that?

VS tries to build in parallel. If you have VS restore enabled then one step of that build is to restore your packages. But we make sure that only 1 paket.exe is active at all times, since they compete to write into `packages` folder. Unfortunately there is no VS event which is called before all projects build.
The only way around this is a pre-script (we use FAKE) which calls paket restore.

> But does paket use targetframework-attributes of the package.config file?

AFAIK: no. /cc @theimowski 

If you want to reduce the number of frameworks, then you currently have to edit the paket.dependencies file manually.

theimowski:
nope, `convert-from-nuget` reads only the packaged id and version:
https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NugetConvert.fs#L48-L52

theimowski:
#609 is related

Thorium:
Some initial testing (with Core-i7/2.7GHz/16GB):
About 150 solutions having total about 1100 projects, and a bit over 80 external packages.
Packages are found from local hard drive (already restored).
Build-script is just msbuild calling to build those solutions.
Original build time (NuGet): 7-8 min
After running Paket convert-from-nuget, everything was ok.
But running the same build as earlier, new build time: about 13-14 min

I can't live with it, can I speed up building solutions somehow?

forki:
> Original build time (NuGet): 7-8 min
> But running the same build as earlier, new build time: about 13-14 min

but what are you doing? Why is paket called at all? you could disable restore during MSBuild.

Thorium:
[...]

Issue #609, opened 2015-02-16
Use framework constraint for paket.dependencies in convert-from-nuget


mausch:
Now that https://github.com/fsprojects/Paket/issues/574 works (thanks!) it would be great if Paket could insert framework constraints in paket.dependencies where appropriate when calling `convert-from-nuget`. 

Otherwise, a new user would still get the same error after `convert-from-nuget` until they figure out that they need to insert those constraints.

forki:
@theimowski what do you think?

theimowski:
can we read that from the packages.config? If so then that shouldn't be a big deal

PR #760, opened 2015-04-09
convert from nuget- read target framework from packages.config file -

theimowski:
references #609 #759 

Not sure if this is enough, couldn't find anywhere info re the `targetFramework` attribute possible values.
Tested on a simple `packages.config` with `net451` requirement

### DIFF
--- a/src/Paket.Core/NugetConvert.fs
+++ b/src/Paket.Core/NugetConvert.fs
@@ -222,7 +222,7 @@ let createPackageRequirement (packageName, version, restrictions) sources depend
        VersionRequirement = VersionRequirement(VersionRange.Exactly version, PreReleaseStatus.No)
        Sources = sources
        ResolverStrategy = ResolverStrategy.Max
-       Settings = { InstallSettings.Default with FrameworkRestrictions = restrictions } //Requirements.parseRestrictions
+       Settings = { InstallSettings.Default with FrameworkRestrictions = restrictions }
        Parent = PackageRequirementSource.DependenciesFile dependenciesFileName }
 
 let createDependenciesFileR (rootDirectory : DirectoryInfo) nugetEnv mode =
--- a/src/Paket.Core/NugetConvert.fs
+++ b/src/Paket.Core/NugetConvert.fs
@@ -44,10 +44,16 @@ type CredsMigrationMode =
 /// Represents type of NuGet packages.config file
 type NugetPackagesConfigType = ProjectLevel | SolutionLevel
 
+type NugetPackage = {
+    Id : string
+    Version : SemVerInfo
+    TargetFramework : string option
+}
+
 /// Represents NuGet packages.config file
 type NugetPackagesConfig = {
     File: FileInfo
-    Packages: (string*SemVerInfo) list
+    Packages: NugetPackage list
     Type: NugetPackagesConfigType
 }
     
@@ -176,7 +182,9 @@ module NugetEnv =
                 { File = file
                   Type = if file.Directory.Name = ".nuget" then SolutionLevel else ProjectLevel
                   Packages = [for node in doc.SelectNodes("//package") ->
-                                    node.Attributes.["id"].Value, node.Attributes.["version"].Value |> SemVer.Parse ]}
+                                    { Id = node.Attributes.["id"].Value
+                                      Version = node.Attributes.["version"].Value |> SemVer.Parse
+                                      TargetFramework = node |> getAttribute "targetFramework" } ]}
                 |> ok 
             with _ -> fail (NugetPackagesConfigParseError file)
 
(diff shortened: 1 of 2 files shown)


src/Paket.Core/NugetConvert.fs
```diff
@@ -209,12 +217,12 @@ module ConvertResultR =
...
           SolutionFiles = solutionFiles }
 
-let createPackageRequirement packageName version sources dependenciesFileName = 
+let createPackageRequirement (packageName, version, restrictions) sources dependenciesFileName = 
      { Name = PackageName packageName
        VersionRequirement = VersionRequirement(VersionRange.Exactly version, PreReleaseStatus.No)
        Sources = sources
        ResolverStrategy = ResolverStrategy.Max
-       Settings = InstallSettings.Default
+       Settings = { InstallSettings.Default with FrameworkRestrictions = restrictions } //Requirements.parseRestrictions
```
forki: what's about this comment?

theimowski: shouldn't be there, fixing

forki:
@Thorium interested in beta-testing this?

Thorium:
I'll test...

Thorium:
If you have multiple projects having package.config files, some of them use .Net 4.5, some Portal, some Silverlight, etc.
And you use technologies like Castle-IoC or Rx, that are used in many frameworks.
Now it just parses one of the frameworks and forgets other ones.
``
nuget Castle.Core 3.3.0 framework: sl50
``
This is wrong, I would say this is not ready yet.
But what should actually happen in multi-case?
a) remove the target totally
b) somehow include list of different targets
c) add multiple lines to paket.dependencies and paket.lock

forki:
I'm voting for a)

theimowski:
It forgets the others because it takes only the one with highest version of the package - there's a yellow warning when you have multiple versions for same package

Thorium:
I have same versions, but still:
c:\MyApp\UI\package.config:
    <package id="Rx-Core" version="2.2.5" targetFramework="sl50" />
c:\MyApp\Logics\package.config:
    <package id="Rx-Core" version="2.2.5" targetFramework="net451" />
result of c:\MyApp\paket.dependencies is:
    nuget Rx-Core 2.2.5 framework: sl50

theimowski:
Ok so Then in case of multiple different targetframework should we omit the framework restriction?

Thorium:
I also vote for a)
as it is simple and clear solution.

theimowski:
ok so here it comes, rebased on master - @Thorium is this acceptable now?

forki:
LGTM

Thorium:
Thanks Tomasz!

theimowski:
You're welcome! ;)

Thorium:
Ok, one more problem by this conflicts of nuget targetframework and the real TargetFrameworkVersion.

If I make a new C#-class library and add reference to for example NServiceBus.Log4Net, by-default Visual Studio will make packages.config have:

        <package id="NServiceBus.Log4Net" version="1.0.0" targetFramework="net45" userInstalled="true" />

Then if people update the .NET-version of their projects, by re-targetting to .NET 4.5.1, Visual Studio will modify the .csproj-file to have ``<TargetFrameworkVersion>v4.5.1</TargetFrameworkVersion>``  but the packages.config is not updated. The project still builds and works ok.

But now, if I convert to Paket, it seems to be ok, but the build will fail. And the reason is that Paket will add to .csproj-file only a condition:

        <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And $(TargetFrameworkVersion) == 'v4.5'">

...which won't hit, because of the conflicting 4.5 vs 4.5.1.

How could these mismatches between csproj TargetFrameworkVersion and nuget targetFramework detected?

forki:
how is the paket.dependencies file for this case?
The idea is that if we have targetframework 4.5 then Paket should install >= 4.5
So the bug might be elsewhere

Thorium:
Generated paket.dependencies (Paket 0.40.0):

        source https://www.myget.org/F/particular
        
        nuget log4net 2.0.0 framework: net45
        nuget NServiceBus 5.0.0 framework: net45
        nuget NServiceBus.Log4Net 1.0.0 framework: net45

and paket.lock:

        NUGET
          remote: https://www.nuget.org/api/v2
          specs:
            log4net (2.0.0) - framework: net45
            NServiceBus (5.0.0) - framework: net45
            NServiceBus.Log4Net (1.0.0) - framework: net45
              NServiceBus (>= 5.0.0 < 6.0.0)
              log4net (>= 2.0.0 < 3.0.0)

forki:
we should generate `>= net45` as described in http://fsprojects.github.io/Paket/nuget-dependencies.html#Framework-restrictions

care to send a PR?

Thorium:
Actually, if we have restriction ">= net45" and ">= net451" then the later one containing the another should be enough. So minor version conflicts could be just merged. But "net45" and "net40-client" would be better to not be merged.

forki:
we already have a function which can do this. see https://github.com/fsprojects/Paket/blob/master/tests/Paket.Tests/DependencySetSpecs.fs#L28

Thorium:
Great, I was thinking of some kind of difficult folding, but I don't have to do it then!
Added PR #768.

PR #764, opened 2015-04-10
Better error message for missing paket.dependencies

TeaDrivenDev:
I created a new project, installed Paket via NuGet (it was even up!) and then wanted to install packages via `paket add nuget ....`, which resulted in the error

    Paket failed with:
        Could not find paket.dependencies

That just leaves me hanging, which is not very user friendly. I suppose there is a good reason why it doesn't simply create `paket.dependencies` when it can't find it, but it should at least tell me what I can/need to do about that.

I have updated the error message to mention `paket init`, so the user has a way to proceed.

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -34,7 +34,7 @@ type Dependencies(dependenciesFileName: string) =
                 let parent = dir.Parent
                 if parent = null then
                     if withError then
-                        failwithf "Could not find %s" Constants.DependenciesFileName
+                        failwithf "Could not find '%s'. To use Paket with this solution, please run 'paket init' first." Constants.DependenciesFileName
                     else 
                         Constants.DependenciesFileName
                 else


forki:
very cool. Thanks.

isaacabraham:
Could it run the init command automatically if no dependencies file is found?

forki:
it probably could, but should show a warning.

@TeaDrivenDev interested?

theimowski:
not sure if that's a good idea - git follows similar pattern - no git command allowed in a dir untill `git init` is called

isaacabraham:
@theimowski yeah, that's true. I'm just thinking that probably 99% of the time people see that error, the very next thing they'll do is call ```paket init``` - so why not just do it for them.

theimowski:
still not convinced, one would invoke the `init` once per repo, shouldn't be such a pain. However, if you feel it's really necessary then go on

isaacabraham:
I'm not convinced either - that's why I said "could it" rather than "it should" :-)

PR #765, opened 2015-04-10
Implemented GetHashCode() for NugetDownloadStrategy.SemVer

TeaDrivenDev:
There was a warning in the build about `SemVer` overriding `Equals()`, but not `GetHashCode()`, which can be a bit irritating, so I asked ReSharper to provide an implementation.


forki:
:lollipop:

PR #768, opened 2015-04-10
Target framework: Added >= and made warnings more clear

Thorium:


src/Paket.Core/NugetConvert.fs
```diff
@@ -184,7 +184,8 @@ module NugetEnv =
                   Packages = [for node in doc.SelectNodes("//package") ->
                                     { Id = node.Attributes.["id"].Value
                                       Version = node.Attributes.["version"].Value |> SemVer.Parse
-                                      TargetFramework = node |> getAttribute "targetFramework" } ]}
+                                      TargetFramework = 
+                                        node |> getAttribute "targetFramework" |> Option.map (fun t -> ">= " + t) } ]}
```
theimowski: so this is actually the only change necessary? i.e. we don't need to call the `optimizeRestrictions` function?

Thorium: Yes. I think that the optimization happens later anyways. :-)

src/Paket.Core/NugetConvert.fs
```diff
@@ -228,21 +229,35 @@ let createPackageRequirement (packageName, version, restrictions) sources depend
...
-    
-    let allVersions =
+
+    let allVersionsById =
 
         nugetEnv.NugetProjectFiles
         |> Seq.collect (fun (_,c) -> c.Packages)
         |> Seq.groupBy (fun p -> p.Id)
+
+    let allVersionsByName =
```
theimowski: the `allVersionsByName` and `allVersionsById` are bit misleading, actually id = name for the packages. maybe we could just extract the set of all versions and set of all targetFramwork seperately?

src/Paket.Core/NugetConvert.fs
```diff
@@ -228,21 +229,35 @@ let createPackageRequirement (packageName, version, restrictions) sources depend
...
+                            name    
+                            (versions |> Seq.map string |> Seq.toList)    
+    let allVersions =
+        allVersionsById
         |> Seq.map (fun (name, packages) -> name, packages |> Seq.map (fun p -> p.Version, p.TargetFramework) |> Seq.distinct)
         |> Seq.sortBy (fun (name,_) -> name.ToLower())
     
     for (name, versions) in allVersions do
-        if Seq.length versions > 1 
+        let targets = versions |> Seq.choose(fun (p,target) -> target)
```
theimowski: how about here if we have say `[ "2.0", Some "net40" ; "2.1", Some "net40" ]` ? TargetFramework is the same but, the warning will be output anyways

Thorium: Yes, my problem was that when I get over a screen full of long warning messages not-in-line (because package name length in the begining varies) and containing something like ["1.2.10" Some(">= net45"); "1.2.10", Some(">= net40"); "2.0.0", Some(">= sl5")] then all the information of the more important matter (of version conflict) was hidden in the mess.

theimowski: What I meant is that in this specific case a warning will be displayed even though there is a single `targetFramework`. Regarding console log - yeah we can make it more readable

Thorium: Could be some kind of distinct. But if you take distinct and you have both target and version conflict, then again it would hide some details. And maybe target framework should be info only, not warning, as it is not so bad as version conflict.

theimowski: ok I understand, but then again the warning below could be like `Package X is referenced multiple times with different target frameworks : [Some "net40"; Some "net40"]`  - which I would find misleading. 
Apart from that it looks ok.

forki:
ready to merge?

theimowski:
I guess there's this one thing that could be corrected in my last code comment - everything else seems fine

Thorium:
Fixed that, and some code clean-up also.

forki:
:heart:

theimowski:
:+1:

Issue #597, opened 2015-02-11
paket reformat my app/web.config files while changing assembly redirects


dnauck:
When using the [Redirects option](http://fsprojects.github.io/Paket/dependencies-file.html#Redirects-option) paket reformats my app.config and web.config files on every operation (restore, install, add, ...)

Especially removing blank lines i added for easier reading.

isaacabraham:
Mmmm. This is because we essentially overwrite the file with a new one - this isn't unsurprising as we use XDocument rather than raw string manipulation

dnauck:
Maybe this helps: [PreserveWhitespace option](http://blogs.msdn.com/b/charlie/archive/2008/09/29/linq-farm-preserving-formats-with-linq-to-xml.aspx)

forki:
could you please check if the change helps?

PR #769, opened 2015-04-11
Improve config file formatting with indented binding redirects

Stift:
unitil now paket puts all binding redirects in one line which is very annoying when some binding is changing. With this pr we should improve the formatting in the config file and make it easier for the users to see what changes when they commit the changed config files.

### DIFF
--- a/src/Paket.Core/BindingRedirects.fs
+++ b/src/Paket.Core/BindingRedirects.fs
@@ -1,6 +1,8 @@
 ﻿module Paket.BindingRedirects
 
 open System
+open System.Text
+open System.Xml
 open System.Xml.Linq
 open System.IO
 open System.Reflection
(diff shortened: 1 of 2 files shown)


forki:
Cool. 
There was already an issue for this somewhere. @dnauck

forki:
what about already existig xml. do we reformat that?

Stift:
Yes, but only the part in the runtime/configuration/assemblyBinding. Any problem with that?
The root cause of the problem is, that we load the existing config with PreserveWhitespaces which prevents reformatting on adding new XElements.

forki:
Ok if we only reformat assembly bindings then it should be fine.

@agross @dnauck?

forki:
owl

agross:
:+1:

isaacabraham:
Lovely

Stift:
line commits with +10k are a little bit strange when unexperienced devs see that. hope it helps also others.

dnauck:
@forki this was issue #597 

I'll test this new stuff today.

PR #770, opened 2015-04-12
Fixed home path detection

Stift:
I discovered a bug when normalizing local nuget path the second parameter is always "rooted" means that it got a leading slash. Therefore we get back a rooted home instead of replacing "~" with the home dir.


forki:
thanks.

PR #771, opened 2015-04-13
Uses 'PAKET.VERSION' environment variable for bootstrap requests

devboy:
When the environment variable `PAKET.VERSION` exists `paket.bootstrap.exe` will download that version when no argument is given: 
```
~/Desktop env PAKET.VERSION=0.38.6 mono paket.bootstrapper.exe
Version 0.38.6 requested.
Paket.exe 0.38.6.0 is up to date.
```

When a version argument (or prerelease flag) is passed to `paket.bootstrap.exe` it will behave as always:
```
env PAKET.VERSION=0.38.6 mono paket.bootstrapper.exe 0.40.1
Version 0.40.1 requested.
Starting download from https://github.com/fsprojects/Paket/releases/download/0.40.1/paket.exe
```

This will simplify pinning the version of paket on local and CI environments. Especially for ProjectScaffold projects where you would have to edit multiple places in the build.cmd and build.sh files.

### DIFF
--- a/docs/content/bootstrapper.md
+++ b/docs/content/bootstrapper.md
@@ -12,3 +12,7 @@ Options:
   `version`: Downloads the given version of paket.exe from github.com.
 
   `--prefer-nuget`: Downloads the given version of paket.exe from nuget.org instead of github.com.
+
+Environment Variables:
+
+  `PAKET.VERSION`: The requested version can also be set using this environment variable. Above options take precedence over the environment variable 
--- a/src/Paket.Bootstrapper/Program.cs
+++ b/src/Paket.Bootstrapper/Program.cs
@@ -145,6 +145,7 @@ namespace Paket.Bootstrapper
                 if (args[0] == PrereleaseCommandArg)
                 {
                     ignorePrerelease = false;
+                    latestVersion = "";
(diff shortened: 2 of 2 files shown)


forki:
could you please add a bit of documentation for this?
Thanks

devboy:
Will do.
`PAKET.VERSION` as env variable is fine?
Or would you prefer something like `PAKET.BOOTSTRAP.VERSION` ?

forki:
PAKET.VERSION is ok

devboy:
Let me know if you think this is sufficient information.

devboy:
Just noticed that this might not be expected bahviour:

```$ env PAKET.VERSION=0.38.6 mono paket.bootstrapper.exe prerelease
Prerelease requested. Looking for latest prerelease.
Starting download from https://github.com/fsprojects/Paket/releases/download/0.38.6/paket.exe```

devboy:
Okay, when given `prerelease` option the env var is ignored. So everything will behave as it did previously and according to the documentation.

forki:
cool. thanks.

Stift:
Very nice idea. :+1:

PR #773, opened 2015-04-13
Fixing wrong local path detection with 'paket install'

Stift:
Assume you have the following setup.
A Nuget source `Externals\NugetStore` (i.e. in your repo a local folder is used for some nuget packages) and you start e.g. your `paket install` in any subdirectory (e.g. Externals). Paket discovers correctly the `paket.dependencies` file. But it then fails to detect the folder correctly because it tries to combine the relative folder from your start directory (and it tries to look for `Externals\Externals\NugetStore`) and not in relative to the location paket.dependencies file.
Other paket commands never had any problems with that.


forki:
It's failing on TravisCI. Is that related?

Stift:
it's linux/windows related. I have to adapt some unit tests.

Stift:
Ok. The Travis build is fine now. Should I also update the documentation? This is not clearly saying that relative pathes are also allowed. I just tried it out and it worked.

forki:
Yes. Please update the docs.

Stift:
From my side everything seems to be okay.

forki:
Cool. Will look at it tomorrow. Thanks a lot.

PR #787, opened 2015-04-20
Fixes img src attribute

mariusschulz:
… so that the GROSSWEBER logo shows up correctly ;-)

### DIFF
--- a/docs/content/testimonials.md
+++ b/docs/content/testimonials.md
@@ -10,7 +10,7 @@ We love what we do and with love comes passion. This passion drives us to active
 
 [@agross](https://github.com/agross), [GROSSWEBER](http://grossweber.com/en), Germany/Leipzig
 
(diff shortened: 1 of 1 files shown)


forki:
mhm it worked for me before the change. now it doesn't anymore

forki:
image

mariusschulz:
Ah, I looked at the file here on GitHub. Check out the relative paths.

mariusschulz:
Here's the link: https://github.com/fsprojects/Paket/blob/master/docs/content/testimonials.md

forki:
mhm. yes the file structure is different.

maybe we should change that instead. but I'll revert the commit for now in order to fix the link in the online docs

mariusschulz:
Right, that makes sense. Sorry, only noted the missing image here.

PR #788, opened 2015-04-20
Protect paket.exe from incomplete github downloads

dnauck:
Currently the github HTTP response writes directly to the target paket.exe, but the the download fails until it completes it left a defect paket.exe. So lets download to a temporary file and only copy it to the target location if the download was successful.

This is the same behavior as the nuget download strategy.


forki:
yay

Issue #790, opened 2015-04-21
Updating the bootstrapper/paket.targets


dnauck:
Currently we don't have updates or update notifications for the bootstrapper.

We could either:
1. check for the latest release when running the bootstrapper and display a notification when a new release is available
2. check for the latest release when running the bootstrapper and auto-update its executable
   1. by Shadow Copying Assemblies 
   2. move the locked bootstrapper.exe, download a new bootstrapper and cleanup on next run
3. let paket.exe handle the bootstrapper like a dependency and display it on `paket outdated` or update it on `paket update`, etc.

It should also update the `paket.targets` file when required.

Current disadvantages:
- bootstrapper is updated on every paket.exe release even if there was no changes

matthid:
Seriously wondering why this wasn't suggested earlier.
I'm voting for option "3" which seems to be the easiest to implement.
IMHO we should not complicate the bootstrapper if we don't really need to. For example "1" could be integrated in the paket startup code instead or as suggested in 'paket outdated' (which I would prefer).

Stift:
I have thought about it earlie and thought about `paket.bootstrapper.exe --self`
I came to that point when github had problems some time ago and we implemented an alternative nuget download. The only problem I had so far with that is that in the nuget package is no paket.bootstraper.

Stift:
BTW you can copy it over the current bootstrapper.exe to self update

forki:
so we need a Paket.Boostrapper package?

dnauck:
@Stift what about the ```paket.targets``` file?

How do we know when to run the update command? every time? or is ```paket.exe outdated``` running this?

Stift:
@forki in case we want a self update via nuget we could need one

Stift:
@dnauck have not considered yet the paket.target file, but could be done easily. Do we consider the paket.target as unchangeable and overwrite it without asking? I'm just asking because users could change it and are not aware of a change. 
My proposal is currently supporting ´paket.bootstrapper.exe --self´. I chose that because when we had outages of github it was difficult to even get paket.exe. Should we just show with ´paket outdated´ that the bootstrapper is outdated and show a suggestion.

forki:
https://www.nuget.org/packages/Paket.BootStrapper/

Stift:
Thank you @forki for the package I have included this now.

Stift:
This issue is still open, because the request of @dnauck for updating the paket.targets file is still unresolved.
[...]

PR #791, opened 2015-04-21
"--self" for self update bootstrapper

Stift:
Like suggested by @dnauck in #790 I have created a self update a while ago. This adds an option --self that updates the bootstrapper in place for the github downloads. The nuget download can't currently support the self update because in the nuget package does not contain the bootstrapper.

The current behaviour updates the bootstrapper always, also with the version argument.
After the self-update was successful the bootstrapper gets restarted with the original parameters, but not "--self"

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -306,6 +306,28 @@ Target "Release" (fun _ ->
     |> Async.RunSynchronously
 )
 
+Target "DetectBootstrapperChanges" (fun _ -> 
+    let test = 1
+    let allTags = getGitResult "" "tag -l \"*\""
+    let latestVersion = allTags
+                        |> Seq.map (fun x -> try 
+                                                let ver = SemVerHelper.parse(x)
+                                                Some ver
+                                             with
+                                              | :? Exception -> None)
+                        |> Seq.filter (fun x -> match x with | Some _ -> true| None -> false)
+                        |> Seq.map (fun x -> match x with | Some y -> y | None -> failwith "Should never happen")
+                        |> Seq.filter (fun x -> match x.PreRelease with | Some _ -> false | None -> true)
+                        |> Seq.sort
+                        |> Seq.toList
+                        |> List.rev
+                        |> List.head
+
+    let res = runSimpleGitCommand "" (String.Format("diff --shortstat {0} src/Paket.Bootstrapper/", latestVersion))
+    if not (String.IsNullOrWhiteSpace(res)) then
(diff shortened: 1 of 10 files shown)


forki:
I think we need a bit of docs for this.

matthid:
Definitely another option to consider.
I tried this out on windows and got:
```
$ .paket/paket.bootstrapper.exe prerelease --self
Prerelease requested. Looking for latest prerelease.
Trying self update
Starting download of bootstrapper from https://github.com/fsprojects/Paket/releases/download/1.0.1/paket.bootstrapper.exe
Self update failed. Resetting bootstrapper.
Access to the path 'c:\Projects\Paket\.paket\paket.bootstrapper.exe.old' is denied. (Github)
```
I'm more surprised that we can actually move the file in the first place and that the update works (it just fails while cleaning up the old version). Never noticed that you can rename locked files on windows :).

Also this is a bit unfortunate:

```
$ .paket/paket.bootstrapper.exe --self
Version --self requested.
Trying self update
Starting download of bootstrapper from https://github.com/fsprojects/Paket/releases/download/--self/paket.bootstrapper.exe
Self update failed. Resetting bootstrapper.
The remote server returned an error: (404) Not Found. (Github)
```

Nevertheless we should consider to integrate the bootstrapper update in `paket update` anyway, this will allow us to update existing installations as well (instead of having to update those  manually).

Stift:
@forki got the message. i just wait a little bit for the discussion.
@matthid I'll fix this.

Stift:
I now inlcuded the new paket.bootstrapper nuget package. Therefore we have two ways to update.
I removed the restart behaviour because it will causing trouble when used in scripts

forki:
Cool. We should still change the build script and delete the bootstrapper package before push when we don't have changes.

Stift:
How do we want to detect 'no changes'?

forki:
We always tag after a release. So we would ask git to show diff for
subfolder since last release. I will try to do this next week.

Stift:
I already created a build step to detect the latest tag and get the diff changes. The step is currenly orphaned not yet connected to the build process. I think it could be a good starting point for you.

forki:
I thought about it. Maybe it's better to make that step explicit by adding
something to the release notes or by adding a second release notes
document.

Stift:
That would be good. I always wondered why we publish the same version again and again. The same with the paket.targets document where it is hard to track a version, when we don't put a version identifier in the header of the document.

Issue #776, opened 2015-04-14
Convert-from-nuget & .targets 


petarvucetin:
Looks like .targets import is not getting picked up on conversion.

    <Import Project="..\Solution Files\packages\PostSharp.4.1.10\tools\PostSharp.targets" Condition="Exists('..\Solution Files\packages\PostSharp.4.1.10\tools\PostSharp.targets')" />

     <Import Project="..\Solution Files\packages\AutoMapper.4.0.0-ci1007\tools\AutoMapper.targets" Condition="Exists('..\Solution Files\packages\AutoMapper.4.0.0-ci1007\tools\AutoMapper.targets')" />

forki:
yes that true. we currently don't remove such imports.


@theimowski

petarvucetin:
And there is more junk that some packages install... 


    <Import Project="..\Solution Files\packages\PostSharp.4.1.8-beta\tools\PostSharp.targets" Condition="Exists('..\Solution Files\packages\PostSharp.4.1.8-beta\tools\PostSharp.targets')" />
    <Target Name="EnsurePostSharpImported" BeforeTargets="BeforeBuild" Condition="'$(PostSharp30Imported)' == ''">
        <Error Condition="!Exists('..\Solution Files\packages\PostSharp.4.1.8-beta\tools\PostSharp.targets')" Text="This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://www.postsharp.net/links/nuget-restore." />
         <Error Condition="Exists('..\Solution Files\packages\PostSharp.4.1.8-beta\tools\PostSharp.targets')" Text="The build restored NuGet packages. Build the project again to include these packages in the build. For more information, see http://www.postsharp.net/links/nuget-restore." />
     </Target>

theimowski:
This might be not trivial to detect. Need to find a way how to recognize all such entries

forki:
Yep that's the big issue. We can't reliably detect this. But maybe we can
have heuristic that will work for most important packages.

petarvucetin:
From what I have seen in my csproj files these always follow each other <Import> then <Target> if that helps.

theimowski:
@petarvucetin can you verify that with the newest version?

PR #792, opened 2015-04-21
Convert remove custom import and targets

theimowski:
solves #776
the issue is also mentioned ind #516 

`convert-from-nuget` will look for `Import`s with package id and version in path, and additionally will interactively ask if one wants to remove following `Target` node (if present) 

Couldn't come up with wiser heuristic for this


forki:
thanks.

PR #802, opened 2015-04-29
Paket should keep paket.dependencies as stable as possible

### LINKED ISSUE #780: Paket Update messes with sources in paket.dependencies
Original paket.dependencies:

source myinternalrepo
package..
package...

source officalrepo
package...

After update:

source myinternalrepo
source officalrepo
package..
package...
package...

forki:
Idea:

Instead of serializing the Dependencies file from the already parsed information we always modify the text and reparse it. Maybe not the best style, but it keep comments and everything in place.

fixes #780, #731

This needs a lot of testing. Especially the convert-from-nuget stuff (cc @theimowski)

/cc @matthid, @cjbhaines, @mexx, @inosik , @haf

### DIFF
--- a/src/Paket.Core/DependenciesFile.fs
+++ b/src/Paket.Core/DependenciesFile.fs
@@ -21,6 +21,10 @@ type InstallOptions =
         Redirects = false
         Settings = InstallSettings.Default }
 
+type VersionStrategy = {
+    VersionRequirement : VersionRequirement
+    ResolverStrategy : ResolverStrategy }
+            
 /// [omit]
 module DependenciesFileParser = 
 
@@ -231,7 +235,10 @@ module DependenciesFileParser =
             packages |> List.rev,
             remoteFiles |> List.rev,
             lines
-
+    
+    let parseVersionString (version : string) = 
+        { VersionRequirement = parseVersionRequirement (version.Trim '!')
+          ResolverStrategy = parseResolverStrategy version }
 
 module DependenciesFileSerializer = 
     let formatVersionRange strategy (version : VersionRequirement) : string =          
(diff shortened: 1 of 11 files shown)


src/Paket.Core/DependenciesFile.fs
```diff
@@ -399,7 +428,7 @@ type DependenciesFile(fileName,options,sources,packages : PackageRequirement lis
                                      if NormalizedPackageName p.Name = NormalizedPackageName packageName then 
                                          { p with VersionRequirement = versionRequirement }
                                      else p)
-            DependenciesFile(this.FileName, this.Options, sources, packages, this.RemoteFiles, comments)
+            DependenciesFile(this.FileName, this.Options, sources, packages, this.RemoteFiles, textRepresentation)
```
mexx: Text representation have to be updated in this case.

forki: oups. correct.

theimowski:
I'll do some testing on Friday

src/Paket.Core/DependenciesFile.fs
forki: fixed in 93ca6f8

forki:
huh. I accidentally realigned the whole file. reverting that

forki:
it's in 1.3.0-alpha001

theimowski:
I think `dependenciesFile` is a good candidate for Property-Based round-trip test (parse, format, compare). Would need better tracking of comments/ custom code and similar stuff tho.

forki:
would love to see that.

2015-04-30 21:07 GMT+02:00 Tomasz Heimowski <notifications@github.com>:

theimowski:
changes in `NugetConver` look ok - tested on a simple scenario, accidently also found a bug in the convert logic itself - fix in #804

forki:
so ready to release?

theimowski:
:shipit:

PR #807, opened 2015-05-01
[WIP] Property tests for dependencies file parsing

theimowski:
As per #802 the dependencies file caches all lines for later serializing, so this round-test may not be that useful, however it can surely serve as a safety net for future refactorings.

The test doesn't yet generate all valid dependencies files, but it could already find some pitfalls:
- can have comment line starting with `#`,
 but can't have a hash-comment inline such as `nuget FsCheck #fscheck is cool`
- can not have comments after: github, gist, http dependencies

Didn't fix those for now, instead made a workaround in generating test data

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -1,5 +1,7 @@
 source https://nuget.org/api/v2
 
+nuget FsCheck
+nuget FsCheck.NUnit
 nuget Newtonsoft.Json
 nuget UnionArgParser
 nuget NUnit.Runners
@@ -13,5 +15,3 @@ github forki/FsUnit FsUnit.fs
 github fsharp/FAKE modules/Octokit/Octokit.fsx
 github fsharp/FAKE src/app/FakeLib/Globbing/Globbing.fs
 github fsprojects/Chessie src/Chessie/ErrorHandling.fs
-nuget FsCheck
-nuget FsCheck.NUnit
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -14,3 +14,4 @@ github fsharp/FAKE modules/Octokit/Octokit.fsx
 github fsharp/FAKE src/app/FakeLib/Globbing/Globbing.fs
 github fsprojects/Chessie src/Chessie/ErrorHandling.fs
 nuget FsCheck
+nuget FsCheck.NUnit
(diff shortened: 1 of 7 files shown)


tests/Paket.Tests/Paket.Tests.fsproj
```diff
@@ -261,9 +265,16 @@
     <Compile Include="InstallModel\Penalty\FrameworkConditionsSpecs.fs" />
     <Compile Include="DependencyModel\ProjectDependencySpecs.fs" />
     <None Include="paket.references" />
+    <Content Include="App.config" />
   </ItemGroup>
   <ItemGroup>
     <Reference Include="mscorlib" />
+    <Reference Include="nunit.core">
```
theimowski: had to add those manually - FsCheck.Nunit package does that through powershell

forki: /cc @kurtschelfthout why is this in the powershell script?

paket.dependencies
```diff
@@ -11,4 +11,6 @@ nuget FSharp.Core
 github forki/FsUnit FsUnit.fs
 github fsharp/FAKE modules/Octokit/Octokit.fsx
 github fsharp/FAKE src/app/FakeLib/Globbing/Globbing.fs
-github fsprojects/Chessie src/Chessie/ErrorHandling.fs
\ No newline at end of file
+github fsprojects/Chessie src/Chessie/ErrorHandling.fs
+nuget FsCheck
```
forki: did you add this manually? If you used paket add I hope it would put it to the other nuget statements (in alphabetical order)

theimowski: I used `paket add nuget FsCheck project Paket.Tests`

forki: ok. then I would consider this a bug. I will create a test case and try to fix

tests/Paket.Tests/Paket.Tests.fsproj
kurtschelfthout: On mobile now so don't have a good overview, but think that's wrong, should be reference/content or whatever the canonical way of doing it is. Would you mind opening an issue?

paket.dependencies
forki: should be fixed in 1.3.3

forki:
ready to merge?

theimowski:
Not yet, I'll do some improvements soon and let know when its ready

theimowski:
OK  I think it can be merged for now - added a bit more possibilities to the generator, however still there are some "hardcoded" values. 
Can be a good start when time comes for refactoring `DependenciesFile` parsing

forki:
thx

PR #821, opened 2015-05-08
Nuget Convert should cope with empty version strings

mavnn:
Especially given we add one to the list of required packages if we find nuget.exe


forki:
thanks. released as 1.3.35

Issue #830, opened 2015-05-20
Failing convert from nuget with exception


Stift:
When I convert a solution I get the following exception:

```
...
Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at <StartupCode$Paket-Core>.$ProjectFile.RemoveImportAndTargetEntries@493.Invoke(XmlNode node)
   at Microsoft.FSharp.Primitives.Basics.List.iter[T](FSharpFunc`2 f, FSharpList`1 x)
   at Paket.NuGetConvert.convertProjects@325.GenerateNext(IEnumerable`1& next)
   at Microsoft.FSharp.Core.CompilerServices.GeneratedSequenceBase`1.MoveNextImpl()
   at Microsoft.FSharp.Collections.SeqModule.ToList[T](IEnumerable`1 source)
   at Paket.NuGetConvert.createPaketEnv@334.Invoke(DependenciesFile _arg1)
   at Chessie.ErrorHandling.Trial.fSuccess@63-11[p,r,q](FSharpFunc`2 f, p x, FSharpList`1 msgs)
   at Paket.NuGetConvert.createResult(DirectoryInfo rootDirectory, NugetEnv nugetEnv, CredsMigrationMode credsMirationMode)
   at Chessie.ErrorHandling.Trial.fSuccess@63-11[p,r,q](FSharpFunc`2 f, p x, FSharpList`1 msgs)
   at Chessie.ErrorHandling.Trial.fSuccess@63-11[p,r,q](FSharpFunc`2 f, p x, FSharpList`1 msgs)
   at Chessie.ErrorHandling.Trial.fSuccess@63-11[p,r,q](FSharpFunc`2 f, p x, FSharpList`1 msgs)
   at <StartupCode$Paket-Core>.$PublicAPI.ConvertFromNuget@66.Invoke(Unit unitVar0)
   at Paket.Utils.RunInLockedAccessMode[a](String rootFolder, FSharpFunc`2 action)
   at Paket.Program.clo@255-11.Invoke(ArgParseResults`1 results)
   at Paket.Program.processWithValidation[T](FSharpFunc`2 validateF, FSharpFunc`2 commandF, Command command, String[] args)
   at Paket.Program.processCommand@71-1.Invoke(Command command, String[] args)
   at <StartupCode$paket>.$Paket.Program.main@()
```

forki:
woupsi. NPE - I love them.

[...]

PR #831, opened 2015-05-20
sibling can be null - references #830

theimowski:
that's my guess - @Stift can you verify if this fixes the issue?


Stift:
I can confirm, that this error is solved. :+1:

theimowski:
thanks @Stift for reporting the issue

Stift:
no problem. That's we are here

forki:
Ha, I see you read @LincolnAtkinson's post

Issue #832, opened 2015-05-20
paket convert-from-nuget doesn't handle sources with spaces


amazingant:
Was trying to help out w/ some paket questions on the fsharp slack channel and decided to run `paket convert-from-nuget` on one of my projects; I'm using the same NuGet sources that caused #616.

While paket correctly pulled in all my project dependencies and package sources, it didn't wrap quotes around the sources that contain spaces, warning (yellow text?) `Local Nuget feed doesn't exist: C:\Path Here\Packages.` and then eventually failing with `Couldn't get package details for package <package name> on any of ["C:\Path"; "https://www.nuget.org/api/v2"].` (note the truncated path here).

Running `paket install` results in the same failure message with the truncated path; after manually adding quotes around the path in `paket.dependencies`, running `paket install` again works fine.

forki:
Do you have a sample to reproduce? 

/cc @theimowski maybe he knows where to look

amazingant:
The project I'm working with is an internal one, but I can whip up a new project that does it if it helps?

theimowski:
yeah we don't escape paths in `convert-from-nuget` bear with me, will send a fix soon

amazingant:
@theimowski take your time; I'm not actually depending on this at the moment and it's easy to work around.

theimowski:
no worries, this one was easy

forki:
Xoxo

amazingant:
Well then lucky for you, I found another issue! :D

Also related to `convert-from-nuget`, not related to spaces, github issue on its way when I can verify what's happening.

amazingant:
Closing because @theimowski's change got merged with #833

theimowski:
thanks @amazingant for reporting the issue!

PR #833, opened 2015-05-20
use String.quoted for feeds on convert-from-nuget - references #832

theimowski:


amazingant:
:+1: build off of AppVeyor runs great on my machine with my evil package source :)

forki:
cool.

PR #835, opened 2015-05-20
Observable.OfAsync now fires OnCompleted

TheAngryByrd:
The ``Observable.OfAsync`` operations never fired the OnCompleted event of the observer.  This could cause issues where Observers are waiting for Async events to finish. This change fixes that.


forki:
thx

PR #843, opened 2015-05-28
Pull request/refactor update process

cr7pt0gr4ph7:
## paket restore --only-referenced
As the title says: This pull requests makes it possible to say:
```bash
paket restore --only-referenced
```
which is the same as:
```
paket restore --references-files <all paket.references files that paket install would find>
```

## paket install--only-referenced
Consequently, it also adds:
```
paket install --only-referenced
```
which works analogous to `paket restore --only-referenced`.

## Refactorings
To make such additions easier in the future, `InstallProcess` and `UpdateProcess` were refactored to use the Parameter Object pattern for the different flags (`Force`, `Hard`,  `WithBindingRedirects`).
The naming for the new record types `InstallerOptions` and `SmartInstallOptions` is not optimal, so I'm open for feedback.

## Testing
I've not yet added tests for the new functionality. I'd be thankful if you could point me to the correct files where tests for `InstallProcess` and `UpdateProcess` sould be added.

## Compatibility
`Dependencies`/`PublicAPI.fs` has been changed in a backwards compatible way by adding two new public members. The changes to `InstallProcess` and `UpateProcess` are not backwards compatible, but this is nearly unavoidable due to the quirks of F#.

### DIFF
--- a/src/Paket.Core/AddProcess.fs
+++ b/src/Paket.Core/AddProcess.fs
@@ -14,7 +14,7 @@ let private addToProject (project : ProjectFile) package =
         .AddNuGetReference(package)
         .Save()
 
-let private add installToProjects addToProjectsF dependenciesFileName package version force hard installAfter =
+let private add installToProjects addToProjectsF dependenciesFileName package version options installAfter =
     let existingDependenciesFile = DependenciesFile.ReadFromFile(dependenciesFileName)
     let (PackageName name) = package
     if (not installToProjects) && existingDependenciesFile.HasPackage package && String.IsNullOrWhiteSpace version then 
@@ -24,7 +24,8 @@ let private add installToProjects addToProjectsF dependenciesFileName package ve
             existingDependenciesFile
                 .Add(package,version)
     
-        let lockFile = UpdateProcess.SelectiveUpdate(dependenciesFile,Some(NormalizedPackageName package),force)
+
+        let lockFile = UpdateProcess.SelectiveUpdate(dependenciesFile, Some(NormalizedPackageName package), options.Force)
         let projects = seq { for p in ProjectFile.FindAllProjects(Path.GetDirectoryName lockFile.FileName) -> p } // lazy sequence in case no project install required
 
         dependenciesFile.Save()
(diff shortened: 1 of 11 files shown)


forki:
Cool. You also need to add the parameter to the update command

cr7pt0gr4ph7:
I'll look into the `paket update` command, and see if I can add the flag.
I also think that some code duplication between `InstallProcess` and `UpdateProcess` can be removed, but I would do another pull request for that.

forki:
Probably also needed in add command (just to keep it aligned)

cr7pt0gr4ph7:
The `paket add` command should  also have the `--redirects` flags, like the other commands that install packages.

forki:
Correct. Would love to receive a pull request

cr7pt0gr4ph7:
`paket remove` needs `--only-referenced` and `--redirects`, too.

cr7pt0gr4ph7:
The problem with `Dependencies` is that the parameter lists grow longer and longer, as more flags are added.

forki:
we don't need binary backwards compat. There is only Paket.VisualStudio as consumer at the moment.

cr7pt0gr4ph7:
Okay, then I'll just add more parameters to `Dependencies`, and then look into simplifying the API using the pattern from the code sample above.

cr7pt0gr4ph7:
After looking a bit into adding `--only-referenced` to `paket add`, `paket update` and `paket remove`, I'd say that it makes more sense to do the refactoring of `Dependencies` first, to limit code duplication.

As it currently stands, `paket add --only-referenced` (and `paket remove --only-referenced`) can be simulated using `paket add --no-install` followed by `paket restore --only-referenced`.

This is also a hint that much of the code from `AddProcess`, `RemoveProcess` and so on can unified by separating the required operations into the three phases:
  1. Modify `paket.dependencies` / `paket.references` / `paket.lock` / ...`,
  2. Download packages as required,
  3. Apply changes to `*.csproj` files,
Of which **(2)** (and maybe **(3)**) can be implemented using shared code..

Because the necessary modifications are a bit of another story (and because I don't want this pull request to get huge), I'd vote to merge this pull request first.
The other reason is that I need `paket install --only-referenced` in my productive work, so I'd limit further additions to this pull request to the ones that are required to get it merged.

forki:
Very cool. Need to try some things.

fast

cr7pt0gr4ph7:
Thank you! That was very fast.

forki:
I think this broke the default behaviour.

forki:
i think I found the issue

forki:
ok it's released in 1.7.1

PR #847, opened 2015-05-28
Implement --no-install and --redirects for "paket update".

cr7pt0gr4ph7:
This pull request adds the switch `--no-install` to `paket update`, and handles `paket update --redirects` also for the "single package update" case.
Additionally, the newly introduced `InstallerOptions` and `SmartInstallOptions` have been refactored, such that they match their usage better.


forki:
LGTM. Thanks

PR #853, opened 2015-06-01
Copy paket exe to cwd on init

Bomret:
As discussed in [Paket.Atom issue 14](https://github.com/fsprojects/Paket.Atom/issues/14) Paket should download the paket.bootstrapper into the .paket dir on init. I created a new module _Paket.Releases_ which handles downloading the latest versions of paket.exe, the bootstrapper and paket.targets.


src/Paket.Core/Releases.fs
```diff
@@ -0,0 +1,61 @@
...
+
+let private existsNotOrIsNewer (file:FileInfo) latest =
+    if (not <| file.Exists) then true
+    else
+        let verInfo = FileVersionInfo.GetVersionInfo file.FullName
+        let currentVersion = SemVer.Parse verInfo.FileVersion
+        currentVersion < latest
+
+let downloadLatestVersionOf files destDir =
+    let releasesUrl = "https://api.github.com/repos/fsprojects/Paket/releases";
```
forki: please move this into the Constants module

src/Paket.Core/Releases.fs
```diff
@@ -0,0 +1,61 @@
...
+    use client = createWebClient("https://github.com",None)
```
forki: Also Constants

src/Paket.Core/Releases.fs
```diff
@@ -0,0 +1,61 @@
...
+            |> List.map (fun file -> download latestVersion file client)
+            |> collect
+        
+        ignore downloads
+    }
+
+let downloadLatestBootstrapper environment =        
+    let exeDir = Path.Combine(environment.RootDirectory.FullName, ".paket")
+
+    downloadLatestVersionOf ["paket.bootstrapper.exe"] exeDir
```
forki: I know this is all refactored from existing code, but please put all these constants into the Constants module

src/Paket.Core/Releases.fs
```diff
@@ -0,0 +1,61 @@
...
+
+let downloadLatestBootstrapperAndTargets environment =
```
theimowski: I'd rather inline this and above functions - they seem pretty straightforward

src/Paket.Core/Releases.fs
```diff
@@ -0,0 +1,61 @@
...
+    trial {
+        tracen (sprintf "%A" file)
+
+        do! createDir(file.DirectoryName)
+        let url = sprintf "https://github.com/fsprojects/Paket/releases/download/%s/%s" (string version) file.Name
+        
+        do! downloadFileSync url file.FullName client
+    }
+
+let private existsNotOrIsNewer (file:FileInfo) latest =
```
theimowski: `doesNotExist` ?

theimowski:
apart from the comments, LGTM

Bomret:
I moved the URLs and other stuff into the Constants module and added comments on the functions. @theimowski what exactly did you mean by "inline these and above functions"? Remove the _downloadLatestBootstrapper*_ functions and only call _downloadLatestVersionOf_?

forki:
I think he meant, that you just remove the functions and use the code directly without a separate function

Bomret:
But this would be unnecessary duplication? The way it is now the _downloadLatestVersionOf_ function can be used to easily extend the functionality of the module later for other files.

forki:
Ok. Thanks a lot

Bomret:
And both other functions are used in the publicAPI module. Or do I still get it wrong?

forki:
thx. released in 1.10

theimowski:
Ok nvmd, I've not looked at the call sites for those functions - they just seemed extraordinary at first glance

PR #864, opened 2015-06-08
Some extra documentation for the bootstrapper

Stift:
I updated some of docu for the bootstrapper to provide better info about the internal proceedings of that.


forki:
awesome

Issue #874, opened 2015-06-14
Output does not show in Package Manager Console


ctaggart:
The output is showing up great in normal PowerShell, but does not show up by default in Visual Studio Package Manager Console. For example, it should show `Downloading file`...

![image](https://cloud.githubusercontent.com/assets/80104/8149324/130cd200-1274-11e5-891c-89c39a18896d.png)

![image](https://cloud.githubusercontent.com/assets/80104/8149336/522d96a4-1274-11e5-8885-94a6f4d68943.png)

I'll investigate to figure out why.

ctaggart:
Also related to logging levels, when `stderr` is used, it doesn't look great in hosted PowerShell. 

![image](https://cloud.githubusercontent.com/assets/80104/8149360/40098478-1275-11e5-9ab0-ea21f308becf.png)

May be look into `Write-Warning`.

ctaggart:
I've been stuck on this for way to long now:
![image](https://cloud.githubusercontent.com/assets/80104/8150959/47c103ec-12b3-11e5-8014-fdeaf95a3f5a.png)

The debugging round trip time isn't super fast. Since ArgParser is still on F# 3.0, it requires a `build.cmd MergePowerShell` each time. https://github.com/nessos/UnionArgParser/pull/37 The .dll often stays locked until I close PowerShell. :/

ctaggart:
The first cause of the error message was that trace listeners were not removed from previous runs. I switched over to using observables, so subscriptions are IDisposable. This allowed me to make it further:

![image](https://cloud.githubusercontent.com/assets/80104/8153792/9a6e88b0-12e7-11e5-9e66-593e7ed9362a.png)

It now appears that having a `tracefn` in an `async` causes the problem:
![image](https://cloud.githubusercontent.com/assets/80104/8153803/f1675a2a-12e7-11e5-87e9-63313ef67446.png)

I'll start a pull request, so my changes thus far can be reviewed.

ctaggart:
Fixed by #875.

PR #875, opened 2015-06-15
Paket.PowerShell support for Package Manager Console

ctaggart:
This is for #874 and is ~~not~~ ready ~~yet~~ to be merged. ~~It is **in progress** and ready for review. I could use some help solving the async / trace dilemma.~~ PowerShell requires that the tracing be on the same thread and it ~~appears~~ async puts it on another.

### DIFF
--- a/src/Paket.Core/Logging.fs
+++ b/src/Paket.Core/Logging.fs
@@ -16,9 +16,6 @@ type Trace = {
 /// [omit]
 let event = Event<Trace>()
 
-/// [omit]
-let subscribe callback = Observable.subscribe callback event.Publish
-
 
 /// [omit]
 let tracen s = event.Trigger { Level = TraceLevel.Info; Text = s; NewLine = true }
@@ -101,4 +98,4 @@ let setLogFile fileName =
     else
         if fi.Directory.Exists |> not then
             fi.Directory.Create()
-    subscribe traceToFile
+    event.Publish |> Observable.subscribe traceToFile
--- a/src/Paket.Core/Logging.fs
+++ b/src/Paket.Core/Logging.fs
@@ -19,12 +19,8 @@ let event = Event<Trace>()
 /// [omit]
 let subscribe callback = Observable.subscribe callback event.Publish
 
-/// [omit]
-let monitor = new Object()
-
 
 /// [omit]
-//let tracen s = traceToRegisteredFunctions TraceLevel.Info s true
 let tracen s = event.Trigger { Level = TraceLevel.Info; Text = s; NewLine = true }
 
 /// [omit]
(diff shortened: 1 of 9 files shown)


src/Paket.PowerShell/CmdletExt.fs
```diff
@@ -37,4 +40,17 @@ type PSCmdlet with
...
     member x.SetCurrentDirectoryToLocation() =
-        System.Environment.CurrentDirectory <- x.SessionState.Path.CurrentFileSystemLocation.Path
\ No newline at end of file
+        Environment.CurrentDirectory <- x.SessionState.Path.CurrentFileSystemLocation.Path
+
+    member x.RegisterTrace() =
+        Logging.verbose <- x.Verbose
+        let id = Threading.Thread.CurrentThread.ManagedThreadId
+        Logging.subscribe (fun trace ->
+            if id = Threading.Thread.CurrentThread.ManagedThreadId then
```
ctaggart: This thread ID check appears it will work. Messages in `async` will not print, but also will not error out the program now.

src/Paket.Core/Logging.fs
```diff
@@ -2,82 +2,103 @@
...
+let event = Event<Trace>()
 
-let inline traceToFile (text:string) =
-    match logFile with
-    | Some fileName -> try File.AppendAllLines(fileName,[text]) with | _ -> ()
-    | _ -> ()
+/// [omit]
+let subscribe callback = Observable.subscribe callback event.Publish
 
-let RegisterTraceFunction(traceFunction:Action<string>) =
```
ctaggart: Who was using this? Will `Logging.subscribe` work for them or does this need to be added back (may be call `subscribeWithAction`). It would return a `IDisposable` instead of having `RemoveTraceFunction`.

forki: I assume Paket.VisualStudio is using this. But we can fix breaking changes
there

ctaggart:
I asked around at work about how to get all the messages on the same thread. I was told to look at `Async.SwitchContext`. There is some good related code already [in Utils.fs](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Utils.fs#L347-L438). Hoping that I can capture `SynchronizationContext.Current` instead of the `ManagedThreadId` and use it to put all the messages on the right thread for PowerShell. I'll try it out tonight.

ctaggart:
I spent hours on this, this evening, without any luck. I opened up a stackover question:
http://stackoverflow.com/questions/30859885/post-messages-on-async-thread-to-original-thread-in-f

ctaggart:
I believe this is ready to go. Logging now works correctly in PowerShell and in Package Manager Console. It displays output as it goes.

image

image

I created type `EventSink` and function `processWithLogging` to allow the trace messages to be output on the main PowerShell thread, which is required by PowerShell (according to the error messages I get when I don't do so).

ctaggart:
The feature is `Paket.PowerShell support for Paket Manager Console` and to add support I had to `refacto Logging to use Trace observable`.

forki:
cool. thanks

forki:
maybe it even fixes the logging issues in paket.vs - you never know ;-)

forki:
Did this by any chance changed the exit code behavior? It seems we always get exit code 1 - even on success.

ctaggart:
:/ I'll have a look.

forki:
I mean it's probably something different, but I have no idea at the moment

ctaggart:
Looking at the changes I made to `Program.fs`, I think I see this issue.

ctaggart:
@forki Just to be clear, the exit code wasn't broken. What was broken was the exception not being logged.

forki:
I tried to update Paket.VisualStudio and got:

image

What do I need to do?

ctaggart:
Paket.Logging.event is public. event.Publish can be subscibed to.

forki:
image

I don't get it working. Could you please help me a bit? thx

dungpa:
@forki Can you change `event` to use `DelegateEvent` to use `+=` notation in C#? See https://github.com/fsprojects/VisualFSharpPowerTools/blob/40bd0f678925a4e64c60d09014660e1f03c9c1e0/src/FSharpVSPowerTools.Logic/ShellEventListener.fs#L28

forki:
https://github.com/fsprojects/Paket.VisualStudio/commit/8bb932a16699c84786cc93d8bdf6395f1705f3b4 does this look OK?

Issue #880, opened 2015-06-18
Exclusion syntax in paket.template files


devboy:
It would be nice to be able to exclude file patterns in paket.template files.

Something like:
```
files
    relative/to/template/file ==> folder/in/nupkg
    second/thing/to/pack ==> folder/in/nupkg
    second/thing/**/file.* ==> folder/in/nupkg
    -- third/*
    -- .DS_Store
```

What do you think?

forki:
are you volunteering? ;-)

I think we already have FAKE's Globbing in place, so there must be a relatively easy way to do this.

devboy:
Yeah, if we can agree on syntax ;)

devboy:
```
files
    relative/to/template/file ==> folder/in/nupkg
    second/thing/to/pack ==> folder/in/nupkg
    second/thing/**/file.* ==> folder/in/nupkg
    ! third/*
    ! .DS_Store
```

would also be an idea, more glob-like...

agross:
Looks good! Perhaps we should align the `!` syntax with `.gitignore` and not require the space between e.g. `!` and `third/*`.

devboy:
Just saw your comment @agross ... will change it

mavnn:
I'm with @agross - you can always use ./!stupidfilename.txt if you've named a file starting with a "!" and it makes things consistent with other systems.

mavnn:
In fact: we should probably disallow the space to avoid ambiguity.

devboy:
Put that into the parser ;)

Now trying to wrap my head around the `NupkgWriter` and how to apply the exclusion pattern there....

devboy:
```
files
    relative/to/template/file ==> folder/in/nupkg
    second/thing/to/pack ==> folder/in/nupkg
    second/thing/**/file.* ==> folder/in/nupkg
    !third/*
    !.DS_Store
```

devboy:
Would this be the right point to check files/dirs against exclude patterns?
https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NupkgWriter.fs#L237-L248

And how could `Fake.Globbing` help with that?

devboy:
Globbing is hard ...

![](http://media.giphy.com/media/5Zesu5VPNGJlm/giphy.gif)

mavnn:
/me couldn't possible comment on why the initial version didn't include
exclusions/globbing...

devboy:
[...]

PR #882, opened 2015-06-19
Exclusion syntax for paket.template files

devboy:
PR for #880 

- [X] parses `--` to exclude files
- [x] change to `!*` pattern
- [x] disallow the space to avoid ambiguity
- [X] apply pattern/globs during packaging
- [x] when fixing a relative path (`..`) append `\` to windows drive (`C:`)
- [ ] how to exclude entire directory

### DIFF
--- a/docs/content/template-files.md
+++ b/docs/content/template-files.md
@@ -100,6 +100,18 @@ If the source part refers to a file then it is copied into the target directory.
 refers to a directory, the contents of the directory will be copied into the target folder.
 If you omit the target folder, then the source is copied into the `lib` folder of the package.
 
+Excluding certain files looks like this:
+
+    [lang=batchfile]
+    files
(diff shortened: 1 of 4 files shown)


forki:
is this ready to merge?

devboy:
Will push the latest fixes after lunch ;)

devboy:
Apart from not matching against directories (`!myDir`) where you'd have to use (`!myDir/*.*`) everything should be in. It would probably be enough to fix that in **Fake.Globbing** without touching this code.

forki:
it's also important to have a bit of documentation for this.

devboy:
:+1:

devboy:
Is the documentation sufficient?

forki:
Yep thanks a lot

forki:
very cool. thanks a lot.

released in paket 1.18

PR #891, opened 2015-06-21
Fix #848: Improve the documentation for the commands

### LINKED ISSUE #848: Confusion between "paket install" and "paket update"
There's considerable overlap in the behavior of `paket install` and `paket update`, and some confusion on which one should be used in which situation.

# Summary
* `paket update` could also be called `paket forget-everything-and-re-resolve (and-install-into-projects)`
* `paket install` could also be called `paket only-update-packages-where-absolutely-necessary (and-install-into-projects)`

**Proposed solution:** My intuition is that `paket install` should really be called `paket update --pin-unchanged`, because this is what it really does.

`paket install` does not really have a purpose anymore then, but should probably be kept for compatibility reasons. The documentation of `paket install` could then say: "`paket install` has the same effect as `paket update --pin-unchanged`". This would also remove the confusion between the two commands.

For completeness: The third use case (no updates to `paket.lock`) is already covered by another command:
* `paket restore` could also be called `paket do-not-update-versions-and-install-into-projects`

# Details:
## Behavior of paket update:
  1. **Delete paket.lock if it exists**, therefore discarding the currently used versions of the packages  (*UpdateProcess.Update*).

  2. Compute a new `paket.lock` file based on `paket.dependencies` (*UpdateProcess.SmartInstall/.SelectiveUpdate*)

  3. (Optionally:) Install everything from paket.lock (*InstallProcess.InstallIntoProjects*)

## Behavior of paket install:
  1. If no `paket.lock` file exists: Same behavior as `paket update` (it just calls *UpdateProcess. SmartInstall*)

  2. **If a `paket.lock` file exists:** (*UpdateProcess.SmartInstall/.SelectiveUpdate*)
       1. Compare `paket.lock` and `paket.dependencies` to find (*DependencyChangeDetection. findChangesInDependenciesFile*):
          * Dependencies that were added to `paket.dependencies`
          * Dependencies where the current version from `paket.lock` does not match the version requirement from `paket.dependencies`
       2. **Pin all unchanged dependencies** (i.e. the ones that are not newly added or modified)
       3. Compute a new `paket.lock`, taking into account the pinned dependencies.
       
  3. Install everything from paket.lock (*InstallProcess.InstallIntoProjects*)

## Behavior of paket restore:
  1. Install everything from paket.lock (*RestoreProcess.Restore*)

cr7pt0gr4ph7:
Besides improvement of the wording of the commands, as per #848, this pull request also introduces some improvements to the visual look of the commands docs.


forki:
wow awesome work.

dance

PR #892, opened 2015-06-22
add tab completion for installed packages

ctaggart:
This is in progress.

image

This will also allow me to add tab completion to several of commands that need a list of currently installed package names.


ctaggart:
This is ready to go. Looks like I updated `paket.lock` accidently. Let me know if you want me to roll that back.

image

forki:
Very cool

PR #896, opened 2015-06-22
Enable compilation on higher VS versions even if the F# plugin is not installed

cr7pt0gr4ph7:
Currently, the build process fails if one is building with an MSBuild version corresponding to a Visual Studio version higher than `11.0` if the Visual F# plugin for Visual Studio is not installed.

The changes in this pull lets the projects use the Visual F# Tools in version `3.1` or `3.0`, in that case, provided that they are installed.

Also, if no matching `Microsoft.FSharp.Targets` file is found, the build fails explicitly - because `<Import Project="$(FSharpTargetsPath)"/>` expands to `<Import Project=""/>` - instead of failing silently because the project has no compilation output.

### DIFF
--- a/src/Paket.Core/Paket.Core.fsproj
+++ b/src/Paket.Core/Paket.Core.fsproj
@@ -43,18 +43,23 @@
     <MinimumVisualStudioVersion Condition="'$(MinimumVisualStudioVersion)' == ''">11</MinimumVisualStudioVersion>
   </PropertyGroup>
   <Choose>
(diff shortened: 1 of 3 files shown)


cr7pt0gr4ph7:
Another alternative would be to use the open source F# compiler that is available on NuGet as [**`FSharp.Compiler.Tools`**](https://www.nuget.org/packages/FSharp.Compiler.Tools/).

forki:
not completely sure if I understand the change, but if make it work on more machines :+1:

PR #897, opened 2015-06-23
paket install fails with KeyNotFound exception if paket.references refers to unknown reference

haraldsteinlechner:
this patch wraps safe access around the packages map in order not to throw boring key not found exception


forki:
wow I just fixed this one minute before you send this.

Krzysztof-Cieslak:
Apparently not only I had same problem :D

forki:
Anyways. I'm really happy that you took the time to fix this. It was just a
really strange coincidence that I fix it in the very same moment.

Thanks again

PR #900, opened 2015-06-23
add section to nuget spec docs around using multiple config options

baronfel:
It was confusing to me how one would use more than one of the supplied configuration options, but clear once I debugged through the Paket code.  In order to prevent confusion in the future, I thought we could document this?


forki:
thx

PR #903, opened 2015-06-25
added fix for windows-style network source-paths in dependencies parser

krauthaufen:
Since (on windows) nuget sources can be specified like

    nuget \\server\share

It would be very helpful to support these paths on linux too.
My fix is simply transforming those paths to

    nuget smb://server/share

whenever the OS is linux or MacOS (which is working at least on ubuntu)


forki:
thx a lot

Issue #682, opened 2015-03-05
Investigate issues when using Klondike Nuget servers


baronfel:
For Klondike-powered nuget feeds it seems the URLs may be different for the 'normal' nuget feeds.

My machine configuration was pointed to <serverroot>/api and the 'convert-from-nuget' command could not communicate with the server.

After some investigation, I realized that my nuget machine config and my VS nuget config differed in URL: the Vs ones were <serverroot>/api/odata.  When I changed the machine config to match (because it seems paket uses the nuget machine config), the convert operation completed successfully.

forki:
I think we would need some help here. I have no idea what a "Klondike-powered nuget feed" is.

baronfel:
(Klondike)[https://github.com/themotleyfool/Klondike] feeds are nuget feeds with a Lucene-powered search frontend on top of them.

I believe it's mostly used as a self-hosted, more performant, private nuget server.

forki:
do you have a repro sample for me to check?

mavnn:
If you configure your paket.dependencies to use the odata end point, then klondike and paket play nicely together:

```
source http://my.klondike.server/api/odata
```

Unfortunately klondike does not implemented most of the nuget v2 api, but doesn't error if you call an unsupported method - instead it redirects to the docs, which paket then (reasonably) tries to interpret as an api response.

forki:
Can we configure paket to understand such feeds?

mavnn:
Not really, I don't think. You'd have to infer that you'd recieved an html
response rather than a valid api response and then fall back to odata as a
second attempt.

forki:
Please see #907

baronfel:
I'll try to investigate this today and see if we can close this issue thanks to @ilkerde's awesome work.

PR #907, opened 2015-06-28
Establish beta-level compatibility with Klondike

ilkerde:
I did some quick fixup in order to get Klondike to be more chatty with Paket. More specifically, I had problems to obtain package details from Klodike during `paket install` and `paket restore` operations. After some investigation, I realized that I was not alone (see #682).

After sniffing on the conversation between original nuget.exe and Klondike, I came up with what you can see now in this PR.

I tested it with several projects of mid-range (10-25 deps) with mixed sources (from nuget and Klondike). NTL, I would still not be to positive about the Klondike support. It _seems_ to work fine.


forki:
cool thx

PR #915, opened 2015-07-06
adjust semver normalization to correctly handle build metadata on prerelease

baronfel:
Normalization right now adds a '.' character in front of the prerelease build information, which results in a bad version string.

Before: (SemVer.Parse "3.0.0-alpha+ci-008").Normalize() ==> "3.0.0-alpha.+ci-0008"
After: (SemVer.Parse "3.0.0-alpha+ci-008").Normalize() ==> "3.0.0-alpha+ci-0008"

### DIFF
--- a/src/Paket.Core/SemVer.fs
+++ b/src/Paket.Core/SemVer.fs
@@ -3,26 +3,62 @@
 open System
 open System.Text.RegularExpressions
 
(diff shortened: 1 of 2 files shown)


baronfel:
closing because I think I found a larger issue that need to be addressed.

PR #920, opened 2015-07-08
Address defects in SemVer comparison and parsing

baronfel:
So this got kinda big, I'm sorry in advance.

At work, we made some nuget packages of the form `Mozu.Component.Contracts.1.1.0-dev1507031354ci`.  That is the package Id, followed by an increasing major, minor, patch.  Prereleases are of the form 'dev', then a YYMMDDHHmm timestamp, and then a few characters from the branch that created the package, in this case 'ci'.

This is fine from Nuget's point of view, because a prerelease term is any non-period string after a patch. This is also fine from SemVer 2.0, because a prerelease term is

    a hyphen and a series of dot separated identifiers immediately following the patch version. 
    Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. 
    Identifiers MUST NOT be empty. 
    Numeric identifiers MUST NOT include leading zeroes

However, Paket seems to be more strict than this, requiring that a prerelease term be only of the form `(optional string part)(optional number part)`.  So what happens is that the prerelease term is not parsed at all, which means a restore/install pulls an arbitrary Major.Minor.Patch version of the package Id in question from the server.

So for this pull request I redid the way that Paket parses and interprets prerelease identifiers, making a Prerelease record track a set of one or more PrereleaseSegments which can be either alphanumeric or numeric.  This is important because the number of segments, as well as their types, has implications for comparisions according to section 11 of the SemVer 2.0 spec.

I've also addressed the build metadata portion of the spec.  Based on my reading I see room for only one chunk of build metadata that isn't necessarily tied to whether or not a particular version string has a prerelease section or not.  So I renamed that item.  The build metadata is also not to be used in comparisions, so I removed it from consideration in the implementation of IComparable for the SemVerInfo class.

### DIFF
--- a/src/Paket.Core/SemVer.fs
+++ b/src/Paket.Core/SemVer.fs
@@ -3,26 +3,63 @@
 open System
 open System.Text.RegularExpressions
 
+module utils =  
+    let zipOpt l1 l2 = 
+        let llength, rlength = List.length l1, List.length l2
+        seq {
+            for i in 0..(max llength rlength) do
+                let l = if llength > i then Some (List.nth l1 i) else None
+                let r = if rlength > i then Some (List.nth l2 i) else None
+                yield l, r
+        }
+
+[<CustomEquality; CustomComparison>]
+type PreReleaseSegment = 
+    | AlphaNumeric of string
+    | Numeric of bigint
+
+    interface System.IComparable with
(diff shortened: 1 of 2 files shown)


forki:
thx a lot

Issue #901, opened 2015-06-24
Add an option to `paket pack` (or paket.template) to depend on current versions of packages


dmitryvk:
Currently, when I have `paket.template` file like this:
```
type project
```
then `paket pack` command will use version requirements from `paket.dependencies`. I would like to also be able to use versions from `paket.lock`.

E.g., I'd like something like this:
```
# paket.dependencies
Foo >= 1.0 < 3.0
```
```
# paket.lock
Foo (1.1)
```
```
# paket.references
Foo
```
```
# paket.template
type project
use-locked-versions >=
```
then `paket pack` would produce package with the following version requirement: `Foo >= 1.1 < 3.0`.

The new directive `use-locked-versions` specifies how to combine version from `paket.lock` with version requirement from `paket.dependencies`. In this case, it combines `>= 1.0 < 3.0` with `>= 1.1`. It might be useful to be able to use this on per-package basis.

Do you think that this would be useful feature?

dmitryvk:
This proposal is different than specifying e.g.
```
# paket.template
type project
dependencies
  Foo >= CURRENTVERSION
```
because currently I'd have to enumerate all referenced packages in dependencies list.

Another option would be to allow wildcards in `dependencies`, e.g.,
```
# paket.template
type project
dependencies
  Bar >= 1.1
  Baz >= CURRENTVERSION
  Foo.* < 3.0
  * >= CURRENTVERSION
```

forki:
I think the second proposition with dependency wildcards would make sense.

dmitryvk:
I gave some more thought to this at noticed that CURRENTVERSION refers to the version of package being packed; we should probably add some more tokens that refer to version of a dependency, e.g.:
```
CURRENT_DEP_VERSION # version of a dependency from paket.lock (e.g., 1.2.3 or 1.2.3-alpha)
CURRENT_DEP_VERSION_MAJOR # trimmed to major version (1 or 1-alpha)   
CURRENT_DEP_VERSION_MINOR # trimmed to minor (1.2, 1.2-alpha)
```

forki:
Yep that sounds reasonable

inosik:
I would prefer a simple `$ paket pack [...] use-locked-versions` or something like that, which would tell Paket to take the version infos from `paket.lock` instead of `paket.dependencies`. Actually that's what I wanted to suggest, too :grin: And I think this would only make sense for `paket.template`-Files with `type project`.

[...]

PR #924, opened 2015-07-16
Get version requirements from paket.lock

inosik:
Addresses #901.

This allows us to get the version requirements from `paket.lock` instead of `paket.dependencies`. That means we then require the specific version (`version="[x.y.z]"`), which is resolved right now. To use this, just issue `paket pack [...] lock-dependencies` on the command line.

What do you say? Should I add some tests or maybe documentation?

### DIFF
--- a/docs/content/commands/pack.md
+++ b/docs/content/commands/pack.md
@@ -48,7 +48,7 @@ Depending on which command you issue, Paket creates different version requiremen
   </tbody> 
 </table>
 
-As you see here, the first command (without the `lock-dependencies` parameter) creates version requirements as specified in your [`paket.dependencies` file][depfile]. The second command takes the currently resolved version from your [`paket.lock` file][lockfile] and "locks" it to this specific version.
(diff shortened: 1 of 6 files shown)


forki:
could you please create small docs which show (maybe in a table) what version requirement we create for which paket.lock version. Maybe we can switch to the locked versions completely.

inosik:
> Maybe we can switch to the locked versions completely.

You mean to always take the constraints from `paket.lock`? I actually like the default behavior. And it completely makes sense, if the upstream packages follow SemVer, of course.

forki:
So this should be a per package decision?

inosik:
I mean we should keep the default behavior, but if one runs `paket pack lock-dependencies`, all the dependencies of the resulting package(s) should be locked.

I thought you mean to drop the current behavior (taking the constraints from `paket.dependencies`).

inosik:
I think I'm done here. Maybe somebody could provide a real-world `paket.dependencies`, `paket.lock` and `paket.references`, to show a better example in the docs?

forki:
thanks.

PR #928, opened 2015-07-21
Allow nuget packages to put version in the path

forki:
- [x] allow to use `version_in_path: true` in dependencies file
- [x] install package with version in the path
- [x] update `packages.config` file to the current version


forki:
this will also allow to solve lots of other NuGet incompatibilities.

Docs: http://fsprojects.github.io/Paket/nuget-dependencies.html#Putting-the-version-no-into-the-path

Please test like crazy and report bugs.

Issue #914, opened 2015-07-06
Paket can't find build configurations specified with dynamic OutputPath node


amazingant:
When processing a project file using `<OutputPath>bin\$(Configuration)\</OutputPath>` in the main `<PropertyGroup>` node, running `paket pack output "bin"` results in:

> Paket failed with:
>     Unable to find Release output path node in file C:\File\Path\Here\Project.fsproj

I'd try to jump in and work on some changes to fix this myself, but I'm not sure how the paket people feel this should be resolved.

forki:
Not really sure I understand the issue.

/cc @mavnn

amazingant:
In a typical project file, Visual Studio builds little sections of XML for each build configuration &mdash; `Debug|AnyCPU`, `Debug|x86`, `Release|AnyCPU`, etc. &mdash; and adds values that are considered specific to each configuration. One of these is an `OutputPath` value, which indicates where the compiled binaries are put. When compiling a project, MSBuild goes through the project file and process each value, dynamically resolving placeholders like `$(Configuration)` to the appropriate value. So rather than providing an `OutputPath` node for each build configuration, if you were writing a project file by hand, you could write a single node like `<OutputPath>bin\$(Configuration)\</OutputPath>` in the section of the project file that applies globally, and MSBuild would dynamically resolve that to `bin\Release\`, `bin\Debug\`, etc. as needed.

With a project file that only has one `OutputPath` node configured as above, `paket pack` commands refuse to run, giving the output I mentioned before. I assume this is because either paket is looking for an `OutputPath` within the block of XML specific to the current build configuration and can't find it, or because paket found the `OutputPath` but isn't processing the placeholder and can't find a folder named `bin\$(Configuration)\`.

forki:
any ideas to fix this in https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/ProjectFile.fs#L585 ?

amazingant:
Not quickly/easily; my only useful thought is to process every node in the project file, loading each value into a dictionary, and process the placeholders like MSBuild does. Realistically, doing that _exactly_ how MSBuild does it means also processing external `.targets` files as well; hence my concern about how you feel like resolving this. Ignoring external `.targets` files would be fine for my current use-case, but would still preclude potential valid project files from being used.

forki:
Mhm, if you send a PR which solves your use case that would still be an improvement...

amazingant:
[...]

PR #942, opened 2015-07-25
Handle dynamic OutputPath

amazingant:
Partial implementation so far, updates `ProjectFile.GetOutputDirectory` to iterate through every `PropertyGroup` node in the project file loading values. Need to find a nice way to process basic placeholder values (the full supported syntax is a [bit expansive](https://msdn.microsoft.com/en-us/library/dd633440.aspx) for my liking) and the [`Condition` attribute](https://msdn.microsoft.com/en-us/library/7szfhaft.aspx). Should also provide some default values in the `Map<string, string>` at start, such as the `buildConfiguration` value.

As a side-effect of rewriting this code, I've accidentally corrected how Paket handles project files with multiple `OutputPath` nodes. MSBuild appears to process everything top-to-bottom, and so the last `OutputPath` node wins, whereas Paket will currently output a warning and use the first `OutputPath` it found.

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -789,9 +789,11 @@ type ProjectFile =
                         then data
                         else addData data node
 
+        let startingData = Map.empty<string,string>.Add("Configuration", buildConfiguration)
+
         this.Document
         |> getDescendants "PropertyGroup"
-        |> Seq.fold handleElement Map.empty<string,string>
+        |> Seq.fold handleElement startingData
         |> Map.tryFind "OutputPath"
         |> function
             | None -> failwithf "Unable to find %s output path node in file %s" buildConfiguration this.FileName
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -584,26 +584,64 @@ type ProjectFile =
         else "Content"
 
     member this.GetOutputDirectory buildConfiguration =
+        let rec handleElement (data : Map<string, string>) (node : XmlNode) =
+            let inline processPlaceholders data text =
+                // TODO: Process placeholders
+                text
+
+            let inline conditionMatches data condition =
+                // TODO: Process conditions
+                true
+
+            let inline addData data (node:XmlNode) =
+                let text = processPlaceholders data node.InnerText
+                // Note that using Map.add overrides the value assigned
+                // to this key if it already exists in the map; so long
+                // as we process nodes top-to-bottom, this matches the
+                // behavior of MSBuild.
+                Map.add node.Name text data
+
+            let inline handleConditionalElement data node =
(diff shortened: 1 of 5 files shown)


amazingant:
Still trying to figure out a way to handle conditions. I've got some code written using `String.Split` creatively that seems to handle simple conditions like `'$(Configuration)|$(Platform)' == 'Debug|AnyCPU'`, but doesn't handle any of the condition syntax beyond `==` and `!=`. If I skip all xml nodes with conditions that this can't process, that should still be enough to handle the project I opened #914 for.

Wanting to support a little more of the condition syntax for [more complex conditions](https://github.com/fsprojects/Paket/blob/927505d0c25935c493c81ffcc461f0b30bee831e/tests/Paket.Tests/ProjectFile/TestData/Project1.fsprojtest#L16), all I've come up with so far is a rather complicated [regular expression](https://gist.github.com/amazingant/33c070f1d36b53428909). Anyone else have any thoughts?

cc/ @forki: I did get _some_ work done on this :)

forki:
nice. 

did you already look at the red builds?

amazingant:
Yeah. I've gotten either the Release or Debug tests to pass by making changes to the `fsprojtest` files, but I can't get both to pass as-is until I put something in that handles conditions. Last time I tried it at home, it was throwing an exception trying to handle line 16 of `Project1.fsprojtest` because I hadn't added handling for `Or` in conditions.

Does a really big regular expression sound like a horrible idea to you? I'll use `sprintf` or `String.Format` to build it out of smaller ones, so there isn't a 300-character regex to deal with, and a recursive function should be able to handle processing the match data it returns.

[1]: https://gist.github.com/amazingant/33c070f1d36b53428909

forki:
>> Does a really big regular expression sound like a horrible idea to you?

yes ;-)

amazingant:
Sorry for the delay; I kept working at the regex idea, and found that as soon as I add three or more conditions separated by `and` or `Or`, the grouping data stops being as useful. The code isn't as short, but I've started building a parser at the [same gist](https://gist.github.com/amazingant/33c070f1d36b53428909), in case you're interested. I'd love to just throw FParsec at this, but I don't think it'd be good to add it as a dependency to Paket.

forki:
we had the idea to use FParsec for all parsers. but at the moment there is nobody who wants to build these.

amazingant:
Finished working on the code from the gist and built it in, tests are happy now. Is there anything else I should be doing with this before it'll be acceptable for merging?

forki:
this looks cool. let's see if it works in the wild wild real world ;-)

forki:
real

amazingant:
Thanks! I'll change one of my work projects back and give it a go tomorrow morning :)

amazingant:
Looks good from my FAKE script :)

PR #957, opened 2015-08-03
Selective update resolves the graph for selected package

mrinaldi:
This PR fixes the behavior of selective update (update specifying a package).

The current behavior is to pin the version for all packages, except the one specified, and resolve the whole graph.

This would cause the dependencies of the specified package to be pinned and the package would not be updated if it requires that one or more dependencies need to be updated as well.

For example:
``Castle.Core-log4net 3.2.0`` depends on ``Castle.Core >= 3.2.0``
``Castle.Core-log4net 3.3.3`` depends on ``Castle.Core >= 3.3.3``

If I try to update ``Castle.Core-log4net`` from ``3.2.0`` to ``3.3.3``, it would not succeed because ``Castle.Core`` is pinned to ``3.2.0``.

This PR only resolve the specified package graph.

### DIFF
--- a/tests/Paket.Tests/Lockfile/ParserSpecs.fs
+++ b/tests/Paket.Tests/Lockfile/ParserSpecs.fs
@@ -47,12 +47,12 @@ let ``should parse lock file``() =
     packages.[1].Source |> shouldEqual PackageSources.DefaultNugetSource
     packages.[1].Name |> shouldEqual (PackageName "Castle.Windsor-log4net")
     packages.[1].Version |> shouldEqual (SemVer.Parse "3.3")
-    packages.[1].Dependencies |> shouldEqual (Set.ofList [PackageName "Castle.Windsor", VersionRequirement.AllReleases, []; PackageName "log4net", VersionRequirement.AllReleases, []])
+    packages.[1].Dependencies |> shouldEqual (Set.ofList [PackageName "Castle.Windsor", VersionRequirement(Minimum(SemVer.Parse "2.0"), PreReleaseStatus.No), []; PackageName "log4net", VersionRequirement(Minimum(SemVer.Parse "1.0"), PreReleaseStatus.No), []])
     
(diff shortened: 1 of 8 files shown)


forki:
selective update should unpin all transitive package dependencies right now. do you have a repro case?

forki:
OK, seems you are right and we have a regression.

I tested the following deps file:

    nuget Castle.Core-log4net

with lock file:

    NUGET
      remote: https://nuget.org/api/v2
      specs:
        Castle.Core (3.2.0)
        Castle.Core-log4net (3.2.0)
          Castle.Core (>= 3.2.0)
          log4net (1.2.10)
        log4net (1.2.10)

and it didn't update. That is weird. I will now look into your PR.

tests/Paket.Tests/Lockfile/ParserSpecs.fs
```diff
@@ -47,12 +47,12 @@ let ``should parse lock file``() =
     packages.[1].Source |> shouldEqual PackageSources.DefaultNugetSource
     packages.[1].Name |> shouldEqual (PackageName "Castle.Windsor-log4net")
     packages.[1].Version |> shouldEqual (SemVer.Parse "3.3")
-    packages.[1].Dependencies |> shouldEqual (Set.ofList [PackageName "Castle.Windsor", VersionRequirement.AllReleases, []; PackageName "log4net", VersionRequirement.AllReleases, []])
+    packages.[1].Dependencies |> shouldEqual (Set.ofList [PackageName "Castle.Windsor", VersionRequirement(Minimum(SemVer.Parse "2.0"), PreReleaseStatus.No), []; PackageName "log4net", VersionRequirement(Minimum(SemVer.Parse "1.0"), PreReleaseStatus.No), []])
```
forki: why did these tests change?

forki:
mhm from looking at the existing code I'm not that sure that this worked before. maybe it only worked in my dreams.

so:

hug

Issue #937, opened 2015-07-23
paket does not accept paths that include Unicode characters


exercitusvir:
When I build a project that has a Unicode character in its name, I get the following error message in Visual Studio 2013:

> The command ""[solution path]\.paket\paket.exe" restore --references-files "[solution path]\source\[project name with Unicode character]\paket.references"" exited with code 1.

forki:
could please give an example? will try to fix that

exercitusvir:
The project name included the character "modifier letter half triangular colon" (U+02D1). But you will probably encounter the problem with any Unicode character. I confirmed that this was the problem by temporarily removing the offending character and it worked afterwards.

forki:
does this also happen when you run the same command manually from the cmd line?

exercitusvir:
The command line does not seem to complain, but the command line does not support Unicode itself (it displays the character as `?`), so I don't know what is actually happening there. But Visual Studio 2013 itself never had any problems with Unicode.

forki:
yeah it seems to be an issue with the targets file. I have no idea what to do here ;-(

exercitusvir:
Are you sure that paket supports Unicode in paths?

forki:
No idea. Never tried such a crazy thing ;-)

exercitusvir:
Are you planning to fix this? Supporting Unicode is not that crazy in the 21st century. `System.Globalization.StringInfo` contains all required methods. While `System.String` internally supports Unicode, it does not expose any Unicode-aware methods (e.g. String.length does not return the number of Unicode characters in the string).

forki:
> Are you planning to fix this?

I have no idea where it goes wrong, that's why I asked if can reproduce outside of the targets file.

exercitusvir:
How can I repro this outside of the targets file and outside of Visual Studio? 
This is easy to repro from within Visual Studio if you just place a Unicode character in a the path. 

If you don't explicitely use ` System.Globalization.StringInfo` on the paths than you probably don't support Unicode, because none of the `System.String` methods do.

forki:
I tried to add ^ into the path name and can't even open in VS:

![image](https://cloud.githubusercontent.com/assets/57396/9080723/de9dc7c0-3b56-11e5-8b76-44e124dc3385.png)

![image](https://cloud.githubusercontent.com/assets/57396/9080710/c20d2fa6-3b56-11e5-9da7-ce4de28cebb2.png)

forki:
[...]

PR #960, opened 2015-08-05
Escape file names

### LINKED ISSUE #959: Pack command doesn't escape filenames correctly
I think we've identified an issue with the Pack step in Paket, whereby it escapes spaces in folder names correctly, but not file names:

https://github.com/fsprojects/Paket/blob/68614ba680deaccbfc759f08bcd7ba08cdf1c833/src/Paket.Core/NupkgWriter.fs#L261

Pull request to follow. Let me know if I'm mistaken!

will14smith:
Fixes #959. Characters such as " " and "%" in file names would break when installing the package with nuget (and possibly paket). 

I couldn't see any tests around `NupkgWriter.fs` so I have created a [batch script](https://gist.github.com/will14smith/3c87a3ac7cc82cd0034b) that reproduces the issue.

### DIFF
--- a/src/Paket.Core/NupkgWriter.fs
+++ b/src/Paket.Core/NupkgWriter.fs
@@ -227,6 +227,10 @@ let Write (core : CompleteCoreInfo) optional workingDir outputDir =
         let path = DirectoryInfo(p).FullName
         exclusions |> List.exists (fun f -> f path)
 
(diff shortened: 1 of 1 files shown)


forki:
related to https://github.com/fsprojects/Paket/issues/937?

forki:
thx. this is now in the latest alpha version.

PR #964, opened 2015-08-06
Check for conflicts in selective update

mrinaldi:
This PR complements the #957.

Since it doesn't resolve the whole graph, it must create constraints from the information in lock file to later check for conflicts while resolving the graph for the package selected.


forki:
thx a lot!

Issue #966, opened 2015-08-07
Paket mixed responses and downloads


SavaNDragos:
I found a pretty ugly bug on the new servers (not sure if it matters but the servers are windows server 2012 with Jenkins). We also are using our internal feed.
The bug is like this: 
When paket asks for details related to a package and you have more packages for which it needs details, it will end up mixing the reponses and also the downloads:
From what I can tell there isn't a secure way to receive the callback from the nuget feed and they look to be processed randonly. 

So from the log:
...
Response from http://nuget.mycompany.net/NugetWebFeed/nuget/Packages?$filter=Id eq 'Graphite.NET' and Version eq '1.1':

Graphite.NET <?xml version="1.0" encoding="utf-8" standalone="yes"?>
<feed xml:base="http://nuget.mycompany.net/NugetWebFeed/nuget/"....
  <title type="text">Packages</title>
...
    <id>http://nuget.mycompany.net/NugetWebFeed/nuget/Packages(Id='Mycompany.Assambly',Version='0.0.075.8635')</id>
    <title type="text">Mycompany.Assambly</title>

In from of the response I added the package name (for debug purpose: Graphite.NET <?xml version="1.0).
Now what is more funny is because everything is parallel you can end up getting as a response a package:  

Response from http://nuget.mycompany.net/NugetWebFeed/nuget/Packages?$filter=Id eq 'Microsoft.Web.Services3' and NormalizedVersion eq '3.0.0':

Microsoft.Web.Services3 PK
     
5GÕ¦à      _

The mix happens when the downloads take place also. So even if you received a good response you can get a wrong package. 

At the end the builds fail because in packages\apackage you will have binaries belonging to anothe package.

This is happening on all our new servers.

At the moment I tried to limit the calls to the internal feed [singleton when making calls] and the response get back ok, but there is no parallel anymore. 
We want to use the official release so any suggestions?

SavaNDragos:
I should not this is happening with both restore and update (stuff i tested)

forki:
That sounds really strange. The parallel calls are not matching the results? So that's a bug in the async parallel calls? 

/cc @dsyme ever heard of something like that in other F# code?

forki:
Is there a way to reproduce this behavior? What did you change exactly to make it work?

SavaNDragos:
I used Monitor.Enter Paket.Constants.InUse in the download packages method  but this is not ok because it is not intended to be used in async

forki:
[...]

PR #969, opened 2015-08-10
[https://github.com/fsprojects/Paket/issues/966] Paket mixed responses and downloads [Fix]

SavaNDragos:
I tested this modifications on several of our windows server 2012 r2 and we didn't see the fail happening again.
I know that this is not the best approuch (not using the async for downloading the string anymore) but it seems to do it's job.
Note: I used '|> Async.RunSynchronously' so it wil be easy to remove it at a later time after we correctly identified the issues.


forki:
replaced with https://github.com/fsprojects/Paket/pull/971

PR #970, opened 2015-08-10
Make Paket compile on systems with only F# 4.0 installed

part-timeDev:
Make Paket compile on systems with only F# 4.0 installed


forki:
thx buddy

PR #977, opened 2015-08-13
Version range semantics changed for >= x.y.z prerelease

forki:
Since this is a common misunderstanding and we see lots of requests I changed the semantics for `>= x.y.z prerelease` to include `x.y.z-alphaabc` versions. Before we would exclude all prereleases of `x.y.z` since they are considered smaller than `x.y.z`.

Please see the tests.


agross:
+1

PR #979, opened 2015-08-13
Update specifying version does not pin version in paket.dependencies

### LINKED ISSUE #973: Update specifying version pins version
This is a question rather than an issue.

While updating, when you specify the version, that version is pinned in paket.dependencies.
Is this correct? I couldn't find anything related to it in the docs.

I don't want to pin the version, I just want to update to that specific version and that's it.
I'm willing to send a PR either to fix this behavior or to introduce a parameter to the update command.

What do you think?

mrinaldi:
This PR fixes #973.


forki:
LGTM thx.

PR #981, opened 2015-08-13
Init adds default feed

### LINKED ISSUE #954: `paket init` should add the main NuGet feed to the dependencies file by default
I think it should, because cases where you really *don't* want that should be all but nonexistent, and it's annoying to always have to look that address up.

mrinaldi:
This PR fixes #954.


forki:
Thx

PR #985, opened 2015-08-14
Only creates requirements from lock file when updating a single package

mrinaldi:
When I'm updating all packages, there should be no requirements created from lock file.

Requirements should only be created when you don't resolve the whole graph (i.e. update a single package).


forki:
thx.

Issue #984, opened 2015-08-14
Add support for Xamarin.iOS + Xamarin.Mac


TIHan:
I noticed Paket only supports MonoTouch, MonoAndroid, and MonoMac. But, with the new unified API, we now have Xamarin.iOS and Xamarin.Mac which Paket does not support.

Issue #982, opened 2015-08-14
MonoTouch + MonoDroid considered compatible with "net45" when in reality, it is not.


TIHan:
https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/FrameworkHandling.fs#L103

I think the problem with this is that if there is a portable library that targets MonoTouch or MonoDroid as well as "net45", then Paket will choose the non-portable library "net45" by priority: https://github.com/TIHan/Paket/blob/master/src/Paket.Core/PlatformMatching.fs#L63

But, the "net45" lib can contain references to "Reflection.Emit" which is incompatible.

I'm looking into this myself, and just making the suggested platforms for MonoTouch/MonoDroid/etc. to an empty list.

PR #986, opened 2015-08-14
Xamarin iOS + Mac Support / Handling fallbacks mainly for Xamarin against PCLs / Removed supported platforms for MonoTouch and MonoAndroid

TIHan:
https://github.com/fsprojects/Paket/issues/982
https://github.com/fsprojects/Paket/issues/983
https://github.com/fsprojects/Paket/issues/984

I had to address all three of the issues above in order to use Paket in day to day work. 

We are working with 50 F# projects, 20+ of them are shared between server and iOS with generated iOS projects; due to XS not supporting PCL and because we are using FsPickler (can never be PCL). Paket is most definitely the solution to keep our third-party libraries in sync across projects and plays nicely with different target frameworks, especially now with this PR.

In the current state of Paket, we cannot use it effectively because there is no Xamarin.iOS support and it will not properly target PCL profiles if it can't find a specific target framework, especially for Xamarin.iOS. This PR address this.

The scary change in this PR was removing "net45" as a supported platform for MonoTouch/MonoAndroid, and now Xamarin.iOS/Xamarin.Mac. I understand this was a fallback, but these platforms are incompatible with each other, regardless if it compiles or not - not all libraries will compile with it, ex: FSharp.Core.


forki:
My mother is getting married tomorrow so I have to review this on Monday. Thanks in advance.

TIHan:
No worries. Have fun! :)

forki:
thx

PR #991, opened 2015-08-17
Paket template comment

MorganPersson:
Comments in paket.template files by using `#`.

### DIFF
--- a/src/Paket.Core/TemplateFile.fs
+++ b/src/Paket.Core/TemplateFile.fs
@@ -34,6 +34,11 @@ module private TemplateParser =
             Some (!! 1 m, m.Groups.[2].Value.Trim())
         | false -> None
 
+    let comment = Regex(@"^\s*#", RegexOptions.Compiled)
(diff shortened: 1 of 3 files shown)


forki:
Could you please also add // as comments? IIRC we support both in dependencies file

src/Paket.Core/TemplateFile.fs
```diff
@@ -224,11 +231,12 @@ module internal TemplateFile =
     let private fromReg = Regex("from (?<from>.*)", RegexOptions.Compiled)
     let private toReg = Regex("to (?<to>.*)", RegexOptions.Compiled)
     let private isExclude = Regex("\s*!\S", RegexOptions.Compiled)
-
+    let private isComment = Regex("\s*#", RegexOptions.Compiled)
```
forki: you already have the same/similar regex above

MorganPersson:
support for // added

forki:
very cool. thx

PR #995, opened 2015-08-18
Fails the update of a nonexistent package

mrinaldi:
This PR makes update of a package that doesn't exist in paket.dependencies fail.

```paket update nuget Ninject```

Before:
```
Paket version 1.31.0.0
Updating Ninject in D:\Projects\paket-test\paket.dependencies
Resolving packages:
0 seconds - ready.
```

After:
```
Paket version 1.31.0.0
Paket failed with:
        Package Ninject was not found in paket.dependencies.
```


forki:
Oh I thought we already do this. Thx

PR #996, opened 2015-08-19
Escape file names in a NuGet compatible way

### LINKED ISSUE #975: Pack command doesn't escape filenames correctly (mark 2)
While https://github.com/fsprojects/Paket/issues/959 resolved issues with spaces, it also unnecessarily encodes parenthesis, which again results in files being missed when installing a nupkg. Need to review the NuGet spec more closely here, as there may be differences between folders and files in how they should be encoded.

will14smith:
Fixes #975 

I did some digging into what NuGet was actually doing. The claim that it URL encodes the paths is technically correct, it just chooses a strange selection of categories to encode. Eventually it calls [this](
http://referencesource.microsoft.com/#WindowsBase/Base/System/IO/Packaging/PackUriHelper.cs,915 ) where it does the actual encoding.

I've also changed the extracting to URL decode the file paths.

Edit: Repro script here: https://gist.github.com/will14smith/3c87a3ac7cc82cd0034b

/cc @jamescrowley

### DIFF
--- a/src/Paket.Core/NuGetV2.fs
+++ b/src/Paket.Core/NuGetV2.fs
@@ -405,14 +405,14 @@ let ExtractPackage(fileName:string, targetFolder, name, version:SemVerInfo) =
             // cleanup folder structure
             let rec cleanup (dir : DirectoryInfo) = 
                 for sub in dir.GetDirectories() do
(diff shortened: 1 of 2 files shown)


forki:
cool stuff

PR #997, opened 2015-08-19
UtilsSpecs: Unit test corrected + compatibility issues with Klondike NuGet server

danielmue:
First commit 71b69f3
- Use Path.Combine instead of +Path.DirectorySeparatorChar.ToString()+
- Fixes error if HOMEPATH=\ (Root directory, result was 'H:\\' or '//data' on Unix)

Second commit fd394fc
- WARNING: I am not a F# developer and this code could possibly break alternative NuGet server versions :-)
- "It works on my machine" using the current master of Klondike (https://github.com/themotleyfool/Klondike)


forki:
cool thanks.

Issue #946, opened 2015-07-30
"paket update nuget XXX" should report why it decides not to


et1975:
if paket decides not to update the package it was explicitly asked to update it should at least tell me why. 
As it is, neither the return code nor the message tell me what happened - in my case dependency of the package was requiring an upgrade as well.

forki:
Do you have a specific sample for me? Deps file + lock file + package you want to update? Would love to improve the output

et1975:
Happened with some private packages, but I'll keep an eye out in my public projects for an opportunity to reproduce it. 
It should be a fairly common scenario though, so I'm wondering what is the expected behavior in this case:
- solution S has a dependency on packages A and B
- package B also has a dependency on package A
- package B has had its A dependency upgraded and the new version of B is released
- solution S upgrades B to latest (or rather fails to, quietly)

In my case the A package was RabbitMQ.Client which happened to have a new _revision_ released.

forki:
Our current algorithm wasn't designed for such questions, but I'd love to
improve it in this direction. So I need good samples.

et1975:
Incidentally, it does upgrade as expected if I specify the version explicitly.

forki:
That is not unusual. Often there's more than one valid resolution.
Specifying more restrictions (concrete versions) leads to picking another.

mrinaldi:
I have a repro case here:

paket.dependencies:
```
source https://nuget.org/api/v2

nuget Ninject.Extensions.Interception
nuget Ninject.Extensions.Logging.Log4Net
```

paket.lock
```
NUGET
  remote: https://nuget.org/api/v2
  specs:
    log4net (2.0.3)
    Ninject (2.2.1.4)
    Ninject.Extensions.Interception (2.2.1.2)
      Ninject (>= 2.2.0.0 < 2.3.0.0)
    Ninject.Extensions.Logging (2.2.0.4)
      Ninject (>= 2.2.0.0 < 2.3.0.0)
    Ninject.Extensions.Logging.Log4net (2.2.0.4)
      Ninject.Extensions.Logging (>= 2.2.0.0 < 2.3.0.0)
      log4net (>= 1.0.4)
```

```paket update nuget Ninject.Extensions.Interception```
Outputs:
``` 
Paket version 1.28.0.0
Updating Ninject.Extensions.Interception in D:\Projects\paket-test\paket.dependencies
Resolving packages:
 - Ninject.Extensions.Interception 3.2.0
[...]

PR #999, opened 2015-08-19
Warns when package is pinned

### LINKED ISSUE #786: Output WHY a package wasn't updated to the latest version
See https://github.com/tpetricek/FSharp.Formatting/issues/301#issuecomment-94270140

mrinaldi:
This PR warns when a package is pinned to a specific version.

This is what it looks like:
```
Paket version 1.31.0.0
Updating Ninject.Extensions.Logging.Log4Net in D:\Projects\paket-test\paket.dependencies
Resolving packages:
 Ninject.Extensions.Logging.Log4Net is pinned to version 2.2.0.4
 - Ninject.Extensions.Logging.Log4Net 2.2.0.4
 Ninject.Extensions.Logging is pinned to version 2.2.0.4
 - Ninject.Extensions.Logging 2.2.0.4
 - Ninject 2.2.1.4
 - log4net 2.0.3
2 seconds - ready.
```

This is the first PR I'm working on to better report why a package had a version chosen, aiming to close #786 and #946.

### DIFF
--- a/src/Paket.Core/PackageResolver.fs
+++ b/src/Paket.Core/PackageResolver.fs
@@ -222,7 +222,22 @@ let Resolve(getVersionsF, getPackageDetailsF, globalFrameworkRestrictions, rootD
                 exploredPackages.[(normalizedPackageName,version)] <- package
                 package
             | _ -> package
-        | false,_ ->         
+        | false,_ ->
+            let getPinnedVersion d = 
+                match d.Parent with
+                | DependenciesFile _ ->
+                    match d.VersionRequirement.Range with
+                    | Specific v -> Some v
+                    | OverrideAll v -> Some v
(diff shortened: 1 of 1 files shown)


forki:
very cool. will look at it on monday when I'm back at work.

forki:
gimme

forki:
could please only warn if there is a new version available and we can't updte due to the lock? I think at the moment it's always reporting the pins, right?

mrinaldi:
It's always reporting the pin, indeed. I can warn only when a new version is available, however, to do that, I'll have to get the available versions for those packages, which will increase the resolve time. 

Is that OK?

forki:
Yes that would be OK. I thought we already do this in the resolver. If so
you can reuse the cached information.

PR #1003, opened 2015-08-20
New optional parameter _input_ for pack command

danielmue:
Hello,

We are using a single output directory for all project builds (like [FAKE](http://fsharp.github.io/FAKE/) suggests in the sample build script). 

Unfortunately I cannot use ```paket pack ..``` + ```type project ``` (_paket.template_) with this approach:
```
$ paket pack output _build.output/nuget/
Paket version 1.32.0.0
Paket failed with:
        Ein Teil des Pfades "some\path\project\bin\Release\project.dll" konnte nicht gefunden werden.
```
Currently _paket.exe_ analyzes the project settings to figure out which bin directory to use (normally bin\Release or bin\Debug).

I have added an optional parameter called ```input```. If defined, _paket.exe_ will lookup all assemblies from this directory.

@forki can you review this code please? :-) As I said in my previous pull request: I am not a F# developer and really afraid to break something crucial.

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -81,11 +81,18 @@ let getDescription attributes =
                       | Description d -> Some d
                       | _ -> None) 
 
-let loadAssemblyId buildConfig (projectFile : ProjectFile) = 
-    let fileName = 
+let buildAssemblyFileName (buildConfig : string, projectFile : ProjectFile, artifactsDirectory : string option) =
+    match artifactsDirectory with
+    | Some path ->
+        Path.Combine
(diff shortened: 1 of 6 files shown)


danielmue:
After talking with one of my colleagues I've got a better idea. Instead of introducing a new command line parameter for ```paket pack``` we could do this within the ```paket.template``` file.

outputDirectory _&lt;path or variable&gt;_

Possible variables:
* ```$(project.outputDirectory)``` default path defined in project settings (e.g. bin/Release)
* ```$(env.VARIABLE_NAME)``` shell / environment variable

```
type project
id Test.Paket.Package
version 1.0
authors Michael Newton
outputDirectory $(env.BUILD_OUTPUT_DIRECTORY)
description
    description of this test package
```

PR #1004, opened 2015-08-20
Update description of push command

inosik:
The `push` command doesn't behave like this since v0.30.0 or so. Found it a bit misleading while using Paket on the command line today.


forki:
thx

PR #1014, opened 2015-08-26
Warn about pinned packages only when a new version is available

mrinaldi:
This PR will warn about pinned packages only when there's a new version available as requested [here](https://github.com/fsprojects/Paket/pull/999#issuecomment-134120569).

It'll also disable the warning for install command because install pins the unchanged packages to the current version.

I could have changed the install to only resolve the changed packages, thus eliminating the need to disable the warning. However it wouldn't re-download the packages if ```--force``` were used.


forki:
very cool.

PR #1018, opened 2015-08-27
Introduce "Groups"

forki:
*This is what's going to be Paket 2.0*

## Basic feature description

This new feature will allow us to organize our dependencies in groups and also allows easier conflict resolutions. Let's consider a small sample:

### paket.dependencies

    nuget Newtonsoft.Json
    nuget UnionArgParser
    nuget FSharp.Core
    
    github forki/FsUnit FsUnit.fs
    github fsharp/FAKE src/app/FakeLib/Globbing/Globbing.fs
    github fsprojects/Chessie src/Chessie/ErrorHandling.fs
    
    group Build
    
        nuget FAKE
        nuget FSharp.Formatting
        nuget ILRepack
    
    group Test
    
        nuget NUnit.Runners
        nuget NUnit

As you can see we have 3 different groups. The first one is the default group (currently called "Main") and represents the backwards-compatible case. So the main group is what you currently have in Paket 1.x. The other two groups contain dependencies that are used for `Build` and `Test`.

Notice the indentation in groups is optional.

After `paket update` the corresponding lock file looks the following: 

### paket.lock

    NUGET
      remote: https://nuget.org/api/v2
      specs:
        FSharp.Core (4.0.0.1)
        Newtonsoft.Json (7.0.1)
        UnionArgParser (0.6.3)
    GITHUB
      remote: forki/FsUnit
      specs:
        FsUnit.fs (81d27fd09575a32c4ed52eadb2eeac5f365b8348)
      remote: fsharp/FAKE
      specs:
        src/app/FakeLib/Globbing/Globbing.fs (991bea743c5d5e8eec0defc7338a89281ed3f51a)
      remote: fsprojects/Chessie
      specs:
        src/Chessie/ErrorHandling.fs (1f23b1caeb1f87e750abc96a25109376771dd090)

    GROUP Build
    NUGET
      remote: https://nuget.org/api/v2
      specs:
        FAKE (4.3.1)
        FSharp.Compiler.Service (1.4.0.1)
        FSharp.Formatting (2.10.0)
          FSharp.Compiler.Service (>= 0.0.87)
          FSharpVSPowerTools.Core (1.8.0)
        FSharpVSPowerTools.Core (1.8.0)
          FSharp.Compiler.Service (>= 0.0.87)
        ILRepack (2.0.5)
        Microsoft.Bcl (1.1.10)
          Microsoft.Bcl.Build (>= 1.0.14)
        Microsoft.Bcl.Build (1.0.21)
        Microsoft.Net.Http (2.2.29)
          Microsoft.Bcl (>= 1.1.10)
          Microsoft.Bcl.Build (>= 1.0.14)
        Octokit (0.14.0)
          Microsoft.Net.Http
    GITHUB
      remote: fsharp/FAKE
      specs:
        modules/Octokit/Octokit.fsx (991bea743c5d5e8eec0defc7338a89281ed3f51a)
          Octokit

    GROUP Test
    NUGET
      remote: https://nuget.org/api/v2
      specs:
        NUnit (2.6.4)
        NUnit.Runners (2.6.4)
        
As you can see every group is listed separately and we still have the backwards-compatible case for the "Main" group.
If you want to reference dependencies from projects you can do this via the following syntax in references files:

## paket.references

    File:FsUnit.fs .
    FSharp.Core
    
    group Test
      NUnit
      NUnit.Runners

## New ways for conflict resolution

We got many requests to support multiple versions of the same package with Paket. This was not possible since we had no way to discriminate these.
Now we can support that by using multiple groups:

    nuget Newtonsoft.Json
    
    group Legacy
    
        nuget Newtonsoft.Json ~> 6

Every group will be resolved independently so that we get:

    NUGET
      remote: https://nuget.org/api/v2
      specs:
        Newtonsoft.Json (7.0.1)
    
    GROUP Legacy
    NUGET
      remote: https://nuget.org/api/v2
      specs:
        Newtonsoft.Json (6.0.8)
        
Paket is downloading both version. We will get `packages/Newtonsoft.Json` with version 7.0.1 and `packages/legacy/Newtonsoft.Json` with the 6.0.8 bits.
In your references files you can now decide which version you want to use:

    group Legacy
    Newtonsoft.Json

## Open stuff

This is currently early alpha and definetly not production ready. But I will update this PR frequently and 

- [x] - Dogfooding: Use groups in Paket itself
- [x] - Create group specific commands like `paket restore group Test`
- [x] - Documentation (see http://fsprojects.github.io/Paket/groups.html)
- [x] - Testing
- [x] - https://github.com/fsprojects/Paket/pull/1014 is not merged - @mrinaldi could you try to merge and send a PR against the `groups`branch?

## Organizational remarks

* Please send all Paket 2.0 PRs against the `groups` branch. I will start to release alpha versions from this branch.
* Please continue to send all bugfix PR's against `master` branch.


Krzysztof-Cieslak:
Hi,

First of all, I like this feature very much - it's somehow similar to NPM with it's "normal" dependencies and "dev" dependencies but is pushing idea even further with custom group definitions. Big :+1: for this feature.

My only remark right now - `paket.dependencies` and `paket.references` file don't seems clear and readable and for me with this syntax. My suggestion would be to introduce indentation for (non-main) groups. It would allow to have backward compatibility for main group and clearer syntax for custom groups.

Adron:
Just throwing in my +1 too.

amazingant:
Another :+1: here with a question: is there any thought of `paket restore` gaining the ability to only restore specified group(s) of packages? Could be useful for when a CI server really only needs a subset of the referenced dependencies for what it's building, or when a dev is trying to get started on project X in a solution and doesn't really want to grab the libs needed for tests or whatever just yet.

forki:
Yes that is planned, but not implemented yet

forki:
@Krzysztof-Cieslak 515c47c shows that optional indentation is already working

Krzysztof-Cieslak:
Amazing :) IMO, looks better

tpetricek:
+1 From me. This looks like a great way to solve the problem with multiple versions!

Stift:
Very nice. I just throw my +1 in the round.
Will start testing ASAP

matthid:
Very nice indeed, +1.

forki:
yay. current paket 2.0 alpha can dogfood https://github.com/fsprojects/Paket/blob/groups/paket.dependencies

agross:
I very much like the idea! :+1: 

Will it be possible to omit the group in `paket.references` when a dependency is defined in a single group only? Just thinking about the single-version case and that you DRY w.r.t. the group name.

tpetricek:
One quick thought - would it make sense to allow specifying _just_ the group in `paket.references`? If you have group with unit testing references, you would just say:

    group Test

... and you would get references for `nunit`, `FsUnit`, `FsCheck` and whatever is in the group. This would save you whole 3 lines, so it's not really big deal, but it might make things even easier.

Krzysztof-Cieslak:
Don't agree with Tomas' suggestion - it would make people declare "project" groups and use one liners "group references" in every `paket.references`. Don't think it would be good practice.

forki:
@gross and @tpetricek I thought about your ideas and I think I don't want to implement this right now. I might have a different idea for some future version. So let's wait a bit with that. At the moment I think it's ok to be explicit in ref files.

tpetricek:
Being explicit sounds good to me - and I think @Krzysztof-Cieslak is right in that it might make people use this in a different/unexpected way.

forki:
@amazingant `paket restore group Test` works now.

agross:
paket restore group --without Test would be useful as well (e. g. to exclude development dependencies) 



Alex
-- 
Alexander Groß
Tiny phone, tiny mail

forki:
@agross yes that sounds useful.

@eiriktsarpalis is it possible to have something like `paket restore groups Test Build references-files file1 file2`?  it seems I can only have one parameter with multiple things.

forki:
first draft of docs is at  http://fsprojects.github.io/Paket/groups.html

matthid:
I started testing this (alpha012) on RazorEngine and get:

```

nuget Microsoft.AspNet.Razor ~> 3.0

group Net40
  framework: net40
  
  nuget Microsoft.AspNet.Razor ~> 2.0

group Razor4
  nuget Microsoft.AspNet.Razor ~> 4.0.0-beta1
```

```
$.paket/paket.exe update                                                                                 
Paket version 2.0.0.0                                                                                    
 - Microsoft.AspNet.Razor 3.2.3                                                                          
Resolving packages for group Net40:                                                                      
Paket failed with:                                                                                       
        Could not find compatible versions for top level dependency:                                     
     "Microsoft.AspNet.Razor >= 3.0 < 4.0 (from c:\Projects\RazorEngine\paket.dependencies)"             
   Available versions:                                                                                   
     - 4.0.0-beta6                                                                                       
     - 4.0.0-beta5                                                                                       
     - 4.0.0-beta4                                                                                       
     - 4.0.0-beta3                                                                                       
     - 4.0.0-beta2                                                                                       
     - 4.0.0-beta1                                                                                       
     - 3.2.3                                                                                             
     - 3.2.3-beta1                                                                                       
     - 3.2.2                                                                                             
     - 3.2.2-rc                                                                                          
     - 3.2.0                                                                                             
     - 3.2.0-rc                                                                                          
     - 3.1.2                                                                                             
     - 3.1.1                                                                                             
     - 3.1.0                                                                                             
     - 3.1.0-rc1                                                                                         
     - 3.0.0                                                                                             
     - 3.0.0-rc1                                                                                         
     - 3.0.0-beta2                                                                                       
     - 3.0.0-beta1                                                                                       
     - 2.0.30506                                                                                         
     - 2.0.20715                                                                                         
     - 2.0.20710                                                                                         
     - 2.0.20505                                                                                         
     - 1.0.20105.408                                                                                     
   Try to relax the dependency or allow prereleases.                                                     
```


Notice that the following works (which is actually what I wanted to use anyway):
```

nuget Microsoft.AspNet.Razor

group Net40
  framework: net40
  
  nuget Microsoft.AspNet.Razor ~> 2.0

group Razor4
  nuget Microsoft.AspNet.Razor ~> 4.0.0-beta1
```

The difference is the "~>" restriction in the "Main" group. Was it wrong to expect the groups to be completely independent?

forki:
good catch - will try to fix

matthid:
Here is possibly another one: It seems like content from different groups (not referenced via 'paket.references') is added to projects. Repro is here https://gist.github.com/matthid/6a1d5f17f176d14d79b7

After
```
paket.bootstrapper.exe prerelease
paket.exe update
```
`git diff` shows that files from `Yaaf.AdvancedBuilding` were added to the project files, but `paket.references` only lists `FSharp.Core`

Further testing reveals that adding `content: none` doesn't help either (tried adding it to both "Main" and "Build")

gross:
I think you all accidentally added me to your project (my username is gross). Please remove me.

Thanks,

garrett ross ceo / garrett@mobelux.com <mailto://jeff@mobelux.com> / 757 254 7376 / mobelux.com <http://www.mobelux.com/>

amazingant:
@gross If you open the page for this issue (https://github.com/fsprojects/Paket/pull/1018) and click the "Unsubscribe" button on the right, that should do it. Your name got typo'd earlier because @agross is pretty similar.

forki:
@matthid first one is fixed

matthid:
@forki Nice, thanks!

Ok I did some more tests and was really impressed to get NO error after specifying
```
Microsoft.AspNet.Razor

group Net40
  Microsoft.AspNet.Razor
  
group Razor4
  Microsoft.AspNet.Razor
```
in `paket.references`. I think what `paket` does is correct even considering that the resulting `csproj` will not compile.

the first two are fine because one is `net45`-only and the other one is `net40`-only. But the third one is `net45` only as well.. This means I need some way to specify which one to take at build time (I don't want to use two references to the same nuget package on a single compilation).

One suggestion would be to allow something like
```
group Main if NET45
  Microsoft.AspNet.Razor

group Net40
  Microsoft.AspNet.Razor
  
group Razor4 if RAZOR4
  Microsoft.AspNet.Razor
```
in `paket.reference` (or even `paket.dependencies`, not sure right now) and then add a condition to the msbuild references like `Condition="$(RAZOR4) == 'True'`:

```
<When Condition="$(RAZOR4) == 'True' And $(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6')">
```

This is a feature request, but imho one that would make the group feature more useful.

Is this a scenario you want to tackle with the new group feature?

forki:
@matthid  second one fixed.

matthid:
Yes all that works now :) You are way too fast ;) 
Specifying a group named "Main" is also interesting: I thought it might be a way to get a `main` directory within the `packages` folder. But it seems to be a way to kind of break paket:

`paket.dependencies`
```
group Main

  nuget FSharp.Core

group Build

  nuget Yaaf.AdvancedBuilding
```

yields

`paket.lock`
```

GROUP Build
NUGET
  remote: https://nuget.org/api/v2
  specs:
    FAKE (4.0.2)
    FSharp.Compiler.Service (1.4.0.1)
    FSharp.Formatting (2.10.0)
      FSharp.Compiler.Service (>= 0.0.87)
      FSharpVSPowerTools.Core (1.8.0)
    FSharpVSPowerTools.Core (1.8.0)
      FSharp.Compiler.Service (>= 0.0.87)
    Yaaf.AdvancedBuilding (0.8.0)
      FAKE (4.0.2)
      FSharp.Formatting (2.10.0)
      FSharpVSPowerTools.Core (1.8.0)
```

forki:
@matthid I don't really understand https://github.com/fsprojects/Paket/pull/1018#issuecomment-135829321 - could you please create a new issue for discussion

I will take a look at the explizit main group case.

for further bugs please create separate issues. that makes it easier to handle for me.

forki:
explicit main group should work now, but it will not create a main directory. it will just put stuff in the main group.

forki:
we now support groups in `paket add` and `paket remove`

PR #1020, opened 2015-08-28
Use .NET4 version of NUnit test runners

rjnilsson:
Get rid of requirement for having .NET 3.5 installed in order to run tests (now tests only run under .NET 4.x or compatible).


forki:
cool. thx

forki:
BTW: Congratulations your are the 100th contributor to Paket!

t100

PR #1022, opened 2015-08-28
rework github strategy to use a more direct method in the common case…

baronfel:
… of just getting the latest version

### DIFF
--- a/src/Paket.Bootstrapper/GitHubDownloadStrategy.cs
+++ b/src/Paket.Bootstrapper/GitHubDownloadStrategy.cs
@@ -22,24 +22,36 @@ namespace Paket.Bootstrapper
 
         public string GetLatestVersion(bool ignorePrerelease)
         {
-            string latestVersion = "";
+            const string latest = "https://github.com/fsprojects/Paket/releases/latest";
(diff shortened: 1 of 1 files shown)


baronfel:
TODO:  pagination on the prerelease-case

forki:
I think there should be no pagination on the prerelease case. If there is no prerelease on the first page then there is no active prerelease.

baronfel:
ah, and the prerelease logic is really 'latest including prerelease', so we're really just grabbing the first of the 'all releases' page then?  that makes that logic simpler as well.

forki:
Yes we only want the super latest and greatest (and broken) version.

baronfel:
Alright, all better.

forki:
thanks!

forki:
sorry. I messed up.

The new bootstrapper doesn't work on prerelease case. It should always get the latest version, but github shows the last released version on top (even if that's older than the latest)

baronfel:
sounds like we should always get the latest version via the new mechanism, then, and then use that version to compare against if the user requests a prerelease?

so /releases/latest => 1.99.0
then use 1.99.0 to compare against
/releases =>
   1.99.0
   2.0.0--alpha1million

?

PR #1027, opened 2015-08-28
Update bootstrapper version parsing for github to handle prereleases

baronfel:
trying again!  now the prerelease takes the latest stable version as a parameter, and parses the releases html for the first non-stable version. if there is none on that page, then we default to the stable version that we already retrieved.

### DIFF
--- a/src/Paket.Bootstrapper/Paket.Bootstrapper.csproj
+++ b/src/Paket.Bootstrapper/Paket.Bootstrapper.csproj
@@ -46,6 +46,7 @@
     <Compile Include="NugetDownloadStrategy.cs" />
     <Compile Include="Program.cs" />
     <Compile Include="Properties\AssemblyInfo.cs" />
+    <Compile Include="SemVer.cs" />
   </ItemGroup>
   <ItemGroup>
     <None Include="paket.template" />
--- a/src/Paket.Bootstrapper/GitHubDownloadStrategy.cs
+++ b/src/Paket.Bootstrapper/GitHubDownloadStrategy.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Reflection;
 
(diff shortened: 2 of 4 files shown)


forki:
we only want the prerelease if it has higher version than stable

forki:
did you see #1041? that would also be cool to have.

baronfel:
I'll take a look. Also I'm planning on really fixing this one soon. Work work had me distracted for a few days.

mishfit:
Just a thought, have you looked into the [Github API for releases](https://developer.github.com/v3/repos/releases/#list-releases-for-a-repository) instead of scraping?

baronfel:
I seem to remember that paket did this in the past, but switched to scraping when it was discovered that we got rate limited quickly because paket didn't provide a way to use an api token as part of the request.

forki:
Yes that's the issue. Github api has very tight limits. At least when an
app is not using a github user to login.

mishfit:
Too bad, I was hoping to write it but it doesn't seem valuable

baronfel:
ideally what'd I'd be able to do here is use the SemVer module from Paket.Core to do the version comparisons, but at the moment Paket.Bootstrapper doesn't have any project references.  Would it be work pulling some of these paket-agnostic utilities into a separate project that could be referenced  independently?

baronfel:
Ok, here's a version of what it would look like if we pulled the SemVer parser out into a separate project that we could use.  Thoughts?

baronfel:
and for super bonus points it'd be wonderful if we had something like an env variable for github api authentication credentials, the presence of which would be a signal that we could use the API :)

Paket.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.31101.0
+# Visual Studio 14
```
forki: can you please revert that change?

src/Paket.Bootstrapper/Paket.Bootstrapper.csproj
```diff
@@ -33,6 +33,7 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="FSharp.Core, Version=4.4.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" />
```
forki: this is what I wanted to avoid. the bootstrapper was supposed to be tiny. now we would have to merge a bg lib into it.

baronfel:
Alright, I've changed over to a simplified change that uses the existing SemVer class in the nuget strategy.  I pulled out that class, made it public, and now things seem to work consistently for me.

forki:
Very cool. Will try out today

PR #1029, opened 2015-08-29
Add analyzers support

vbfox:
This PR add the support for the "analyzers" folder in NuGet packages.

They contain roslyn based analyzers. Packages can either contain only analyzers or normal libraries could include their analyzers along with the main lib.

The documentation details for the NuGet feature is still in flux https://github.com/NuGet/NuGetDocs/pull/362/files / https://github.com/jmarolf/NuGetDocs/blob/AnalyzerConventions/NuGet.Docs/Create/Analyzers-Conventions.md but as VS2015 support them, the current status is "Whatever works in VS2015".

Languages
---------------

Analyzers are contained either in per-language folders or shared between languages so I had to add code to get the language of a project file (Implemented using both the "ProjectTypeGuids" node and the file extension)

As the documentation specify that F# is a supported analyzer language I added it but the F# project type doesn't support analyzers for now :(

Framework
----------------

In theory their folders can also specify the framework of the host but nobody does that for now (All analyzers are in 'dotnet' subfolder as it is the platform supported by visual studio) but I ignored that completely.
if it is needed at some point the `paket.dependencies` files would need to specify it.

It seem to me that the feature of specifying the framework in there is broken anyway and microsoft didn't think long enough before introducing it.

Tests
-------

I don't know if there is enough tests, especially the code in `LanguageEvaluation` and my additions to `ExtractPackage` aren't tested. But I wanted to make this review to know if I was on the good track or if I did something really stupid somewhere.

To manually test the feature the easiest is to add 'RefactoringEssentials' as reference to any C# project and run `paket install`.

Then code like `int[] values = new int[] { 1, 2, 3 };` will trigger the "Array Creation can be replaced with Array Initializer" analyzer.


vbfox:
After a few hours setting up a VM to reproduce the Linux build problems I managed to reproduce it.

I'll debug tomorrow, something around paths is fishy.

forki:
This look like a big feature. What do you think if we put this into the paket 2.0 time frame?

vbfox:
No problem with having it in the paket 2.0 timeframe.

I created the feature originally in our FAKE script at work because I wanted to be able to force a series of analyzers on all projects. But I realized that I could simplify my life by teaching paket how to do that instead and just fixup the paket.references via FAKE when a new project is added.

(Then I discovered that due to the Install.ps1 being ignored in project.json based projects, nuget was also working on a similar feature as it is mandatory for analyzers to work)

As we don't use paket for the main repository (the group feature might solve some of the problems I had before trying to move to paket) it would be introduced just for that so we can live with 2.0-pre versions for the time being.

forki:
Ok. Cool. So could you please try to rebase this on the groups branch?

vbfox:
Github made a mess of the pull request after the rebase i'll recreate one

PR #1031, opened 2015-08-30
Add analyzers support

vbfox:
(Follows #1029 but the GitHub doesn't let me change the base branch so it got confused when I rebased from `master` to `groups`)

This PR add the support for the "analyzers" folder in NuGet packages.

They contain roslyn based analyzers. Packages can either contain only analyzers or normal libraries could include their analyzers along with the main lib.

The documentation details for the NuGet feature is still in flux https://github.com/NuGet/NuGetDocs/pull/362/files / https://github.com/jmarolf/NuGetDocs/blob/AnalyzerConventions/NuGet.Docs/Create/Analyzers-Conventions.md but as VS2015 support them, the current status is "Whatever works in VS2015".

Languages
---------------

Analyzers are contained either in per-language folders or shared between languages so I had to add code to get the language of a project file (Implemented using both the "ProjectTypeGuids" node and the file extension)

As the documentation specify that F# is a supported analyzer language I added it but the F# project type doesn't support analyzers for now :(

Framework
----------------

In theory their folders can also specify the framework of the host but nobody does that for now (All analyzers are in 'dotnet' subfolder as it is the platform supported by visual studio) but I ignored that completely.
if it is needed at some point the `paket.dependencies` files would need to specify it.

It seem to me that the feature of specifying the framework in there is broken anyway and microsoft didn't think long enough before introducing it.

Tests
-------

I don't know if there is enough tests, especially the code in `LanguageEvaluation` and my additions to `ExtractPackage` aren't tested. But I wanted to make this review to know if I was on the good track or if I did something really stupid somewhere.

To manually test the feature the easiest is to add 'RefactoringEssentials' as reference to any C# project and run `paket install`.

Then code like `int[] values = new int[] { 1, 2, 3 };` will trigger the "Array Creation can be replaced with Array Initializer" analyzer.

### DIFF
--- a/src/Paket.Core/InstallModel.fs
+++ b/src/Paket.Core/InstallModel.fs
@@ -12,6 +12,7 @@ type Reference =
     | Library of string
     | TargetsFile of string
     | FrameworkAssemblyReference of string
+    | Analyzer of string
 
     member this.LibName =
         match this with
@@ -40,6 +41,7 @@ type Reference =
         | Reference.Library path -> path
         | Reference.TargetsFile path -> path
         | Reference.FrameworkAssemblyReference path -> path
+        | Reference.Analyzer path -> path
 
 type InstallFiles = 
     { References : Reference Set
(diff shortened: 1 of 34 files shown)


forki:
kiss

I also created a stub for docs. Could you please fill in some details? Thx

https://github.com/fsprojects/Paket/commit/40d6ee24b44d8b4a4f102c0329d908b656ae62d2

PR #1032, opened 2015-08-31
Added documentation for Analyzers

vbfox:
Need to be re-read but I think that it contains enough for now.


forki:
cool stuff. thanks so much

PR #1034, opened 2015-09-01
Analyzers tests

vbfox:
Add some unit tests around all the detection code for the analyzers.

Also fix a typo in the docs


forki:
cats

thanks so much. We always need more tests.

Issue #1013, opened 2015-08-25
Install modifies local nuget packages


rjnilsson:
Given something like this in `paket.dependencies`

    source imports/nuget
    nuget MyPackage

When running `paket install` the MyPackage.nupkg file under imports/nuget is modified, as reported by my SCM tool. Calculating the MD5 checksum also gives different results before and after.

Using Paket 1.33.0 under Win 8.1/.NET 4.5.2.

forki:
yes it's fixing date convetions in the zip order to make unzip work in mono. we need to work around mono bugs here.

rjnilsson:
Ok. Might be a stupid question, but is it really necessary to modify the source package instead of the copy under packages/?

forki:
Mhm I think I agree, but the issue is we need to extract the meta data. So
we could (maybe should) copy the file to a temp folder and modify it over
there.

At the moment I'm really busy with other stuff, but I think this would make
a great up-for-grabs issue.

rjnilsson:
What about detecting the platform at runtime and only fixup when running under mono, just as a workaround for the time being?

forki:
Yes that would make even easier fix for you ;-)

rjnilsson:
I'll check it out. If only it'd been C# instead... ;)

forki:
If it would be C# it wouldn't be in the state that it is today. Believe me

rjnilsson:
I wasn't complaining about the language itself. Good or bad code can be
written in any language.

I was simply referring to my zero experience with F#, or rather, my decent
experience with C#.

Anyway, I'll take a look.
Den 25 aug 2015 15:52 skrev "Steffen Forkmann" <notifications@github.com>:

forki:
The code in question is at https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NuGetV2.fs#L324

you could port the isMono  function from https://github.com/fsharp/FAKE/blob/master/src/app/FakeLib/EnvironmentHelper.fs#L124 into Paket's Util.fs 

I think that's pretty straight forward. If you have question, then just ask.

rjnilsson:
I'll give it a shot. Thanks for the pointers.

rjnilsson:
Trying to build the project now, but when using build.cmd (default target) from the command-line I get the following error:

    Running build failed.
    Error:
    Fake.UnitTestCommon+FailedTestsException: NUnit test failed. Process finished with exit code FatalError (-2146232576).
[...]

PR #1036, opened 2015-09-02
Enable Mono runtime detection

rjnilsson:
Baby steps to work around #1013


forki:
looks good so far ;-)

PR #1039, opened 2015-09-02
Fix selective update for Paket 2.0

mrinaldi:


forki:
Thx for taking care of this. It's really really cool to have someone helping with this.

mrinaldi:
No man, thank *you*.
We're switching to paket and we are seeing our update-package workflow changing from a half hour to a few seconds to complete.

Updating a package with paket is like this:
giphy

PR #1044, opened 2015-09-04
Eagerly create app.config files and add to all projects.

isaacabraham:
Can someone review this, as it's quite an invasive change in terms of files that can be amended / created as a result of this. Whenever the binding redirects code now gets called, Paket will: -

1. Locate all paket.references files.
2. Create an very bare app.config file if one doesn't already exist.
3. Go through the standard app.config binding redirects code, applying BRs to both existing configuration files and those that were just created.
4. Add any new app.config files to any cs / fs / vbproj that it finds alongside it.

### DIFF
--- a/src/Paket.Core/BindingRedirects.fs
+++ b/src/Paket.Core/BindingRedirects.fs
@@ -82,10 +82,12 @@ let private applyBindingRedirects bindingRedirects (configFilePath:string) =
     config.Save configFilePath
 
 let private configFiles = [ "app"; "web" ] |> Set.ofList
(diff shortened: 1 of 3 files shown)


isaacabraham:
Hmmm. System.IO.Path behaves differently between Travis and AppVeyor?

isaacabraham:
OK, so Path.GetDirectoryName etc. use different formats between windows and unix. But on Windows it will eagerly turn e.g. "C:/foo/bar/" into "C:\\foo\\bar\\".

forki:
cool. let's see how this works in practice ;-)

pavelhritonenko:
Hm... but I don't need app.config for my class libraries.

Issue #1041, opened 2015-09-03
Paket bootstrapper always downloads prereleases even if paket.exe is latest


forki:
run paket.bootstrappper.exe prerelease 
==> downloads latest prerelease

run paket.bootstrappper.exe prerelease 
==> downloads latest prerelease again

baronfel:
Looks to be because we're setting the FileInfoVersion to be whole versions, not prereleases, in the assembly info of the prerelease exe.  So when we go fetch versions, the version to be downloaded is never equal to the version locally.

We could either
a) compare based on the non-prerelease parts of versions for the 'is same version' check, or
b) set prerelease versions in the prerelease executables and allow the check to occur as normal

I prefer (b) because if we compare on normalized versions then we'll never download subsequent versions of a particular major/minor/patch version. For example,  2.0.0-beta005 and 2.0.0-beta006 would both compare down to 2.0.0 in scheme (a).

What do you want to do?

baronfel:
probably need to look at FAKE's Release Notes helper, because that's how Paket sets the fileversion number.  If it only parses stable version numbers, then we'll need to think of another solution.

baronfel:
however! we do set the [informational version](https://github.com/fsprojects/Paket/blob/master/build.fsx#L94) at build/release time, so if we switch to reading that, then we should have all the detail we need?

PR #1049, opened 2015-09-06
Use AssemblyInformationalVersionAttribute to do version comparisons in the bootstrapper for remote vs local

baronfel:
We set the informational version on release to the same version as the nuget package/github release.
So for purposes or determining if the local version is the same as the discovered version from the download strategy we should compare against the informational version.

This implements the quick-escaping behavior requested by #1041

### DIFF
--- a/src/Paket.Bootstrapper/BootstrapperHelper.cs
+++ b/src/Paket.Bootstrapper/BootstrapperHelper.cs
@@ -1,7 +1,8 @@
 ﻿using System;
-using System.Diagnostics;
 using System.IO;
 using System.Net;
+using System.Linq;
+using System.Reflection;
 
 namespace Paket.Bootstrapper
 {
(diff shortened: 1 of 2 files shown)


forki:
does  c842c5d belong into this PR?

baronfel:
Augh, no. Let me clean that up. I was working on something for a different
PR and got my branches twisted.

baronfel:
ok, I've removed that commit from this branch.

forki:
Cool. Thanks a lot

forki:
this introduced blocking of paket.exe - I solved it in 7906b735c09d4708ae9c76aa9d5cc8a252c20062

could you please review?

baronfel:
Yes, this is a better way to read the assembly. Thanks!

Also it makes me a little sad that we have to read an entire ~3MB file to pull out a version string.

PR #1050, opened 2015-09-07
address #1014 by adding a warning and example

baronfel:
I tried to address the concerns that @konste brought up in that issue.  Could you sanity check me?


forki:
thanks so much!

baronfel:
because there is no linking when the number is put in the title of a PR: fixes #1014

PR #1051, opened 2015-09-07
Fixes analyzers that were never removed during updates

vbfox:
The issue is #1047 the nodes where created with `<Paket>True</Paket>` markers but they weren't removed as all others.


forki:
cool!

PR #1052, opened 2015-09-07
Support for authentication and complex hosts.

colinbull:
This P.R adds support for authentication to HTTP dependencies and also prevents HTTP dependencies failing with an invalid path exception is the authority part of the URL is a IP or a DNS +  PORT. 

Couple of things to note. I have added an additional parameter to the http spec that allows you to specify the sourceName. This is the key that is used to look up the credentials in the auth store. And example of the full spec is now 

    http http://10.235.123.16:19/test1.fs test1.fs MySource

This will use the `MySource` as the auth key, and the folder name should become `10.235.123.16_19/MySource`. 

Also the documentation has been updated to reflect this change.


forki:
awesome

PR #1054, opened 2015-09-10
Fixes a bug where install and restore use different paths when specifying a project spec on a HTTP link

colinbull:
As the title says really.


forki:
thx.

Issue #1057, opened 2015-09-11
Pack fails to find project output path when no spaces around MSBuild Condition operator


ascjones:
Paket pack was failing in some cases with the error:
```
Unable to find Release output path node in file C:\code\MySolution\MyProject.fsproj
```
It appears some project templates we use produce a Condition where there are no spaces around the operator.

**Fails to parse:**
```
'$(Configuration)|$(Platform)'=='Release|AnyCPU'
```
**Parses okay (spaces around operator)**
```
'$(Configuration)|$(Platform)' == 'Release|AnyCPU'
```
Simply adding the spaces fixed the issue, however seems like this is a bug since Visual Studio and [MSBuild](https://msdn.microsoft.com/en-us/library/7szfhaft.aspx) appear to handle this okay.

This problem seems to have been introduced in v1.33 as a result of this commit:
https://github.com/fsprojects/Paket/commit/d2b731f94a19ce8cf186569e545b7ad66eb4e67c

I've made a quick fix and will submit a PR.

PR #1058, opened 2015-09-11
fix for parsing output path when condition has no spaces

ascjones:
Fix for issue: https://github.com/fsprojects/Paket/issues/1057

The tests pass, but @amazingant (who wrote the parsing code) may need to double check that to see that it doesn't break any other scenarios.

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -633,15 +633,16 @@ type ProjectFile =
                         |     _, false, c    -> parseWord (data.Append(c)) input (index + 1) false
 
                 let rec parseComparison (data:System.Text.StringBuilder) (input:string) index =
+                    let isCompChar c = c = '<' || c = '>' || c = '!' || c = '='
                     if input.Length <= index
                     then None
                     else
                         let c = input.[index]
                         if data.Length = 0 && c = ' '
                         then parseComparison data input (index + 1)
-                        elif data.Length = 2 && c <> ' ' && c <> '\''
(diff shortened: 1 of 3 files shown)


src/Paket.Core/ProjectFile.fs
```diff
@@ -639,7 +639,7 @@ type ProjectFile =
                         let c = input.[index]
                         if data.Length = 0 && c = ' '
                         then parseComparison data input (index + 1)
-                        elif data.Length = 2 && c <> ' '
+                        elif data.Length = 2 && c <> ' ' && c <> '\''
```
amazingant: What you wrote here should work fine for #1057.
If I had thought of this when I wrote the parser, I might have written this as something like this:

```FSharp
let isCompChar c = c = '<' || c = '>' || c = '!' || c = '='
...
elif data.Length = 2 && isCompChar c
```

(helper function so the next `elif` could be simplified as well)
If you want to do that with this PR, let me know how the tests go; otherwise this looks good to me :+1:

ascjones: Indeed you're right that is a significantly less naive approach. Pull request amended.

ascjones:
Travis build fails, but only because it fails to install mono due to some apt-get 404s:

```The command "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget referenceassemblies-pcl" failed and exited with 100```

#itbuildsonmymachine

forki:
yes same on FAKE build see https://github.com/fsharp/FAKE/pull/949

forki:
thx

PR #1061, opened 2015-09-14
Add support for custom internet proxy credentials with env vars

dedale:
At work our internet proxy uses custom credentials (different from Active Directory standard credentials). E.g AD login is like sforkman150915 but proxy login is steffen.forkmann.

By default, Paket and bootstrapper set default credentials in IWebProxy. They fail with 407 if I want to use them. I've seen other strange behaviors like partial nuget conversion. I cannot use FsReveal for the same reasons (407!)

At work some of our tools use the following env vars: http_proxy, https_proxy and no_proxy (bypass list).

So I added support for these variables in Paket and bootstrapper.

I did it twice (once in C# for bootstrapper and once in F# for Paket Utils). I could not come with a pretty solution to factor this.

I tried to add as many tests as I could (did not test all the possibilities though)

Please review this commit. I'm not very fluent in F# yet.
I am wondering if env var are case sensitive in mono environment.

See also: https://msdn.microsoft.com/fr-fr/library/Hh272656.aspx

Next step: try to convince my team to ditch nuget...


forki:
@kot_2010 could you please take a look if this would still work for you? Thx

forki:
Seems the tests fail on TravisCI / mono ;-)

forki:
and could you please rebase the PR? thx

vasily-kirichenko:
Sorry, I'm on a vocation till 23/09.

dedale:
I will rebase the PR.
I need to setup some linux distro in VirtualBox to fix the mono failures (I dont think I can reproduce on Windows)

dedale:
I'm trying tro reproduce travis/mono failures.
I have installed the following ubuntu packages: `fsharp 4.0.0.3-0xamarin1`, `mono-complete 4.0.4.1-0xamarin1`.
Unfortunately, building paket fails with very strange errors:

image

dedale:
Well problem disappeared after a couple of reinstallations and playing with alpha and beta mono channels. Not understood... No I can reproduce my bugs on mono.

dedale:
Env vars ARE case sensitive under mono environment. Should be fixed now (too late to wait for CI builds)... :zzz:

dedale:
:v: wow this was fast!

forki:
danger

thanks. Let's see how this works in practice ;-)

forki:
I guess we could need some docs for this.

dedale:
Sure! I guess you're talking about http://fsprojects.github.io/.
Where do you think the docs can be added?
In the **FAQ**? e.g. *How to use Paket behind an internet proxy?*
In the **Documentation** section? (adding a topic *Internet Proxy support*)

forki:
A faq entry would be a good start. Please also tell something about the
default behavior.

dedale:
OK will do :-)

Issue #711, opened 2015-03-20
How should redirects be created when a package contains different assembly versions for a single assembly?


agross:
depfile:
```
nuget Sample
```

`Sample.nuspec` contains  `<dependency id="Spark" version="[1.8,2.0)" />`.

Spark 1.8.0 contains `lib\NET45\Spark.dll` which has
```
[assembly: AssemblyVersion("1.7.0.0")]
[assembly: AssemblyFileVersion("1.7.5.3")]
[assembly: AssemblyInformationalVersion("1.7.5.3")]
```

As far as I know the [version the CLR binds to is `AssemblyVersion`](http://stackoverflow.com/q/64602/149264), `1.7.0.0` in this case.

Paket generates this version redirect, though:
```xml
<dependentAssembly>
  <assemblyIdentity name="Spark" publicKeyToken="7f8549eed921a12c" culture="neutral" /> 
  <bindingRedirect oldVersion="0.0.0.0-1.8.0.0" newVersion="1.8.0.0" />
</dependentAssembly>
```

isaacabraham:
Hmmm. We *use* AssemblyVersion to get the version used in the binding redirects file though. Are there other versions of the assembly in the nuget package, which have a different version perhaps? Maybe the binding redirects code is using one of those (incorrectly).

agross:
WTF, you were right.

`lib\NET35\Spark.dll`: `1.8.0.0`
`lib\NET40\Spark.dll`: `1.8.0.0`
`lib\NET45\Spark.dll`: `1.7.5.3`

My project is 4.5. Seems like the redirects will break when you change the .NET version of your project. At least when package authors decide to use different version numbers for different platforms.

agross:
It's alread reported. https://github.com/SparkViewEngine/spark/issues/54

agross:
Question is, how would we handle such situations. Should we rather throw or warn and make it explicit that redirects will break than just picking any version?

isaacabraham:
Yeah, the part of code that determines which version of the assembly to use to pull out the AssemblyVersion isn't aware of which version will ultimately be deployed by Paket. If that was added in, it then might just work.

agross:
For sure. But only until you change e.g. from 4.5.1 to 4.0 without an additional `paket install`. :-)

isaacabraham:
Could we do redirects that are framework-dependent I wonder?

agross:
I'm not aware of app.config settings depending on the framework version. 




Alex
[...]

Issue #595, opened 2015-02-11
Wrong Assembly Binding Redirect


dnauck:
When [Redirects option](http://fsprojects.github.io/Paket/dependencies-file.html#Redirects-option) is on and you install the package ```Microsoft.Net.Http``` in a .NET 4.5 project you'll get a wrong binding redirect.

The package contains 2 extensions assemblies for .NET 4.5 and 3 for .NET 4.0.
During install on a .NET 4.5 project it will also add the redirect for the ```System.Net.Http``` assembly that is already in the GAC (and not in the package's 4.5 folder, only in the package's 4.0 folder).

Result is a assembly not found exception.

isaacabraham:
@forki - I think that this is due to the way that we pick the appropriate assembly when there are multiple versions in the package - it was smartened a bit with a previous issue (can't remember which one) but looks like it needs some more work.

isaacabraham:
Just ran an empty net45 project with the single paket dependency ``Microsoft.Net.Http``. I got the following binding redirects: -
```xml
    <runtime>
        <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
            <dependentAssembly>
                <assemblyIdentity name="System.IO" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-2.6.9.0" newVersion="2.6.9.0" />
            </dependentAssembly>
            <dependentAssembly>
                <assemblyIdentity name="System.Runtime" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-2.6.9.0" newVersion="2.6.9.0" />
            </dependentAssembly>
            <dependentAssembly>
                <assemblyIdentity name="System.Threading.Tasks" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-2.6.9.0" newVersion="2.6.9.0" />
            </dependentAssembly>
            <dependentAssembly>
                <assemblyIdentity name="System.Net.Http.Extensions" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
                <bindingRedirect oldVersion="0.0.0.0-2.2.28.0" newVersion="2.2.28.0" />
            </dependentAssembly>
            <dependentAssembly>
[...]

PR #1070, opened 2015-09-18
Binding redirects from target platform only

### LINKED ISSUE #1042: Binding redirects filtering
Some of our projects reference [Microsoft BCL Portability Pack](https://www.nuget.org/packages/Microsoft.Bcl/) NuGet (Indirectly but that's not the point) the problem is in the generation of the binding redirects.
```
redirects: on
```

Any project referencing this package will have this redirect generated :
```xml
  <dependentAssembly>
    <assemblyIdentity name="System.Runtime" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
    <bindingRedirect oldVersion="0.0.0.0-999.999.999.999" newVersion="2.6.10.0" />
  </dependentAssembly>
```

Version `2.6.10.0` is the current version in silverlight and a few other platforms but our projects are all net40 or net45 and in theses there correct version is of a higher version number (The NuGet package contains no file at all in the corresponding folders)

So I think that Paket is either missing a way to disable redirects for one package or some way to tell it that he should trust the framework in the `.csproj` file (So he could generate only the required Reference and only the required redirects)

mrinaldi:
This PR fixes #1042, #595, #813, #703, #711.

It'll only add the binding redirects for the assemblies of the project's target platform.

### DIFF
--- a/src/Paket.Core/BindingRedirects.fs
+++ b/src/Paket.Core/BindingRedirects.fs
@@ -68,19 +68,6 @@ let internal indentAssemblyBindings config =
     let newAssemblyBindingNode = XElement.Parse(sb.ToString(), LoadOptions.PreserveWhitespace)
     parent.Add(newAssemblyBindingNode)
 
-/// Applies a set of binding redirects to a single configuration file.
-let private applyBindingRedirects bindingRedirects (configFilePath:string) =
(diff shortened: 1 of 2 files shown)


forki:
thanks so much.

Issue #1074, opened 2015-09-21
NullReferenceException in applyBindingRedirects during "update nuget package"


vbfox:
Just hapenned to me on 2.3.0.0

```
c:\Code\project>paket update nuget stidgen
Paket version 2.3.0.0
Updating stidgen in c:\Code\project\paket.dependencies group Main
Resolving packages for group Main:
 - stidgen 0.5.0
Locked version resolution written to c:\Code\project\paket.lock
Downloading stidgen 0.5.0

Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at <StartupCode$Paket-Core>.$ProjectFile.framework@684.Invoke(FSharpOption`1 o)
   at Microsoft.FSharp.Collections.IEnumerator.map@111.DoMoveNext(b& )
   at Microsoft.FSharp.Collections.IEnumerator.MapEnumerator`1.System-Collections-IEnumerator-MoveNext()
   at Microsoft.FSharp.Collections.SeqModule.TryHead[T](IEnumerable`1 source)
   at Paket.ProjectFile.GetTargetProfile()
   at Paket.BindingRedirects.applyBindingRedirects(FSharpFunc`2 bindingRedirects, String configFilePath)
   at Paket.BindingRedirects.applyBindingRedirectsToFolder@146-3.Invoke(String configFilePath)
   at Microsoft.FSharp.Collections.SeqModule.Iterate[T](FSharpFunc`2 action, IEnumerable`1 source)
   at Paket.InstallProcess.InstallIntoProjects(InstallerOptions options, DependenciesFile dependenciesFile, LockFile lockFile, FSharpList`1 projects)
   at <StartupCode$Paket-Core>.$PublicAPI.UpdatePackage@192-3.Invoke(Unit unitVar0)
   at Paket.Utils.RunInLockedAccessMode[a](String rootFolder, FSharpFunc`2 action)
   at Paket.Program.handler@300-15.Invoke(ParseResults`1 results)
   at Paket.Program.processWithValidation[T](FSharpFunc`2 validateF, FSharpFunc`2 commandF, Command command, String[] args)
   at Paket.Program.processCommand@66-1.Invoke(Command command, String[] args)
   at Paket.Program.main()
   at <StartupCode$paket>.$Paket.Program.main@()
```

Part of the `paket.dependencies` :

```
framework: >= net40

#############################################################################
# Official NuGet package repository
source https://nuget.org/api/v2

nuget stidgen
```

[...]

PR #1076, opened 2015-09-21
Fix the nullref in applyBindingRedirects without try/with

vbfox:
Fix differently #1074 by using Seq.choose instead of a try/with


forki:
Cool thanks

Issue #1072, opened 2015-09-21
paket.bootstrapper.exe fails on Ubuntu


NatElkins:
paket.bootstrapper.exe fails to download paket.exe on Ubuntu using Mono.  See screenshot:

![screenshot from 2015-09-20 21 37 58](https://cloud.githubusercontent.com/assets/1795462/9984126/3768eef4-5fe0-11e5-9431-b5a5da70c24a.png)

I downloaded the solution and debugged using MonoDevelop.  Looks like the error occurs in the download strategies (GithubDownloadStrategy and NugetDownloadStrategy) whenever the WebClient tries to download anything.  The error is specifically thrown at https://github.com/fsprojects/Paket/blob/master/src/Paket.Bootstrapper/GitHubDownloadStrategy.cs#L48 and https://github.com/fsprojects/Paket/blob/master/src/Paket.Bootstrapper/NugetDownloadStrategy.cs#L72.

Here's the stack trace:

[...]

PR #1077, opened 2015-09-22
Changing ServerCertificateValidationCallback to accept all certificat…

NatElkins:
…es as valid, as default Mono implementation does not accept github.com and nuget.org certificates as valid.

### DIFF
--- a/src/Paket.Bootstrapper/Program.cs
+++ b/src/Paket.Bootstrapper/Program.cs
@@ -11,6 +11,19 @@ namespace Paket.Bootstrapper
 {
     class Program
     {
+        static Program()
+        {
+            // The default Mono implementation does not certify
+            // the github.com or nuget.com certificates as valid.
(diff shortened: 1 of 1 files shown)


NatElkins:
This is to fix https://github.com/fsprojects/Paket/issues/1072

vbfox:
Please don't merge this, it would be a reason for us to go back to nuget and not use paket anywhere.

Any tool downloading and executing binaries should check for where and what it is downloading as best as it can and have a very clean security story. Especially things like this bootstraper that are executing very often. Any MITM attack would auto succeed for any user of paket bootstraper.

Defaulting to no security is the worse that can happen.

There are technical solutions that can be done (Disabling ssl certificate security and relying on a custom signature) but it's an order of magnitude more complex and anyway the sll certificate checking should still be on by default even with that.

forki:
People suggested to use `mozroots --import --sync` instead.

Issue #1079, opened 2015-09-22
Paket fails to get package versions (2.1.0 - 2.3.4)


MiloszKrajewski:
paket outdated fails with:

```
Could not find versions for package DotNetZip.Reduced in any of the sources in [Nuget {Url = "https://www.nuget.org/api/v2"; Authentication = null;}].
```

Thank to paket.bootstrapper I was able to check out multiple versions. So it works fine up to 2.0.3, then starts to fail at 2.1.0 up to 2.3.4 (the most recent version at the time I'm writing this). It might be important that I am behind proxy but even then the older version works.

```
PS D:\prj\external\lz4net\src> .\paket.cmd outdated --verbose
No version specified. Downloading latest stable.
Paket.exe 2.3.4.0 is up to date.
Paket version 2.3.4.0
found: D:\prj\external\lz4net\src\paket.dependencies
Parsing D:\prj\external\lz4net\src\paket.dependencies
Resolving packages for group Main:
  0 packages in resolution. 4 requirements left
  Trying to resolve DotNetZip.Reduced  (from D:\prj\external\lz4net\src\paket.dependencies)
  - fetching versions for DotNetZip.Reduced
getAllVersionsFromNuGet2 from url 'https://www.nuget.org/api/v2/package-versions/DotNetZip.Reduced?includePrerelease=true'
getAllVersionsFromNugetOData from url 'https://www.nuget.org/api/v2/FindPackagesById()?id='DotNetZip.Reduced''
getAllVersionsFromNugetODataWithFilter from url 'https://www.nuget.org/api/v2/Packages?$filter=Id eq 'DotNetZip.Reduced''
Paket failed with:
        Could not find versions for package DotNetZip.Reduced in any of the sources in [Nuget {Url = "https://www.nuget.org/api/v2";
        Authentication = null;}].
StackTrace:
     at Paket.NuGetV2.GetVersions@651.Invoke(String message)
   at Paket.NuGetV2.GetVersions(String root, IEnumerable`1 sources, PackageName _arg1)
   at Paket.PackageResolver.getAllVersions@235(FSharpFunc`2 getVersionsF, Dictionary`2 allVersions, FSharpList`1 sources, PackageName packageName, VersionRange vr)
[...]

PR #1080, opened 2015-09-22
Fix unsupported https scheme in web proxy

dedale:
Finally found a way to push the fix.
Hope it solves #1079


forki:
cool thanks. 
I fixed the red test cases. could you please review f796058
? thx

dedale:
Sorry for the unit tests. You just need to fix the scheme, not the port.
Updated my PR BTW.

dedale:
Mmm If i look at AppVeyor logs, it seems it automatically replaces http://proxy.local:443/ with https://proxy.local/

PR #1083, opened 2015-09-23
Fixing failing test on mono due to NewLine differences

colinbull:
Title says it all really.


colinbull:
Hummm!! Actually this seems to be something with my mono install I'll investigate further...

PR #1090, opened 2015-09-25
Update paket-folder.md

konste:
Added clarification about Paket directory structure relative to projects and solutions.


forki:
`thanks a lot

PR #1091, opened 2015-09-27
Improve error message for bootstrapper if download of Paket.exe fails…

NatElkins:
…; added hint to look at 'mozroots' tool for users of Mono.

This is in relation to this issue: https://github.com/fsprojects/Paket/issues/1072


forki:
thanks

Issue #1081, opened 2015-09-22
Regression: empty <runtime> section generated in config files


konste:
While executing "paket.exe install" command for all projects affected paket adds empty section 
<runtime><assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1" /></runtime>
even though we have "redirects: off" at the top of paket.dependencies file.

I don't remember seeing it with the version from a couple weeks ago, so it looks like regression.

isaacabraham:
Not sure how that can be then - the changes related to adding new config file etc. only run from within the redirects code as far as I know.

Maybe this has to do with the changes to allow redirects against dependencies?

konste:
-- Maybe this has to do with the changes to allow redirects against dependencies?

As far as I understand that's what it is.

isaacabraham:
I have no clue where that change came from - happy to have a look around through the code base when I get some time.

konste:
It is not urgent. Let me know if it does repro for you.

konste:
@mrinaldi I think it is your change to BindingRedirects.fs 5 days ago introduced that regression. The problem is in applyBindingRedirects method. It updates configuration files even when no binding redirects required there or binding redirects globally disabled. 
The issue is cosmetic, but very annoying. 
I don't know F# unfortunately, so I cannot provide meaningful PR. Could you please fix it?

Thank you!
Konstantin

mrinaldi:
Sorry @konste, I'm on vacation and I can't access my computer in the meanwhile.

Anyway, thanks for fixing it.

PR #1092, opened 2015-09-28
Proposed fix for #1081 

konste:
Do not touch config files when the list of binding redirects to add is empty.

Caution: it is the first time I see F# and Git.


forki:
LGTM thanks

PR #1101, opened 2015-09-30
Memoize to reduce web requests to external sites like GitHub.

Thorium:
I had some problems with GitHub:
System.Net.WebException: The remote server returned an error: (403) Forbidden.
"API rate limit exceeded"
So after this it shouldn't make duplicate network calls when resolving versions inside one Paket-command (like paket outdated or paket update).


forki:
did you see where it repeatably calls the same url?

Thorium:
No, I didn't try to track with Fiddler.

If I blindly add counter to this memoize function, it will be hit 239 times with this paket.dependencies:

		nuget FAKE
		nuget FSharp.Formatting
		nuget FSharp.Data
		nuget Microsoft.AspNet.SignalR.Core
		nuget Microsoft.AspNet.WebApi.OwinSelfHost
		nuget Microsoft.AspNet.Identity.Owin
		nuget Microsoft.Owin.Cors
		nuget Microsoft.Owin.StaticFiles
		nuget Microsoft.Owin.Diagnostics
		nuget Microsoft.Owin.Security.Facebook
		nuget Microsoft.Owin.Security.Google
		nuget Owin.Security.AesDataProtectorProvider
		nuget MySql.Data
		nuget Rx-Main
		nuget SQLProvider
		nuget SourceLink.Fake

		github zurb/bower-foundation css/foundation.css
		github zurb/bower-foundation css/normalize.css
		github zurb/bower-foundation js/foundation.min.js
		github SignalR/bower-signalr jquery.signalR.js
		github aFarkas/html5shiv dist/html5shiv.min.js
		github reactjs/react-bower react.js
		github FortAwesome/Font-Awesome css/font-awesome.min.css
		github FortAwesome/Font-Awesome fonts/fontawesome-webfont.eot
		github FortAwesome/Font-Awesome fonts/fontawesome-webfont.svg
		github FortAwesome/Font-Awesome fonts/fontawesome-webfont.ttf
		github FortAwesome/Font-Awesome fonts/fontawesome-webfont.woff
		github FortAwesome/Font-Awesome fonts/fontawesome-webfont.woff2
		github danielm/uploader src/dmuploader.min.js
		github danielm/uploader demos/js/demo-preview.js
		github booncon/slippry dist/slippry.css
		github booncon/slippry dist/slippry.min.js
		github booncon/slippry images/arrows.svg
		github booncon/slippry images/sy-loader.gif
		github lodash/lodash lodash.min.js

		http http://ajax.aspnetcdn.com/ajax/jQuery/jquery-2.1.4.min.js jquery.min.js
		http https://code.jquery.com/ui/1.10.4/jquery-ui.min.js
		http https://code.jquery.com/ui/1.10.4/themes/sunny/jquery-ui.min.css

239 requests are not so much but if those (or like 20% of them) are happening at exact the same time to the same web-site, the firewall may not like that. I assume that this .NET concurrent dictionary will not just remove the dublicate calls, but also reduce the parallel request count by slowing down the process a few milliseconds (as it syncs between the threads) so that may also help GitHub to not block the user.

forki:
For me that sounds like you found another perf issue. Will investigate
tomorrow. Can't believe it's correct to make so many calls.

src/Paket.Core/Utils.fs
```diff
@@ -196,6 +196,12 @@ let inline createWebClient(url,auth:Auth option) =
     client.Proxy <- getDefaultProxyFor url
     client
 
+//More generic variant of http://www.fssnip.net/c4
+open System.Collections.Concurrent
+let cache = ConcurrentDictionary<(string * obj) option,Lazy<obj>>()
+let memoizeConcurrent (caller:string) (f: ('a -> 'b)) =
+    fun (x :'a) -> 
+        (cache.GetOrAdd(Some (caller, x|>box), lazy ((f x)|>box)).Force() |> unbox) : 'b
```
forki: why the `lazy ((f x) |> box)).Force()`?
why not `(f x) |> box)`?

Why is the key `(string * obj) option` and not just `(string * obj)`?

forki: are you sure it's really caching?

Thorium:
Lazy is due to the reason that ConcurrentDictionary is locking the threads during the update-action. So lazy will cause threads to be not locked while f is executed.

Just tuple for the key should do well also. I just took the simplest and best memoize from fssnip.net to be the base implementation and improved that.

You can test this code with Tomas Petricek's Fibonacci, just replace his non-threadsafe memoize command with memoizeConcurrent: http://www.fssnip.net/8P

forki:
I as able to understand the root cause.
your deps file specifies multiple files of the same repo. so we look up the hash for every file. that's always the same hash for files of the same repo -> we can cache that.

I changed your strategy to only cache the hashes (and possible remote dependencies files). Please review https://github.com/fsprojects/Paket/commit/ad109f36347de219b7bc91a9536c33395a0a7c5e

Thorium:
A bit more verbose code. 
But as long as I'm not banned by GitHub, I'm happy.
So this is also fine.

forki:
yeah. and it doesn't do full memoization. I don't really like that because it's harder to explain later. It's "just" cache lookup.

Anyways it saves you 19 github calls per update.

forki:
I was able to reduce the calls by checking if we already downloaded the corresponding dependencies file in the right version. please review https://github.com/fsprojects/Paket/commit/f40234520c976b0518028332b3eea2a76756cdc0

forki:
(nice side effect it's also much faster for the second update)

PR #1102, opened 2015-10-01
don't add -prerelease to nuspec dependency nodes for project references

zzdtri:
Pack is adding -prerelease to the end of project reference dependency nodes, e.g:

    <dependency id="BlahBlah.BlahBlah" version="1.0.0-prerelease" />

when that project has been packed correctly as 1.0.0.

Fairly easy to repro - have two projects, both with paket.template files (type project), and one referencing the other.   Paket pack version 1.0.0 will generate the above.


forki:
thanks. That's a bad bug!

PR #1103, opened 2015-10-02
Find the mono binary on OSX 10.11

nosami:
It is not in $PATH anymore if System Integrity Protection is on.

Paket works fine from the command line, but not when it's ran from a GUI app.

This PR is essentially the same as [this one for Nuget](https://github.com/fsharp/fsharp/pull/492)


forki:
What GUI are you talking about?

forki:
Thanks for caring and fixing

nosami:
Xamarin Studio specifically.... but this would affect any editor that integrates Paket also that isn't launched from the command line.
screen shot 2015-10-02 at 08 42 48

forki:
I assume we need similar thing for FAKE. Any volunteers!?

nosami:
I can do it

PR #1105, opened 2015-10-03
Ignore disabled upstream feeds

theimowski:


forki:
any ideas why the mono build fails on the sample?

forki:
I guess it was https://github.com/fsprojects/Paket/commit/6cce93377a84d285fbd741de7755c128851ec1cd#diff-86117e2b5523a68369ab3089fdcff773R79

thanks for contributing!

theimowski:
ah didn't know Mono doesn't like these - do you know exact reason why that happens?

forki:
Yes. Filename is case-sensitive. You named the folder "NuGet..." but wrote
"Nuget..." in test. It just didn't find the file...

theimowski:
ok I understand know - didn't notice case sensitive file names. 
you also changed `preservenewset` to `copyalways` - is there any reason for that why mono would fail otherwise?

forki:
The copyalways was only my first thought. But it wasn't the reason.
Everything was ok, only the file was not good on nix

theimowski:
I see, glad I could be of help!

PR #1108, opened 2015-10-03
Selective update needs to consider remote files

forki:
/cc @mrinaldi 

It now works like "paket install". But instead of discovering changes from last lock file, we just add all transitive dependencies of the package in question.

### DIFF
--- a/src/Paket.Core/UpdateProcess.fs
+++ b/src/Paket.Core/UpdateProcess.fs
@@ -23,14 +23,12 @@ type UpdateMode =
             | Some(groupName,package) -> SelectiveUpdate(groupName,package)
 
 let selectiveUpdate resolve (lockFile:LockFile) (dependenciesFile:DependenciesFile) updateAll package =
-    let resolve (dependenciesFile : DependenciesFile) packages =
-        match package with
-        | None -> dependenciesFile.Groups
-        | Some (groupName,_) -> dependenciesFile.Groups |> Map.filter (fun g _ -> g = groupName)
+    let resolve (dependenciesFile : DependenciesFile) =
+        dependenciesFile.Groups
         |> Map.map (fun groupName group ->
             { Name = group.Name
               RemoteFiles = group.RemoteFiles
-              RootDependencies = packages
+              RootDependencies = None
               FrameworkRestrictions = group.Options.Settings.FrameworkRestrictions
               PackageRequirements = 
                 match package with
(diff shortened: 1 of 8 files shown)


src/Paket.Core/UpdateProcess.fs
```diff
@@ -23,14 +23,12 @@ type UpdateMode =
...
-        match package with
-        | None -> dependenciesFile.Groups
-        | Some (groupName,_) -> dependenciesFile.Groups |> Map.filter (fun g _ -> g = groupName)
+    let resolve (dependenciesFile : DependenciesFile) =
+        dependenciesFile.Groups
         |> Map.map (fun groupName group ->
             { Name = group.Name
               RemoteFiles = group.RemoteFiles
-              RootDependencies = packages
+              RootDependencies = None
```
forki: I wonder if we still need that record field...

mrinaldi:
This test fails:

```
let lockFileData7 = """NUGET
  remote: http://nuget.org/api/v2
  specs:
    Package (3.2.0)
    Newtonsoft.Json (6.0.8)
"""
let lockFile7 = lockFileData7 |> getLockFile

let graph7 = 
    [ "Package", "3.2.0", []
      "Package", "4.0.0", ["Newtonsoft.Json", VersionRequirement(VersionRange.AtLeast "7.0.0",PreReleaseStatus.No)]
      "Newtonsoft.Json", "7.0.1", []
      "Newtonsoft.Json", "6.0.8", [] ]
      
[<Test>]
let ``SelectiveUpdate updates package that has a new dependent package that also is a direct dependency``() = 

    let dependenciesFile = DependenciesFile.FromCode("""source http://nuget.org/api/v2

    nuget Newtonsoft.Json
    nuget Package""")
    
    let resolve = resolve' graph7

    let lockFile = 
        UpdateProcess.UpdateMode.UpdatePackage(Constants.MainDependencyGroup, PackageName "Package")
        |> selectiveUpdate resolve lockFile7 dependenciesFile
    
    let result = 
        lockFile.GetGroupedResolution()
        |> Map.toSeq
        |> Seq.map (fun (_,r) -> (string r.Name, string r.Version))

    let expected = 
        [("Package","4.0.0");
        ("Newtonsoft.Json","7.0.1")]
        |> Seq.sortBy fst

    result
    |> Seq.sortBy fst
    |> shouldEqual expected
```
```
Expected: seq [("Newtonsoft.Json", "7.0.1"); ("Package", "4.0.0")]
Actual: seq [("Newtonsoft.Json", "6.0.8"); ("Package", "3.2.0")]
```

It's because now you pin all dependencies that are not dependencies from the package's current version.
When a new dependency is introduced and that dependency is already a dependency in your graph, it won't let you update.

The other code would not pin, but would only resolve the updating package's graph, creating package requirements from lock file (from packages that are not in the updating package's graph) which takes allowed version range into consideration.

BTW, why does it need to work like "paket install"? I mean, what's the bug that triggered this change?

forki:
thanks for testing. I will look into this case.
This issue was the selective update did not respect github dependencies and nuget packages that are retrievied via github dependencies. it also did not really handle the installsettings.

forki:
mhm that case is really interesting. I understand why it's not working, but I wonder if that really is a bad thing here. Is this something you encountered in praxis?

mrinaldi:
I think it's more common with internal packages, although I don't think it's often found.

However, this change would also break the #1025 for selective update. I say would because it was broken already by commit 49bbec7c7209e599936e11afa8004263a036f1c8. If this commit were rolled back, you'd get a lot of incorrect warnings.

forki:
Thanks for reporting. I think I have an idea how to solve it. Will try
tomorrow

mrinaldi:
Great. Let me know if you need any help.

Issue #509, opened 2015-01-03
Authorize GitHub calls


sergey-tihon:
Unauthorized API calls has very small call per hour limit:  
>Unauthenticated requests will be limited to 60 per hour ... To enjoy the default rate limit of 5,000 requests per hour, you’ll need to authenticate via Basic Auth or OAuth

It is very annoying when you has GitHub dependencies
![image](https://cloud.githubusercontent.com/assets/1197905/5603647/ca25d5b6-93a0-11e4-8a5a-89d7cfb829bc.png)

As workaround I would like to be able to create `paket.exe.config` and specify my GitHub `username` and `password`. This should allow me work without limits and delays.

@forki Your thoughts about this?

forki:
Related: https://github.com/fsprojects/Paket/issues/510

sneal:
I'd prefer to use a GitHub API key instead of my username and password.

JonCanning:
Just been looking at this, we would like to use a token to retrieve files from private repos. The possibilities seem to be:

Adding the token to the top of the dependencies file:
```source https://github.com/forki/FsUnit token```

Adding it in line:
```github forki/FsUnit FsUnit.fs token```

Or adding it to the paket.config

Any thoughts/guidance?

forki:
I guess the config would be most helpful.

baronfel:
For bonus points, once we have the concept of github tokens we could use the github releases api in the bootstrapper instead of HTML parsing, if the token was present.

forki:
I don't think that's needed. Would just introduce more stuff to the
bootstrapper

forki:
released in 2.9

PR #1111, opened 2015-10-03
#509 allow tokens to be used in credentials

JonCanning:
This introduces the concept of `<token>` as an alternative to `<credential>` in the config

`paket config add-token githubAuthKey token`

### DIFF
--- a/src/Paket.Core/ConfigFile.fs
+++ b/src/Paket.Core/ConfigFile.fs
@@ -88,14 +88,17 @@ let private readPassword (message : string) : string =
     password
 
 let getAuthFromNode (node : XmlNode) = 
-    let username = node.Attributes.["username"].Value
-    let password = node.Attributes.["password"].Value
-    let salt = node.Attributes.["salt"].Value
+    match node.Name.ToLowerInvariant() with
+    | "credential" ->
+        let username = node.Attributes.["username"].Value
+        let password = node.Attributes.["password"].Value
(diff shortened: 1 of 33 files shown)


src/Paket.Core/ConfigFile.fs
```diff
@@ -88,14 +88,17 @@ let private readPassword (message : string) : string =
...
-    let username = node.Attributes.["username"].Value
-    let password = node.Attributes.["password"].Value
-    let salt = node.Attributes.["salt"].Value
-
-    username, Decrypt salt password
-
-let private createSourceNode (credentialsNode : XmlNode) source =
-    let node = credentialsNode.OwnerDocument.CreateElement("credential")
+    match node.Name.ToLowerInvariant() with
+    | "credential" ->
```
forki: is this backwards-compatible?

forki:
checkout

==> very cool stuff!


could you please add a bit of docs abot the two modes? maybe at http://fsprojects.github.io/Paket/github-dependencies.html#Referencing-a-private-github-repository?

thx.

src/Paket.Core/ConfigFile.fs
JonCanning: It is for me but you've got me worried now. Not sure how else to test it.

forki: No I guess it is. But I don't know that part of the code very well so I was
just asking.

PR #1114, opened 2015-10-06
Change "update package" and install to not pin unchanged versions

forki:
this new version does not pin unchanged depedencies in the resolver phase.
Instead we introduce "preferred versions" and try these versions first.

/cc @mrinaldi @dnauck

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,4 +1,5 @@
-#### 2.9.1 - 07.10.2015
+#### 2.10.0-alpha001 - 07.10.2015
+* Make resolver to evaluate versions lazily 
 * BUGFIX: Wrong paket ProjectRefences name causes incorrect packaging - https://github.com/fsprojects/Paket/issues/1113
 
 #### 2.9.0 - 05.10.2015
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,10 @@
-#### 2.10.0-alpha002 - 08.10.2015
-* Make resolver to evaluate versions lazily 
+#### 2.10.0-alpha003 - 08.10.2015
+* Make resolver to evaluate versions lazily
+
+#### 2.9.2 - 08.10.2015
+* BUGFIX: Paket.Pack was broken on filesystems with forward slash seperator - https://github.com/fsprojects/Paket/issues/1119
+
+#### 2.9.1 - 07.10.2015
 * BUGFIX: Wrong paket ProjectRefences name causes incorrect packaging - https://github.com/fsprojects/Paket/issues/1113
 
 #### 2.9.0 - 05.10.2015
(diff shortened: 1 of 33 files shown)


forki:
This allows us to "update package" in much more scenarios since we allow more possible resolutions. In order to keep the lock file as stable as possible we probe the old locked versions of unrelated packages first. 

@mrinaldi I also added your new test case which didn't work before.

forki:
@dnauck I added you since I remember you reporting issues with updating transitive dependencies. This would work now since we don't pin the direct dependencies anymore

mrinaldi:
Very cool, indeed.

There's one downside though.

This is how long it takes to do a selective update now:
`4 minutes, 48 seconds - ready.`

Comparing to the 2.9.0 version:
`43 seconds - ready.`

and to the 2.5.0 version:
`15 seconds - ready.`

Seems that this lag is due to fetching all versions for every package of the whole graph.
Version 2.9.0 would not fetch versions when the package is pinned (I've a lot of pinned packages)
Version 2.5.0 would fetch versions but only for the packages from the updating package graph.

So you know, my `paket.dependencies` has 102 packages and 51 of them are pinned.

forki:
Yes I knew this would happen, and I already have plans to tackle perf.

At this point I just wanted to make sure it's "correct" and that I didn't
oversee something important.

mrinaldi:
Looks correct to me :+1:

forki:
@mrinaldi could you please test again. just be pretty fast now.

forki:
it's in 2.10.0-alpha

mrinaldi:
Much better now:
`41 seconds - ready.`

forki:
Still much slower than expected. Is there a chance I can reproduce your
sample?

mrinaldi:
The `NuGetV2.GetVersions` seems to be causing the problem.
Parallelizing the calls apparently floods the server.

Some calls are taking 2, 3x more time to complete comparing to the exactly same call when not parallelizing.
Using only `tryGetPackageVersionsViaOData` dropped the total time to `14 seconds - ready.`

I'm using ProGet 3.6.1, BTW.

forki:
Mhm maybe we need a smart way to filter unsupported api calls.

forki:
I mean I already tried but didn't really find a smart way

mrinaldi:
The problem is not exactly unsupported calls, but the `tryGetAllVersionsFromNugetODataWithFilter` which is supported but takes around 2 seconds in ProGet.

For example, `FindPackagesById()?id='Microsoft.Bcl.Build'` takes `0:00:00.252` and `Packages?$filter=Id%20eq%20'Microsoft.Bcl.Build'` takes `0:00:02.243`.

mrinaldi:
Of course, Async.Choice chooses the first call to return, but the `Packages?$filter=Id%20eq%20'Microsoft.Bcl.Build'` call definitely takes a significant amount of resources in ProGet, slowing down all other calls.

forki:
Any ideas are appreciated.

mrinaldi:
I'm working on it, will send a PR when I'm done.

PR #1118, opened 2015-10-08
Always call the fastest NuGet protocol

mrinaldi:
This PR optimizes PR #1114.

It caches the fastest protocol for each source and always call it.

Now, it's taking `15 seconds - ready.`

### DIFF
--- a/src/Paket.Core/NuGetV2.fs
+++ b/src/Paket.Core/NuGetV2.fs
@@ -610,6 +610,21 @@ let GetPackageDetails root force sources packageName (version:SemVerInfo) : Pack
       LicenseUrl = nugetObject.LicenseUrl
       DirectDependencies = nugetObject.Dependencies |> Set.ofList }
 
+let protocolCache = System.Collections.Concurrent.ConcurrentDictionary<_,_>()
+
+let getVersionsCached key f (auth, nugetURL, package) =
+    async {
+        match protocolCache.TryGetValue((auth, nugetURL)) with
+        | true, v when v <> key -> return None
+        | _ ->
+            let! result = f (auth, nugetURL, package)
+            match result with
+            | Some x ->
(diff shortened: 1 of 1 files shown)


forki:
thanks.
I need to tests this against multiple feeds

forki:
seems to work, but I had to do 530f23f - this makes sure v3 feed will always be tried if available.

forki:
if this still floods your server we can also try

                        match NuGetV3.getSearchAPI(auth,source.Url) with
                        | None -> 
                            [ getVersionsCached "Json" tryGetPackageVersionsViaJson (auth, source.Url, packageName)
                              getVersionsCached "OData" tryGetPackageVersionsViaOData (auth, source.Url, packageName)
                              getVersionsCached "ODataWithFilter" tryGetAllVersionsFromNugetODataWithFilter (auth, source.Url, packageName) ]
                        | Some v3Url -> 
                            [ tryNuGetV3 (auth, v3Url, packageName) ]

mrinaldi:
530f23f89fb444e36ea41d7f29f9a5c46056a2f4 doesn't flood. Actually, it won't even request the v3 feed (ProGet doesn't support v3), which is good.

Just out of curiosity / ignorance, why do you always need to try v3? I mean, if it's fast enough it'll be picked as default, right?

forki:
no, we pick the fastest feed for the first response and that is sometimes a different one. which results in overall slower perf

mrinaldi:
Are you assuming v3 will always be faster than v2?

If that's the case, I'd rather go with the https://github.com/fsprojects/Paket/pull/1118#issuecomment-146447039
Otherwise, you'll be firing two requests at the same time for each package, which might result in slower perf after all.

forki:
> Are you assuming v3 will always be faster than v2?

In most cases it is.

PR #1120, opened 2015-10-08
Skip unchanged groups during install

forki:
When a group has no changes at all (compared to the lock file) we can skip the resolver completely.

/cc @mrinaldi

### DIFF
--- a/src/Paket.Core/DependencyChangeDetection.fs
+++ b/src/Paket.Core/DependencyChangeDetection.fs
@@ -70,31 +70,19 @@ type RemoteFileChange =
     override this.Equals(that) = 
         match that with
         | :? RemoteFileChange as that -> 
-            this.Owner = that.Owner &&
-             this.Project = that.Project &&
-             this.Name = that.Name &&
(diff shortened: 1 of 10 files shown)


src/Paket.Core/DependencyChangeDetection.fs
```diff
@@ -60,21 +58,113 @@ let findChangesInDependenciesFile(dependenciesFile:DependenciesFile,lockFile:Loc
...
+      AuthKey : string option }
+
+    override this.Equals(that) = 
+        match that with
+        | :? RemoteFileChange as that -> 
+            this.Owner = that.Owner &&
+             this.Project = that.Project &&
+             this.Name = that.Name &&
+             this.Origin = that.Origin &&
+             ((this.Commit = that.Commit) || this.Commit = None || that.Commit = None) &&
```
forki: that's the only interesting line here. the rest of CustomEquality and CustomComparison could be removed, but how?

mrinaldi:
LGTM :+1:

PR #1125, opened 2015-10-10
Make SemVer update available in CLI

forki:
It was always possible to use paket in strict SemVer mode by using ~> operator in the dependencies file. 
Now we add three optional parameters to the command line to make SemVer compliant updates even easier. 

`--keep-major` restricts the resolver to updates that don't change major versions. 
`--keep-minor` restricts the resolver to updates that don't change major or minor versions.
`--keep-patch` restricts the resolver to updates that don't change major, minor or patch versions. 

These flags are available for install, add and update. Of course all restrictions in the dependencies file will still be met.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+#### 2.12.0-alpha001 - 10.10.2015
+* Better SemVer update by adding --keep-major, --keep-minor, --keep-patch to the CLI
+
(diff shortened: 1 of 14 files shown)


forki:
Now available for testing in latest alpha.

Stift:
I don't have a use case for that currently, but I think it could be helpful for some users.

forki:
I guess it depends a lot on how much you trust semver.

PR #1127, opened 2015-10-11
Handle framework identifiers comparison (WIP)

### LINKED ISSUE #1124: Framework version comparison have a strong opinion about silverlight
This `paket.references` :

```text
NLog framework: >= net46
```

Generate the following references :

```xml
<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Choose>
    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And $(TargetFrameworkVersion) == 'v4.6'">
      <ItemGroup>
        <Reference Include="NLog">
          <HintPath>..\packages\NLog\lib\net45\NLog.dll</HintPath>
          <Private>True</Private>
          <Paket>True</Paket>
        </Reference>
      </ItemGroup>
    </When>
    <When Condition="$(TargetFrameworkIdentifier) == 'Silverlight' And $(TargetFrameworkVersion) == 'v4.0'">
      <ItemGroup>
        <Reference Include="NLog">
          <HintPath>..\packages\NLog\lib\sl4\NLog.dll</HintPath>
          <Private>True</Private>
          <Paket>True</Paket>
        </Reference>
      </ItemGroup>
    </When>
    <When Condition="$(TargetFrameworkIdentifier) == 'Silverlight' And $(TargetFrameworkVersion) == 'v5.0'">
      <ItemGroup>
        <Reference Include="NLog">
          <HintPath>..\packages\NLog\lib\sl5\NLog.dll</HintPath>
          <Private>True</Private>
          <Paket>True</Paket>
        </Reference>
      </ItemGroup>
    </When>
  </Choose>
</Project>
```

The framework comparison seem to think that silverlight is some sort of super version of the `netXX` framework that is always sorted after other versions.

Sliverlight is dead and comming back as a zombie framework ?

vbfox:
`FrameworkIdentifier` instances were compared structurally before but it was problematic in some cases as `>= net45` matched `sl40` that wasn't in the same framework.

This PR add new operators (`.>`, `.>=`, `.<`, `.<=`) and methods (`Min`, `Max`, `IsSameFramework`) to `FrameworkIdentifier` that do the comparison with the expected semantic.

Fixes #1124 

## Questions

This PR is a work in progress because I have a few questions about the way I did the change.

### Structural Equality
The original problem come from the fact that structural comparison was used for `FrameworkIdentifier`.
But in fact no comparison (that could be used for ordering) could ever be correct for this type because `net40 > sl40` should return `false` but `net40 < sl40` should return `false` too.

Problem with that is that Sets are used over `FrameworkIndentifier` and due to their implementation require comparison... and I don't see any clean easy way out. Solutions that came to mind are :

1. Adding a immutable Set structure that only require equality and use it for `FrameworkIdentifier` instead of the built-in one
2. Use the HashSet<T> structure from the framework, it's less "F#" but doesn't require comparison (Potentially with a small helper making it's usage more "F#")
3. Keep structural comparison in addition to the semantic one, the risk is that any new code would need to be written carefully as the standard operators would be present but act in a very un-intuitive way

In the end I applied **3** as the rest of the changes were too big (`Set` is used pretty often and `FrameworkIdentifier` is part of other types that are themselves used in Sets) but I wonder if it's the correct decision.

### isTargetMatchingRestrictions

I extracted this method to be able to test it but it's behavior relative to portable profiles is pretty specific and I don't know if the name is specific enough.


forki:
would https://github.com/fsprojects/Paket/pull/1130 be enough?

PR #1134, opened 2015-10-13
Allowed for non semver InformationalVersion.

### LINKED ISSUE #1133: AssemblyInformationalVersion with non integer value causes opaque error message.
I have an assembly with `1.2.foo.4` as the `AssemblyInformationalVersion`. 

`$ paket.exe pack output bin` fails with 

```
Input string was not in a correct format.
```

Not very obvious what the problem is without attaching a debugger.

xavierzwirtz:
Silently ignores when not a semver compliant version.
Fixes #1133


forki:
thx

PR #1137, opened 2015-10-14
Fixes combine restrictions

mrinaldi:
This PR correctly combines AtLeast with Between and Exactly with Between restrictions.

PR #1138, opened 2015-10-15
Handle silverlight framework identifiers comparison

vbfox:
Continue on & rebase #1130 (That was itself based on #1127)

I didn't keep the history because the back & forth was just making the pull request unreadable.

`FrameworkRestriction.IsSameCategoryAs` was introduced in addition to the version on `FrameworkIdentifier` to avoid repeating the checks over and over.


forki:
so is this already ready?

vbfox:
Yes I think it is, I ran `paket install` on our work solution and between this change and #1137 it really clean the references & Conditions in the projects.

Actually #1137 even found a bug in our restrictions where a `>= net451` package was used in a `net40` project :+1:

forki:
very cool

PR #1140, opened 2015-10-15
Add monoandroid403 to be matched as Some MonoAndroid

dmunch:
Some Xamarin Nugets (e.g.  Xamarin.Android.Support.v4) use platform 'monoandroid403' which isn't recognized by Paket yet. The suggested fix is pretty basic by just adding it to the list of recognized platforms.


forki:
thx

PR #1144, opened 2015-10-16
Fixes lock file restrictions parsing

mrinaldi:
This PR fixes the parsing of restrictions of transitive dependencies in lock file.

Using the `paket.dependencies` below
```

nuget Microsoft.Owin.Host.SystemWeb ~> 2.1
```

It generates a different `paket.lock` when I run `paket install` twice.
The same happens if I run a `paket update` followed by a `paket install`.

### DIFF
--- a/src/Paket.Core/LockFile.fs
+++ b/src/Paket.Core/LockFile.fs
@@ -208,6 +208,16 @@ module LockFileParser =
     let Parse(lockFileLines) =
         let remove textToRemove (source:string) = source.Replace(textToRemove, "")
         let removeBrackets = remove "(" >> remove ")"
+        let parsePackage (s : string) =
+            let parts = s.Split([|" - "|],StringSplitOptions.None)
+            let optionsString = 
+                if parts.Length < 2 then "" else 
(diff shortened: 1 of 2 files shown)


src/Paket.Core/LockFile.fs
```diff
@@ -208,6 +208,16 @@ module LockFileParser =
     let Parse(lockFileLines) =
         let remove textToRemove (source:string) = source.Replace(textToRemove, "")
         let removeBrackets = remove "(" >> remove ")"
+        let parsePackage (s : string) =
+            let parts = s.Split([|" - "|],StringSplitOptions.None)
+            let optionsString = 
+                if parts.Length < 2 then "" else 
+                if parts.[1] <> "" && parts.[1].Contains(":") |> not then
+                    ("framework: " + parts.[1]) // TODO: This is for backwards-compat and should be removed later
```
mrinaldi: Should this be removed by now?

forki: Nope. Please remove the comment. I guess we better keep it forever

PR #1148, opened 2015-10-18
Removing unnecessary comment

mrinaldi:
As stated [here](https://github.com/fsprojects/Paket/pull/1144#discussion_r42306602).


forki:
Yeah I guess we should support it until everyone forgets about it. ;-)

PR #1153, opened 2015-10-19
try to cache proxy

forki:
for a couple of users it seems to be very very slow to calc the proxy. see https://gitter.im/fsprojects/Paket discussion with @alexeyzimarev

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,9 +1,10 @@
-#### 2.16.0-alpha003 - 20.10.2015
+#### 2.16.0-alpha004 - 21.10.2015
 * Log Request performance
 * Check that download http status code was 200
 * Try to report better error when file is blocked by Firewall - https://github.com/fsprojects/Paket/pull/1155
 * BUGFIX: Fixed loading of Project files on mono - https://github.com/fsprojects/Paket/pull/1149
 * PERFORMANCE: Caching proxy scheme - https://github.com/fsprojects/Paket/pull/1153
+* USABILITY: If caching fails we should recover - https://github.com/fsprojects/Paket/issues/1152
 
 #### 2.15.1 - 17.10.2015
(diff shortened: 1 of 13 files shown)


forki:
//cc @vasily-kirichenko could you please take a look?

vasily-kirichenko:
Looks like an ordinary thread safe memoization. Maybe it is worth to extract `memoze` into some core module.

forki:
the question is more: do you think it's enough to cache proxy per HTTP host?

vasily-kirichenko:
No idea.

forki:
fair enough ;-)

forki:
I released this as 2.16.0-alpha001.
ypu can get it with `paket.bootstrapper.exe prerelease`

@alexeyzimarev, @alistair please check if it improves things

@vasily-kirichenko  could you please verify that we don't break you

thanks everbody.

forki:
@dnauck you are also behind proxy, right?

alexeyzimarev:
I am at home now and have no proxy. Will be able to check tomorrow if I would reach my machine at work, I will be at another office for three days.

alistair:
So I have just updated to the latest,  cleaned cache and packages directory and run the following

https://gist.github.com/alistair/65cba5d92556e6f0457c

Time taken has dropped from 7 minutes to 1.

dnauck:
@forki no .. not with my dev machine, cannot test

alexeyzimarev:
I tested with 2.16.0 and it works splendidly :+1: 46 seconds instead of seven minutes.

forki:
Awesome. Thanks for all the help here.

vasily-kirichenko:
Tested `paket outdated` on 2.16.0. Works OK, but I'm not sure it's faster.

forki:
It's only faster for a handful of people where the proxy detection was
super slow. Most people (like myself) will not have perf gains. But I was
worried that caching might break things.

Issue #1107, opened 2015-10-03
Convert-from-nuget should simplify framework restrictions if possible


forki:
consider the following dependencies file:

    source https://www.nuget.org/api/v2/
    
    nuget angularjs 1.4.3 framework: >= net45
    nuget AngularTemplates.Compile 1.0.0 framework: >= net45
    nuget Antlr 3.4.1.9004 framework: >= net45
    nuget Autofac 3.5.0 framework: >= net45
    nuget Autofac.Owin 3.1.0 framework: >= net45
    nuget Autofac.WebApi 3.1.0 framework: >= net45
    nuget Autofac.WebApi2 3.4.0 framework: >= net45
    nuget Autofac.WebApi2.Owin 3.2.0 framework: >= net45
    ...
    93 lines, 15 of which is private packages on local repo

It would be nice if we generate a global `framework: >= net45` if everything has the same framework restrictions. Like:

    source https://www.nuget.org/api/v2/
    framework >= net45
    
    nuget angularjs 1.4.3
    nuget AngularTemplates.Compile 1.0.0
    nuget Antlr 3.4.1.9004
    nuget Autofac 3.5.0
    nuget Autofac.Owin 3.1.0
    nuget Autofac.WebApi 3.1.0
    nuget Autofac.WebApi2 3.4.0
    nuget Autofac.WebApi2.Owin 3.2.0
    ...
    93 lines, 15 of which is private packages on local repo

PR #1159, opened 2015-10-22
Convert-from-nuget should simplify framework restrictions if possible

forki:
implements #1107 

/cc @theimowski could you please take a short look? Thx

### DIFF
--- a/src/Paket.Core/DependenciesFile.fs
+++ b/src/Paket.Core/DependenciesFile.fs
@@ -407,7 +407,7 @@ type DependenciesFile(fileName,groups:Map<GroupName,DependenciesGroup>, textRepr
                 let sameRequirements =
                     rest |> Seq.forall (fun p' -> p'.Settings.FrameworkRestrictions = package.Settings.FrameworkRestrictions)
 
-                if not sameRequirements then dependenciesFile else
+                if not sameRequirements || package.Settings.FrameworkRestrictions = [] then dependenciesFile else
                 
                 let newDependenciesFile = dependenciesFile.AddFrameworkRestriction(group.Name,package.Settings.FrameworkRestrictions)
                 group.Packages
@@ -471,7 +471,7 @@ type DependenciesFile(fileName,groups:Map<GroupName,DependenciesGroup>, textRepr
         |> Map.map resolveGroup 
 
 
-    member __.AddFrameworkRestriction(groupName, frameworkRestrictions:FrameworkRestrictions) =
+    member private __.AddFrameworkRestriction(groupName, frameworkRestrictions:FrameworkRestrictions) =
         let restrictionString = sprintf "framework %s" (String.Join(", ",frameworkRestrictions))
 
         let list = new System.Collections.Generic.List<_>()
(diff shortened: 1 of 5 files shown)


src/Paket.Core/NugetConvert.fs
```diff
@@ -296,7 +296,9 @@ let createDependenciesFileR (rootDirectory : DirectoryInfo) nugetEnv mode =
 
             Paket.DependenciesFile(DependenciesFileParser.parseDependenciesFile dependenciesFileName newLines))
 
-    if File.Exists dependenciesFileName then read() else create()
+    match (if File.Exists dependenciesFileName then read() else create()) with
```
theimowski: use `Trial.lif` (map) ?

src/Paket.Core/Simplifier.fs
```diff
@@ -66,7 +66,7 @@ let simplify interactive environment = trial {
     let! lockFile = environment |> PaketEnv.ensureLockFileExists
 
     let flatLookup = lockFile.GetDependencyLookupTable()
-    let dependenciesFileRef = ref environment.DependenciesFile
+    let dependenciesFileRef = ref (environment.DependenciesFile.SimplifyFrameworkRestrictions())
```
theimowski: this will happen also on `paket simplify` command?

forki: on simplify and convert

src/Paket.Core/DependenciesFile.fs
```diff
@@ -398,6 +398,27 @@ type DependenciesFile(fileName,groups:Map<GroupName,DependenciesGroup>, textRepr
 
     member __.Groups = groups
 
+    member this.SimplifyFrameworkRestrictions() = 
+        let transform (dependenciesFile:DependenciesFile) (group:DependenciesGroup) =
+            if group.Options.Settings.FrameworkRestrictions <> [] then dependenciesFile else
```
theimowski: what happens when there are global restrictions and standalone restrictions per package? The latter override global?

src/Paket.Core/DependenciesFile.fs
```diff
@@ -398,6 +398,27 @@ type DependenciesFile(fileName,groups:Map<GroupName,DependenciesGroup>, textRepr
...
+            match group.Packages with
+            | [] -> dependenciesFile
+            | package::rest ->
+                let sameRequirements =
+                    rest |> Seq.forall (fun p' -> p'.Settings.FrameworkRestrictions = package.Settings.FrameworkRestrictions)
+
+                if not sameRequirements || package.Settings.FrameworkRestrictions = [] then dependenciesFile else
```
theimowski: must check if restrictions are empty? does it harm if you add empty restrictions globally?

src/Paket.Core/DependenciesFile.fs
forki: we don't do anything in this case.
during normal run both restrictions work together as two filters.

src/Paket.Core/DependenciesFile.fs
forki: yes we create a stupid line

theimowski:
LGTM, wondering if `SimplifyFrameworkRestrictions` could be made a bit more idiomatic, but that's probably just some cosmetics

forki:
What would you call it?

theimowski:
don't mean name itself, but rather the logic - maybe replace ifs with pattern matching where possible - but as I said it's just cosmetics, not crucial

theimowski:
:+1:

forki:
we now allow to simplify common framework restrictions. see https://github.com/fsprojects/Paket/blob/master/tests/Paket.Tests/Simplifier/BasicScenarioSpecs.fs#L342

theimowski:
neat!

PR #1160, opened 2015-10-22
Global resolver strategy option

mrinaldi:
Currently, Paket uses the Max strategy and you can override it per package using the [`!` modifier](http://fsprojects.github.io/Paket/nuget-dependencies.html#Paket-s-NuGet-style-dependency-resolution-for-transitive-dependencies)

This PR adds a new setting to specify the default resolver strategy for the group.

Now, you can set the default strategy either by doing `strategy min` or `strategy max` in `paket.dependencies`. If no strategy is set, it'll continue using the Max by default.

In addition, a new strategy modifier has been introduced, similar to the [`!` modifier](http://fsprojects.github.io/Paket/nuget-dependencies.html#Paket-s-NuGet-style-dependency-resolution-for-transitive-dependencies), the new modifier is the `@`, like in the example below:

```
strategy min
source https://nuget.org/api/v2

nuget FAKE @~> 4.0
```

This makes all transitive packages of FAKE resolve using the Max strategy, albeit the global strategy is Min.

Although improbable, to resolve conflicts between strategy modifiers (`!` and `@`), it'll compare depth of the requirements in conflict and if both are equal, it'll select the default strategy. In case no default strategy is set, it'll use Max by default.

In case this PR is merged, I'll begin working on the docs.

PR #1163, opened 2015-10-23
Queries every NuGet feed in parallel and combines the results

forki:
relates to #1160

tumblr_inline_nv03wtrh881raprkq_500


mrinaldi:
Looks good :+1:

PR #1172, opened 2015-10-28
Use Async.choice to query all feeds in parallel for package details

forki:


mrinaldi:
Nice improvement!
Looks good :+1:

PR #1175, opened 2015-10-28
Remove Seq.toArray before Async.Parallel

yever:
No need to convert a sequence to an array before Async.Parallel. Order is preserved anyway.


forki:
thx

Issue #55, opened 2014-09-04
Nuget-style pessimistic version constraint isn't honored with other nugets when another nuget has the same indirect dependency


agross:
Direct dependency of N.B.Windsor is specified explicitly:

```
source "http://nuget.org/api/v2"

nuget "Castle.Windsor" "!>= 0"
nuget "Nancy.Bootstrappers.Windsor" "!~> 0.23"
```

```
NUGET
  remote: http://nuget.org/api/v2
  specs:
    Castle.Core (3.3.0)
    Castle.Windsor (3.3.0)  // should be 3.2.1
      Castle.Core (>= 3.3.0)
    Nancy (0.23.2)
    Nancy.Bootstrappers.Windsor (0.23.2)
      Nancy (>= 0.23.2)
      Castle.Windsor (>= 3.2.1)
```
## 

Direct dependency of N.B.Windsor is specified indirectly through C.W-log4net:

```
source "http://nuget.org/api/v2"

nuget "Castle.Windsor-log4net" "!>= 0"
nuget "Nancy.Bootstrappers.Windsor" "!~> 0.23"
```

```
NUGET
  remote: http://nuget.org/api/v2
  specs:
    Castle.Core (3.3.0)
    Castle.Core-log4net (3.3.0)
      Castle.Core (>= 3.3.0)
      log4net (1.2.10)
    Castle.LoggingFacility (3.3.0)
      Castle.Core (>= 3.3.0)
      Castle.Windsor (>= 3.3.0)
    Castle.Windsor (3.3.0)  // should be 3.2.1
      Castle.Core (>= 3.3.0)
    Castle.Windsor-log4net (3.3.0)
      Castle.Core-log4net (>= 3.3.0)
      Castle.LoggingFacility (>= 3.3.0)
    Nancy (0.23.2)
    Nancy.Bootstrappers.Windsor (0.23.2)
      Nancy (>= 0.23.2)
      Castle.Windsor (>= 3.2.1)
    log4net (1.2.10)
```

PR #1177, opened 2015-10-29
Favors strategy overrides for top-level dependencies

mrinaldi:
Fixes regression for #55.

### DIFF
--- a/src/Paket.Core/PackageResolver.fs
+++ b/src/Paket.Core/PackageResolver.fs
@@ -316,9 +316,9 @@ let Resolve(groupName:GroupName, sources, getVersionsF, getPackageDetailsF, stra
                 | Install
                 | UpdateAll
                 | UpdateGroup _ ->
-                    match currentRequirements |> Set.exists (fun r -> r.Parent.IsRootRequirement()) with
-                    | true -> ResolverStrategy.Max
-                    | false -> combined
+                    match currentRequirement.Parent.IsRootRequirement(), Set.count currentRequirements with
+                    | true, 1 -> ResolverStrategy.Max
+                    | _ -> combined
                 | UpdatePackage (g, p) ->
                     match groupName = g && currentRequirement.Name = p with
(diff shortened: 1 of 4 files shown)


forki:
thanks. it also fixes https://github.com/fsprojects/Paket/commit/0e4557086fb4c5f3dda7761efa1037de24d459cf

mrinaldi:
For a `paket.dependencies` like this:
```
strategy min
source http://nuget.org/api/v2

nuget Nancy.Bootstrappers.Windsor ~> 0.23
nuget Castle.Windsor
```

What do you think the version of `Castle.Windsor` should be when we update all?
`3.2.1` (favoring the strategy option) or `3.3.0` (lastest one)

I mean, when we are updating all and resolving top-level dependencies that are also transitive, should we only favor [strategy overrides](http://fsprojects.github.io/Paket/nuget-dependencies.html#Strategy-modifiers) (`!` and `@`), or should we favor the [strategy option](http://fsprojects.github.io/Paket/dependencies-file.html#Strategy-option) also?

forki:
3.2.1 is the only version that works with Nancy.Bootstrappers.Windsor 0.23

mrinaldi:
Sorry, may be that was a bad example.

How about this one:
```
strategy min
source http://nuget.org/api/v2

nuget Castle.Core
nuget Castle.Windsor
```

Note the `strategy min`. What do you think the version of Castle.Core should be when we update all?
`3.2.1` (favoring the strategy option) or `3.3.0` (lastest one)

forki:
yeah that is really tricky. I guess we can take any Castle.Windsor (probably the latest) and need to find the lowest matching Castle.Core

mrinaldi:
Alright. This PR is fine then.

forki:
cool. https://github.com/fsprojects/Paket/commit/7f10a2fc3964786937a48924c5bc7b767d12787b

agross:
Cool! :+1:

PR #1178, opened 2015-10-30
Allow filtered updates of packages matching a regex

mavnn:
Especially useful for situations like "update all packages that start with my organisations prefix"


forki:
do you have a small sample for this? Then I could add a regression test

mavnn:
Of course.

Given dependency file:

```

nuget Castle.Windsor                           
nuget NUnit                             
nuget Microsoft.AspNet.WebApi.SelfHost
```
Then:

``paket update nuget Microsoft.* --filter`` should update AspNet

``paket update nuget [MN].* --filter`` will update NUnit and AspNet but leave Castle.Windsor untouched.

Without the ``--filter`` flag, package names are treated as exact matches.

src/Paket.Core/PackageResolver.fs
```diff
@@ -189,6 +189,7 @@ type Resolved = {
 type UpdateMode =
     | UpdatePackage of GroupName * PackageName
```
forki: can we remove that case and express it as `UpdateFiltered of GroupName * PackageFilter`?

mavnn: As in, escape ``.``s and treat it as a filter? I think it would work. Hope your regression tests are good up to date though! :D

mavnn: Give me a few minutes and I'll see if all the tests pass.

forki:
image

mavnn:
My code obviously rules :)

mavnn:
Mind pushing that test to my master branch? I'll add it to the pull request.

mavnn:
Ah, ok - I'll grab it.

forki:
https://github.com/fsprojects/Paket/commit/40ecd395cb512ea62cf9647d11d66bd0b431ff90

forki:
you need to run `build.cmd runintegrationtests` since we don't execute that target during standard build

mavnn:
Should having working unit and integration tests now

tests/Paket.Tests/UpdateProcessSpecs.fs
```diff
@@ -348,27 +348,11 @@ let ``SelectiveUpdate conflicts when a dependency is contrained``() =
 
     (fun () ->
     selectiveUpdate true noSha1 (VersionsFromGraph graph) (PackageDetailsFromGraph graph) lockFile dependenciesFile
-        (PackageResolver.UpdateMode.UpdatePackage(Constants.MainDependencyGroup, PackageName "Castle.Core-log4net")) SemVerUpdateMode.NoRestriction
+        (PackageResolver.UpdateMode.UpdateFiltered(Constants.MainDependencyGroup, PackageFilter "Castle\\.Core-log4net")) SemVerUpdateMode.NoRestriction
     |> ignore)
     |> shouldFail
 
 [<Test>]
-let ``SelectiveUpdate does not update any package when package does not exist``() =
```
mavnn: I removed this test as it's expectation didn't seem meaningful anymore. With pattern based updates, a pattern that didn't match any packages should be a no-op, not an exception.

The update package command still checks that the specific named package exists before calling UpdateFiltered

tests/Paket.Tests/UpdateProcessSpecs.fs
```diff
@@ -736,11 +720,11 @@ let ``SelectiveUpdate updates package that conflicts with transitive dependency
     nuget Ninject.Extensions.Interception
     nuget Ninject.Extensions.Logging""")
     
-    let packageName = PackageName "Ninject.Extensions.Logging"
+    let packageFilter = PackageFilter "Ninject\\.Extensions\\.Logging"
```
forki: it would be cool if we would have a function which transform PackageNames into PackageFilter

mavnn: Static members ``PackageName.toFilter`` and ``PackageFilter.ofName``?

forki: yes something like that. then we could keep the tests more stable

forki:
Could also please write one or two samples in the docs of the update command?

mavnn:
@forki crap. Push to wrong remote by mistake. Sorry.

mavnn:
/me goes off to change default push repository

forki:
no worries ;-)

PR #1179, opened 2015-10-30
search for paket.references in startup directory (auto-restore feature)

punker76:
search for paket.references in startup diretory if there is no such file in the project folder.
this makes it possible to use the auto-restore feature too (e.g. building with visual studio)
without that, the packages are not restored.


forki:
cool thx

PR #1180, opened 2015-10-30
Fix compiling Paket.Bootstrapper.Tests with Visual Studio

punker76:
import paket.targets fixes visual studio compiling.
a clean visual studio build will be failed without the paket.targets


forki:
thx

PR #1194, opened 2015-11-04
Improve binding redirects performance

mrinaldi:
After PR #1187, `paket install` takes `24 seconds - ready.`
After this PR, `paket install` takes `8 seconds - ready.`


forki:
thx

PR #1196, opened 2015-11-05
Generates and parses strategy option in lock file

mrinaldi:
This PR generates and parses strategy option in lock file. The lack of it is causing the install process to detect changes when there is none.

Running `paket install` several times against this `paket.dependencies`:
```
strategy: min

nuget Paket.Core
```

would not skip the resolver. Here is the output:
```
Paket version 2.22.0.0
 - Paket.Core 2.22.0
 - FSharp.Core 3.0.2
 - Newtonsoft.Json 3.5.8
1 second - ready.
```

This is the output of this PR:
```
Paket version 2.22.0.0
Skipping resolver for group Main since it is already up-to-date
0 seconds - ready.
```

### DIFF
--- a/src/Paket.Core/LockFile.fs
+++ b/src/Paket.Core/LockFile.fs
@@ -176,6 +176,7 @@ module LockFileParser =
     | CopyLocal of bool
     | Redirects of bool
     | ReferenceCondition of string
+    | ResolverStrategy of ResolverStrategy option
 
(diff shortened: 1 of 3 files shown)


forki:
very good.

in the default case (Strategy max) we should not touch the lock file. is this the case?

mrinaldi:
The default case is actually no strategy at all, which internally will be treated as max.
Indeed, for that case, we don't touch the lock file.

If the strategy max is specify though, it'll touch the lock file, and I believe it should, since it's explicit specified.

forki:
very good!

PR #1202, opened 2015-11-08
Fixes binding redirect integration test

mrinaldi:
AutoFixture 3.36.10 was released with no signature, hence no binding redirect is generated for it.
This PR fixes AutoFixture in a version that is known to be signed.


forki:
Yeah I guessed something like that was happening. Thx

Issue #1201, opened 2015-11-08
open up the TemplateFile API in Paket.Core


krgn:
I would be great to be able to re-use the TemplateFile parser code in a project I'm building (https://github.com/krgn/Paket2Nix) but its currently internal/private.

Would you consider making the API public for use in cases where external tools (in this case my tool which reads `.lock` files to create `nix` derivations from) need access to this metadata without necessarily having access to the assemblies themselves? 

Or, which other process would be recommended in this case? 

Reading the assemblies themselves would mean the project would have to be built when my tool is invoked on a project, which might not always be the case.

Thanks!

Karsten

forki:
Sure. Please add methods to PublicAPI.fs - we try to keep that interface
stable.

PR #1203, opened 2015-11-08
expose TemplateFile types and Dependencies member

krgn:
This PR corresponds to an issue I opened #1201 

I removed `internal` from all involved types in the TemplateFile API and added a member in `Dependencies` to enumerate them. Since most of the code is already tested as far as I can see, I didn't add a test case. Ping me if I should anyways ;)


krgn:
Awesome! Thanks a lot

PR #1205, opened 2015-11-09
Support for only using nuget v3 api

xavierzwirtz:
Can be used by setting `source https://api.nuget.org/v3/index.json` in paket.dependencies

Adds a reference to NuGet.Versioning so that we can reuse there parsing logic for the new version range syntax.

I wrote this so that paket could be used against nuget servers that do not support the v2 api. The v3 api, while still crazy in places, is much easier to implement.


paket.dependencies
```diff
@@ -3,6 +3,7 @@ source https://nuget.org/api/v2
 nuget Newtonsoft.Json redirects: on
 nuget Argu
 nuget FSharp.Core redirects: on
+nuget NuGet.Versioning
```
forki: why do we need this?

forki: ah "Adds a reference to NuGet.Versioning so that we can reuse there parsing logic for the new version range syntax."

I don't think that's a good idea. we already have parse logic and I think we should take that dependency. We're much faster with fixes if this little parse code is under our control.

paket.lock
```diff
@@ -4,6 +4,68 @@ NUGET
     Argu (1.1.2)
     FSharp.Core (4.0.0.1) - redirects: on
     Newtonsoft.Json (7.0.1) - redirects: on
+    NuGet.Versioning (3.2.0)
+      System.Diagnostics.Debug (>= 4.0.10-beta-23109)
```
forki: Sorry I don't think it's a good idea to take this dependency ;-)

xavierzwirtz:
I removed the dependency on `NuGet.Versioning`, it looked like the dependency range syntax had changed radically in v3, but it hasn't changed at all. The v3 api is just handing out valid, but odd, version ranges.

xavierzwirtz:
One thing I didn't do, which could be a target for a future pr, is call the v3 apis whenever a v2 api is handed in. That might speed up the typical dependency resolution.

xavierzwirtz:
Wooooo!

forki:
hurray

PR #1209, opened 2015-11-10
make ListTemplateFiles more robust

krgn:
A TemplateFile may not be loadable if some of the referenced files are not available (e.g. the Paket.PowerShell project and its dependencies on Linux) and thus renders the method useless in these scenarios.

I refined the loading process to skip TemplateFiles that crash on Load.


forki:
thx

PR #1212, opened 2015-11-12
WIP: Try to activate V3 API

forki:
- [x] - create integration test
- [x] - reactivate API
- [ ] - fix integration test

/cc @xavierzwirtz

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+#### 2.24.5 - 12.11.2015
+* Propagate framework restriction correctly - https://github.com/fsprojects/Paket/issues/1213
+
 #### 2.24.4 - 12.11.2015
 * Revert v3 stuff in order to work around https://github.com/fsprojects/Paket/issues/1211 
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,3 +1,6 @@
+#### 2.24.6 - 12.11.2015
+* Propagate framework restriction with top-level dependencies correctly - https://github.com/fsprojects/Paket/issues/1215
+
 #### 2.24.5 - 12.11.2015
 * Propagate framework restriction correctly - https://github.com/fsprojects/Paket/issues/1213
 
--- a/integrationtests/Paket.IntegrationTests/FrameworkRestrictionsSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/FrameworkRestrictionsSpecs.fs
@@ -1,4 +1,4 @@
-﻿module Paket.IntegrationTests.FrameworkRestrictionsSpecs
+module Paket.IntegrationTests.FrameworkRestrictionsSpecs
 
 open Fake
 open Paket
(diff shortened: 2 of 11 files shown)


forki:
@xavierzwirtz do you see the issue?

xavierzwirtz:
I have most of the issue resolved, working on some minor differences between the lock file generated against a V2 api and a lock file generated against a V3 api.

xavierzwirtz:
I have submitted a pr (#1216) that fixes most of this issue.

I updated your regression test to compare the paket.lock when run against the `v2` api and the `v3` api. [There is only one difference now,](https://www.diffchecker.com/i0lfaxjv), and I believe the bug is actually in the `v2` api. 

From what I can tell `FSharp.Data` depends on `Zlib.Portable (1.11.0)`, but when using the `v2` api it doesnt register a dependency for `Zlib.Portable (1.11.0)`.

forki:
so you found a bug in our v2 implementation?

xavierzwirtz:
I think so, I'd like you to take a look at it.

forki:
image

looks like v2 is correct. ZLib.Portable is only used for portable projects.

xavierzwirtz:
I was not calling the right method to parse framework dependencies. Fixed that. Also added caching in.

The integration test is passing now.

forki:
Awesome. 
Will look into this tomorrow. Thanks buddy

xavierzwirtz:
No problem, glad this was caught quickly.

xavierzwirtz:
Submited pr #1222 

I rewrote the `v2` and `v3` getDetails to use a single cache function.

Also changed the behavior of the cache. When an error occurs it doesnt write an error file, it simply falls back to the api. Logs to the console with `traceWarnfn`.

forki:
the error file was a idea that future runs of paket don't have to retry the same API call

xavierzwirtz:
Requires the user to call paket with force to blow them out though, doesn't it? Feels rather irritating from a user perspective, but I can understand if that's out of scope for this pr.

forki:
yes, but the idea was: if all calls failed (e.g. because of error files) we
will also throw error files away and retry.
So usaully nobody should call with -f

2015-11-13 18:17 GMT+01:00 Xavier Zwirtz <notifications@github.com>:

xavierzwirtz:
Ah, okay. I think I missed that in the v3 implementation of caching.

xavierzwirtz:
I tested the old implementation of the v2 api's caching, if an error file exists you have to run paket with -f.

forki:
maybe things got broken. ;-)

2015-11-13 18:25 GMT+01:00 Xavier Zwirtz <notifications@github.com>:

xavierzwirtz:
I think so. Would you like me to try to re-implement the behavior you describe?

forki:
if you think it's useful.
I don't really know if it's still needed, since we also change to
Async.Parallel and Async.Choice combo

2015-11-13 18:27 GMT+01:00 Xavier Zwirtz <notifications@github.com>:

xavierzwirtz:
I think the way it works right now, without error files, is simpler and not any slower. Your probably right, the async work makes it not needed.

forki:
Ok. Then please drop the error files. I never really liked them anyways.
Actually I'm happy to get rid of these.

xavierzwirtz:
Heh, awesome. Thats done in the current state of the pr.
On 11/13/2015 11:35:10 AM, Steffen Forkmann <notifications@github.com> wrote:
Ok. Then please drop the error files. I never really liked them anyways.
Actually I'm happy to get rid of these.

Issue #1221, opened 2015-11-13
Combining type project and specifying files included in nuget package with different framework requirement


theggelund:
When using project type _project_ in paket.template and specifying different files for different framework, the packages for 4.5.2 and 4.6 are correct but the package for 4.0 is replaced with the content of package for 4.6

paket.template

    type project
    id DIPS.Infrastructure
    authors DIPS ASA
    description Infrastructure libraries
    files
	    ..\..\..\..\bin\v4.6\AnyCPU\release\DIPS.Infrastructure.dll ==> lib/net46
	    ..\..\..\..\bin\v4.6\AnyCPU\release\DIPS.Infrastructure.pdb ==> lib/net46
	    ..\..\..\..\bin\v4.5.2\AnyCPU\release\DIPS.Infrastructure.dll ==> lib/net452
	    ..\..\..\..\bin\v4.5.2\AnyCPU\release\DIPS.Infrastructure.pdb ==> lib/net452
	    ..\..\..\..\bin\v4.0\AnyCPU\release\DIPS.Infrastructure.dll ==> lib/net40
	    ..\..\..\..\bin\v4.0\AnyCPU\release\DIPS.Infrastructure.pdb ==> lib/net40

forki:
do you have this as a zip or git repo for me?

theggelund:
I can create a git repo, give me a couple of hours

theggelund:
Updated https://github.com/theggelund/paket_bug

I've created 2 project files in separate folder. They have the same dependencies but that doesn't matter.
They have each a paket.template file and a reference from DIPS.Infrastructure to DIPS.Infrastructure.Services. 

I believe that the error occurs when one project references the other.

the bin folder contains separate files for v4.0, v4.5.2 and v4.6. These dll and pdb files are only text files with the version expected inside.

the out folder contains the nupkg files that has error, DIPS.Infrastructure has the error in v4.0\AnyCPU\release\DIPS.Infrastructure.dll

forki:
I tried 
   paket install
   paket pack output D:\code\paket_bug\out

on your solution but got: 

![image](https://cloud.githubusercontent.com/assets/57396/11155198/6cef6772-8a44-11e5-82c4-63396f137811.png)

something is fucky here

forki:
it can't load the assemblies to retrieve meta data. can you try to reducethe repro  to "type file"?

theggelund:
[...]

PR #1227, opened 2015-11-16
Update ProjectFile.GetTargetProfile to work with conditional nodes

amazingant:
Working on #1221 (cc/ @theggelund and @forki), the contents of the test file I changed in 73e6f8c686ca5a26fd9a7decee82435cad929daf appear to work as of b905c750d736a328931ac436359e8eddf18fa2a1.

The output looks correct to me, but I'll need one of you to confirm, as I'm not sure what the output from the paket_bug project is supposed to be. Just testing from an fsx file like so:

```FSharp
#r @"bin\Paket.Core.dll"
let proj = Paket.ProjectFile.LoadFromFile @"tests\Paket.Tests\ProjectFile\TestData\ProjectWithConditions.fsprojtest"
let profile = proj.GetTargetProfile()
```

`profile` outputs as `SinglePlatform (DotNetFramework V4_6)`. If I change the `TargetFrameworkProfile` node's condition to check for framework `v4.6` instead, then I get `SinglePlatform (DotNetFramework V4_Client)`. Doesn't make sense since there isn't a client profile for 4.6 (that I know of?), but shows that the condition on the node is being evaluated.

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -1,4 +1,4 @@
-﻿namespace Paket
+namespace Paket
 
 open Paket.Domain
 open Paket.Logging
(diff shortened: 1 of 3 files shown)


forki:
this looks good to me. In debugger I see that the files are copiec twice - but the correct files. So it works as expected!

yay

Issue #1225, opened 2015-11-16
Issue with paket convert-from-nuget


theggelund:
I tried to convert a project from nuget to paket.

![image](https://cloud.githubusercontent.com/assets/648714/11177050/e7db662a-8c42-11e5-832f-c74718b34533.png)
 
but paket stops, looks like its waiting for something that doesn't happen.

Paket has put a json file in nuget cache folder with the following content
```json
{"Dependencies":[],"PackageName":"odp.net.managed","SourceUrl":"http://dips-nuget/nuget/TeamFrost","Unlisted":false,"DownloadUrl":"http://dips-nuget/api/v2/package/TeamFrost/odp.net.managed/121.1.2","LicenseUrl":"http://www.oracle.com/technetwork/licenses/ea-license-noexhibits-1938914.html","CacheVersion":"2.0"}
```

paket.dependencies
```
source \\p-fs01\exetest\repository
source http://dips-nuget/nuget/TeamFrost

nuget Nuget.CommandLine @>= 0
nuget Antlr4.StringTemplate @= 4.0.6.9004
nuget Common.Logging @= 3.0.0 framework: >= net451
nuget Common.Logging.Core @= 3.1.0
nuget CommonServiceLocator @= 1.0 framework: >= net451
nuget Dapper @= 1.40
nuget DIPS.Lab.IfsServer.Plugin @= 0.0.0.9 framework: >= net40
nuget DIPS.Lab.IfsServer.Plugin.Lab.Persistence @= 0.0.0.9 framework: >= net40
nuget DIPS.Lab.Infrastructure @= 0.0.0.59
nuget DIPS.Lab.Infrastructure.Connector.MSMQ @= 0.0.0.59
nuget DIPS.Lab.Infrastructure.Database @= 0.0.0.57 framework: >= net451
nuget DIPS.Lab.Infrastructure.Logging.Serilog @= 0.0.0.56 framework: >= net451
nuget DIPS.Lab.Infrastructure.Messaging @= 0.0.0.59
nuget DIPS.Testing.NUnit @= 1.3.0.0
nuget Extended.Wpf.Toolkit @= 2.3.0 framework: >= net451
nuget FluentAssertions @= 3.3.0
nuget ICSharpCode.SharpZipLib.dll @= 0.85.4.369 framework: >= net451
nuget Jil @= 2.9.0
nuget MiniProfiler @= 3.1.1.140 framework: >= net451
nuget Moq @= 4.2.1502.0911
[...]

PR #1228, opened 2015-11-16
Sets default resolver strategy for convert from nuget to None

mrinaldi:
See https://github.com/fsprojects/Paket/issues/1225#issuecomment-156954518


forki:
awesome

Awesome work buddy. As ayways.

PR #1231, opened 2015-11-17
Fixed issue with V3 feeds doing api requests even when the paket.lock is fully specified.

xavierzwirtz:
The way I wrote the `NugetV3Source` caused a http request during parsing of the `paket.lock`. This fixes that.


forki:
thx

PR #1236, opened 2015-11-18
Compare SemVers instead of strings

inosik:
This allows updates from e.g. 2.26.0-alpha002 to 2.26.0. Couldn't update today with the bootstrapper because of this.


forki:
thanks.

Issue #1224, opened 2015-11-16
Privacy Leakage: feed password may be revealed in output


cdrnet:
When paket fails it may leak the user's password in the output:

    Paket failed with:
            Could not find versions for package SomePackageName in any of the sources in [Nuget
       {Url =
         "https://feed-url";
        Authentication = Some (ConfigAuthentication ("user name","actual password"));}].

We may want to skip printing the actual password there.

forki:
Yes we need to implement tostring for the Auth type. Care to send a PR?

cdrnet:
Sure. Would it also make sense to use `SecureString` instead of `String` for the password?

forki:
I guess it could be done, but most of the the passwords can be found on
your system (often in deps file) anyway.

cdrnet:
Fair enough, I'll just open a PR for ToString of Credentials etc for now. Thanks.

PR #1238, opened 2015-11-19
PackageSources: do not print feed password to output #1224

cdrnet:
In `NuGetV2.GetVersions` use the same approach as in `GetPackageDetails` to just print the feed url instead of the full record including the password, to avoid leaking the password unnecessarily to the console output.

It is correct that the password is lying around (at least encrypted) in paket.config, but it's still awkward seeing your personal password floating by on the console when pair-programming ;)

See also #1224


forki:
thx

PR #1241, opened 2015-11-19
Do not crash if no local version exists - fixes #1240

### LINKED ISSUE #1240: paket.bootstrapper.exe 2.27.0 fails
paket.bootstrapper.exe 2.27.0 fails with 

>.paket\paket.bootstrapper.exe
No version specified. Downloading latest stable.
**Input string was not in a correct format. (Github)**

inosik:
Sorry for this again, this makes me sad. :cry:


forki:
thx

PR #1245, opened 2015-11-19
Add's monoandroid41 framework moniker

gusennan:
monoandroid41 is a valid framework moniker

(Greetings from the Berlin Mobile .NET User's Group, @BerlinDotNet )


forki:
I assume it is. 

Cheers

PR #1246, opened 2015-11-20
ContinueOnError in paket.targets restore

forki:
the build in VS will fail anyways if restore forgot important stuff.
This is mainly a workaround because of VS locking


forki:
doesn't work

forki:
maybe it works

Issue #1254, opened 2015-11-23
No install when nuget package dependency has been unlisted.


philderbeast:
Paket cannot install a nuget package with a dependency on another nuget package when that other package has been unlisted. This is reproducable with the following minimal ``paket.dependencies`` ..

    source https://nuget.org/api/v2
    nuget ServiceStack.Host.AspNet

    C:\...>.paket\paket install
    Paket version 2.27.10.0
    Resolving packages for group Main:
     - ServiceStack.Host.AspNet 4.0.48
     - ServiceStack.OrmLite.SqlServer 4.0.48
     - ServiceStack.Server 4.0.48
    Paket failed with:
            Could not find versions for package WebActivator in any of the sources in ["https://nuget.org/api/v2"].

"The owner has unlisted this package. This could mean that the package is deprecated or shouldn't be used anymore."
SOURCE: https://www.nuget.org/packages/WebActivator

Even with this unlisted dependency nuget.exe itself can be used to install. Both the following nuget installs work ...

    C:\..>nuget install ServiceStack.Host.AspNet

    C:\...>nuget install WebActivator -Version 1.5.3

philderbeast:
Paket also fails on other commands, such as upgrade.

philderbeast:
I was able to work around this problem by adding one line to the ``paket.dependencies`` file, in order to  pickup the package that I had already installed before it became unlisted ...

    source https://nuget.org/api/v2
    source packages

    nuget ServiceStack.Host.AspNet

forki:
I jus talked to @maartenba - seems it's a NuGet v3 issue.

forki:
root cause is https://api-v3search-0.nuget.org/autocomplete?id=WebActivator&take=100000&prerelease=true doesn't show versions.

I worked around that, by retrying with v2 if no versions are reported.

forki:
@maartenba suggested to use https://api.nuget.org/v3-flatcontainer/WebActivator/index.json instead

PR #1255, opened 2015-11-24
#1254 typo in the lock file.

philderbeast:
The issue number is 1254 and the WebActivator package number trying to
be resolved is 1.5.3. The typo is some kind of off by one error ;-)


forki:
I guess you also have to rename the folder in /integrationtests/scenarios

forki:
thx a lot for taking care of this

Issue #784, opened 2015-04-18
Referencing unlisted packages should lead to a warning


AlexAtNet:
I've added unlisted nuget package https://www.nuget.org/packages/System.Web.Http/ to the and the Paket  resolved it without any warning. Maybe it would be useful to display a warning if the referenced package is unlisted on nuget?

forki:
I like that idea and I think it would be easy to add. ==> up-for-grabs

forki:
this was merged.

PR #1258, opened 2015-11-25
#784 warning when resolving unlisted package

alex-bogomaz:
hi. Please review this pull request. Now warning is displayed when unlisted package is resolved


forki:
I think we should only warn when the resolver terminates with unlisted packages.

forki:
thx for submitting. I move the code a bit. see 88b8023 - does this work for you?

alex-bogomaz:
sure, thanks!

Issue #1100, opened 2015-09-29
Allow specifying custom source for the bootstrapper to find paket.exe


amazingant:
As much as I love that paket auto-updates itself whenever I run my build scripts (loosely based on the ProjectScaffold project, so `build.cmd` calls `paket.bootstrapper` before doing anything else), new features sometimes break existing processes. In my work environment, for any third-party packages we use (including build tools like FAKE), we keep the latest .nupkg that we've tested against in our local package source.

To match this, it'd be useful to have a command-line option for the bootstrapper that tells it where to get `paket.exe` from, so that build scripts don't have to be updated to get the latest-and-greatest, but someone can run a new version of paket through some hoops before using it globally.

FWIW, it's rare that I actually have any concern with paket's auto-update (or new versions of FAKE, for that matter), and it's been great to see new features show up without having to do anything. This is an amazing tool that helps simplify otherwise complicated workflows, and the hard work put into this is very much appreciated!

mexx:
@amazingant can it be closed?

amazingant:
@mexx Just got a chance to try it at work and look at the code changes. While #1261 initially looked to solve the meat of this issue, the new `--nuget-source` parameter is limited to calling a web server. This originally made sense since the bootstrapper is intended to pull paket off nuget.org as a backup source when GitHub is down. Since the main `paket.exe` supports pulling `.nupkg` files out of a directory, I don't actually have a proper nuget "server" anywhere though, as my package source is a directory path pointing to an SMB share.

Actually running the new bootstrapper, I noticed that since it can't find the nuget API it's looking for, it decided to fallback to downloading paket off GitHub. While it should be a rare occurrence, if I forgot to mount that network share (or it went down for maintenance, or whatever), anything using paket during that time is going to get the latest release off GitHub. Rather defeats the original intent I had in opening this issue.

I can work on a pull request to make a few more changes, if that's alright? May take me a few days to get to it though.

forki:
yes please

2015-11-30 14:22 GMT+01:00 Anthony Perez <notifications@github.com>:

amazingant:
No complaints here as I'm happy working with C#, but slightly curious: how did the bootstrapper end up written in C# instead of F#?

forki:
we wanted to keep it small (in terms of KB) and therefore wanted to avoid FSharp.Core reference

amazingant:
Makes sense, and appears to have worked well given the size of the bootstrapper vs FSharp.Core.dll

[...]

PR #1261, opened 2015-11-26
#1100 Allow specifying Nuget Source and provide option to specify parameters with config file in bootstrapper

DavidHayden87:
I've added the ability to specify the NuGet source in the bootstrapper using the --nuget-source command line argument.
I also added the ability to specify the NuGet source, preferring NuGet, and a specific Paket version in appSettings in a config file for the bootstrapper. It seemed to me that this would be an easy way to commit the settings to a repo (certainly works for my needs!).


forki:
could you describe why you need to specify the source?

DavidHayden87:
I have a similar issue to #1100 where we host copies of the NuGet packages we use on our own internally hosted NuGet server. This allows us to keep track of the third party dependencies we use, and manage version upgrades. Our build servers do not have access to the web, and so cannot pull down Paket from Github or Nuget.org, so I need to be able to tell the bootstrapper to use an alternative Nuget server.

forki:
ok makes sense, but if I would be you I would really avoid that extra enterprisey nuget server with blessed versions. I mean that's why we introduced the paket.lock file...

still thanks for contributing. I guess it's a good intermediate solution for very enterprisey environments

PR #1266, opened 2015-11-27
Switched to using Chessie Nuget package.

### LINKED ISSUE #1253: Use Chessie package from Nuget
I am referencing Paket.Core in a project, as well as using Chessie. The way Paket includes a full copy of Chessie is causing lots of pain and suffering translating between the Paket Chessie and the regular Chessie assembly. Would it be possible to switch Paket.Core to reference the Chessie Nuget package?

xavierzwirtz:
Resolves #1253

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -184,7 +184,7 @@ Target "MergePaketTool" (fun _ ->
     CreateDir buildMergedDir
 
     let toPack =
-        ["paket.exe"; "Paket.Core.dll"; "FSharp.Core.dll"; "Newtonsoft.Json.dll"; "Argu.dll"]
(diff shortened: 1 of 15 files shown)


forki:
could you please also add it to the ILRepack phase:

https://github.com/fsprojects/Paket/blob/master/build.fsx#L193
https://github.com/fsprojects/Paket/blob/master/build.fsx#L204

xavierzwirtz:
Did that.

xavierzwirtz:
Thanks!

forki:
it's relased as alpha. please confirm that it works, then I will release.
Btw: the impact on the dll is 2KB. So it went from 3935 KB to 3937KB. I think that's OK ;-)

xavierzwirtz:
Tested `2.30.0-alpha002`. Everything looks good. Compiles and all my tests pass.

PR #1267, opened 2015-11-29
Changes required when Chessie would be internal - Do not merge

mexx:
If `Chessie` would provide a way to have an internal version of the types, we would have to apply those changes to make the compiler happy.
@forki can you review it, as I don't know whether some of the functions were being used by other tools, like `ionide`.

My thoughts about the way to provide internalized version:
The author of the library, like `Chessie`, would have to mark the types with a comment, something like `(*Paket:internal*)`. `Paket` could than parse the file and replace the comment with the content after the colon.

### DIFF
--- a/src/Paket.Core/ConfigFile.fs
+++ b/src/Paket.Core/ConfigFile.fs
@@ -159,7 +159,7 @@ let GetAuthenticationForUrl (source : string) url =
 let GetAuthentication (source : string) =
     GetAuthenticationForUrl source source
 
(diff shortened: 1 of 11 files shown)


forki:
that sounds pretty cool. I don't think other tools should use this internals anyway.

forki:
what do we do with this?

mexx:
As Chessie is now linked as library, we can forget about it. ;)

PR #1268, opened 2015-11-30
Add options to force Nuget source and use local file paths with bootstrapper

amazingant:
Added a `--force-nuget` command-line argument to the bootstrapper to get `paket.exe` from nuget rather than github, and to *not* automatically fallback to github as the `--prefer-nuget` option does. Did this by just setting the `FallbackStrategy` to `null`, as the bootstrapper appears to already handle that [here][1].

Additional changes coming to allow using a local file path for the nuget source specified with `--nuget-source`, rather than a web API.

[1]: https://github.com/fsprojects/Paket/blob/5da9a479b14527df509eec18315ab9a22365258c/src/Paket.Bootstrapper/Program.cs#L114

### DIFF
--- a/docs/content/bootstrapper.md
+++ b/docs/content/bootstrapper.md
@@ -14,7 +14,11 @@ Options:
   `version`: Downloads the given version of paket.exe from github.com.
 
   `--prefer-nuget`: Downloads the given version of paket.exe from nuget.org instead of github.com. Uses github.com as fallback, when nuget.org fails
-  
+
+  `--force-nuget`: As with the `--prefer-nuget` option, downloads paket.exe from nuget.org instead of github.com, but does *not* use github.com as a fallback.
+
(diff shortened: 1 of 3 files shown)


forki:
@DavidHayden87 could you please review. thx

amazingant:
@DavidHayden87 Should work now as of 14b9b4c; still need to update the code for the bootstrapper to update itself with.

amazingant:
@forki does it make sense to be checking the `latestVersion` value against `String.Empty` like this, versus using `String.IsNullOrWhiteSpace`? Trying to match the surrounding code.

forki:
I guess nullorwhitespace would be better

DavidHayden87:
I've pulled down the binary from AppVeyor and tried it in our case where we pull from an internal NuGet web server, and it all seems to still work.

amazingant:
@forki Just updated with changes to use `String.IsNullOrWhiteSpace` in the places I've touched, left all other use of `String.Empty` as-is for someone else to worry about. :wink: 

@DavidHayden87 Thanks!

PR #1272, opened 2015-12-01
Support local read-only .nupkg-files

Karamell:
We needed support for read-only unc-directories. Fixed by not opening the nupkg FileStream with write access.


forki:
thx

Issue #868, opened 2015-06-12
Creating symbol package with sources


domaslasauskas:
NuGet supports `-Symbols` parameter for [creating symbol package](https://docs.nuget.org/create/creating-and-publishing-a-symbol-package#creating-a-symbol-package) which contains source files (in `\src` inside the package).

Paket is missing a way to include sources in the package when using paket.template.

JohanLarsson:
Not sure this is a good idea but one way could be to transform the `.template` to a `.nuspec` and feed it to `nuget.exe`.

cdrnet:
I'd have a use case for this as well, although in my case sources would need to go into `/src/Common` as it is a multi-platform package where all variants are compiled from the same sources. I'd rather not call into NuGet.exe for this as that's what I'm trying to get rid of by using Paket :).

DavidHayden87:
I need this too... working on a pull request

kichristensen:
Could this not issue be closed, as @forki merged the PR back in December 2015?

PR #1275, opened 2015-12-01
Added ability to create symbol/source packages

DavidHayden87:
As per issue #868, I have added a symbols option to the pack command which will create symbol/source packages in addition to the regular packages, where for project templates.


src/Paket.Core/NupkgWriter.fs
```diff
@@ -141,7 +141,7 @@ let corePropsDoc (core : CompleteCoreInfo) =
     let xsi = XNamespace.Get "http://www.w3.org/2001/XMLSchema-instance"
     let root =
         XElement
-            (ns + "Relationships", XAttribute(XName.Get "xmlns", ns.NamespaceName),
+            (ns + "coreProperties", XAttribute(XName.Get "xmlns", ns.NamespaceName),
```
forki: is this correct? /cc @mavnn

DavidHayden87: I noticed this when comparing a package created with NuGet.exe vs. one created with Paket.exe, I'm assuming "Relationships" was a copy-paste error from the code for the .rels file.

forki: ok so it's unrelated. could you please create a second PR with only this fix?

DavidHayden87: Done #1276

DavidHayden87:
@forki Is this pull request still being considered, or do I need to create a new one without the coreProps commit?

forki:
no it's fine, but you can rebase this one if you like. but it's not that important

DavidHayden87:
I've done the non-trivial merge with other changes to the Pack command on master

forki:
goat

forki:
could you please take a look at the docs and add a bit about this feature? thx

DavidHayden87:
Created pull request #1279 for the documentation

PR #1283, opened 2015-12-02
Fixed auth and nugetURL being backwards on FindAutoCompleteVersionsForPackage

xavierzwirtz:
It was backwards, was confusing for a moment.


src/Paket.Core/NuGetV3.fs
```diff
@@ -101,9 +101,9 @@ let internal findAutoCompleteVersionsForPackage(v3Url, auth, packageName:Domain.
     }
 
 /// Uses the NuGet v3 autocomplete service to retrieve all package versions for the given package.
-let FindAutoCompleteVersionsForPackage(auth, nugetURL, package, includingPrereleases, maxResults) =
+let FindAutoCompleteVersionsForPackage(nugetURL, auth, package, includingPrereleases, maxResults) =
```
forki: I assume we have tons of this ...

Issue #1281, opened 2015-12-02
Use AssemblyTitle if no title is specified in template file


DavidHayden87:
When using NuGet and nuspec files for packaging, we made use of [replacement tokens](https://docs.nuget.org/create/nuspec-reference#replacement-tokens) in project nuspec files, to substitute in various values in the nuspec file.
When using Paket, many of these values are substituted for me when not specified, which made the switch over fairly easy.
However it is not exactly like-for-like, as we also used the $title$ replacement token (which does not appear to actually be documented - not sure how we found it!), which substitutes for AssemblyTitle value. Since our project names are simplified forms of the usually rather longer assembly names they create, we now have quite long titles. I'd like to avoid having to manually specify the title in each template file (currently they are mostly same, containing little to nothing which is project specific). I believe it would be fairly simple to update the packing logic to set the package title to be the same as the AssemblyTitle if it is not explicitly specified.
I'm happy to do the work...

forki:
Sounds good

PR #1285, opened 2015-12-03
Use AssemblyTitle if no title is specified in a project template

DavidHayden87:
As per issue #1281 if have updated the packing logic to use the AssemblyTitleAttribute value if not title is supplied in a project template file


forki:
thx

PR #1290, opened 2015-12-04
Option to force a binding redirects

mrinaldi:
This PR introduces the `redirects: force` to force a binding redirect to be written in the config as discussed in https://github.com/fsprojects/Paket/issues/1270#issuecomment-161368074.

It also introduces the redirects settings to references files.
I'm not sure if this was supposed to work before, but it wasn't and there were no doc regarding it.

Docs were updated accordingly.


forki:
very cool. /cc @isaacabraham @dnauck

forki:
awesome stuff

forki:
works like a charme

works

mrinaldi:
:smiley:

Issue #1291, opened 2015-12-04
User prompts do not work with redirected stdin


DavidHayden87:
We have multiple repositories all using NuGet for package management, and there are a number of steps I currently need to take in each to convert them to using Paket, beyond just calling convert-from-nuget.
I am trying to write a script which does all this for me, and calls into paket.exe for convert-from-nuget and simplify, redirecting the standard input/output so my script can make decisions about what to remove. But am having trouble when the executable prompts about deleting files:

> Paket failed with:
> 	Cannot read keys when either application does not have a console or when console input has been redirected from a file. Try Console.Read.

This seems to be due to using `Console.ReadKey()` which according to the [documentation](https://msdn.microsoft.com/en-us/library/471w8d85(v=vs.110).aspx) cannot read from redirected standard input. I am proposing changing this to use `Console.ReadLine()`, which can read from standard input. The consequence of this, is that users would have to hit y/n followed by enter, rather than just y/n.

Thoughts?

forki:
I don't think that would be an issue.

PR #1292, opened 2015-12-04
Make user prompts work with redirected stdin

DavidHayden87:
As per issue #1291, the prompt is now implemented with Console.ReadLine() which works with redirected standard input

### DIFF
--- a/src/Paket.Core/Utils.fs
+++ b/src/Paket.Core/Utils.fs
@@ -280,7 +280,7 @@ let safeGetFromUrl (auth:Auth option, url : string, contentType : string) =
         with _ -> return None
     }
 
-let readKey() = System.Console.ReadLine().Trim()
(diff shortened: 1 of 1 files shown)


forki:
Should we clarify the questions, then?

src/Paket.Core/Utils.fs
```diff
@@ -280,7 +280,7 @@ let safeGetFromUrl (auth:Auth option, url : string, contentType : string) =
         with _ -> return None
     }
 
-let readKey() = System.Console.ReadKey().KeyChar.ToString()
+let readKey() = System.Console.ReadLine().Trim()
```
forki: should be renamed

DavidHayden87: Fair enough, I've renamed to readAnswer

Issue #1248, opened 2015-11-22
Paket changes my config files


haf:
<img width="691" alt="screen shot 2015-11-22 at 10 38 05" src="https://cloud.githubusercontent.com/assets/193115/11323013/2b56b42e-9105-11e5-88d0-1b5b0a42b9f2.png">

Do you know why?

XML should not contain BOMs imo, as no other language really support them.

forki:
//cc @isaacabraham @mrinaldi

isaacabraham:
Urgh. Don't know why that's there. I think this should be done as part of the cleanup of the formatting of the BRs anyway (you can see from the ```</assemblyBinding></runtime>``` line that it's not formatted properly).

konste:
How about Paket leave XML formatting alone? We have never ending problem with Paket always changing our config files (incl. those it should not be touching at all). It keeps removing spaces in lines like 'value="false" />' So it removes the space between doublequote and slash and does it for ALL configs. We would leave it like that, but some other automated tool has different idea what the right spacing is and it keep adding that space back! This creates enormous churn and makes reviewing config file changes very hard. Please, please let Paket NOT CHANGE existing XML formatting.

forki:
Yes it's definitely the plan to keep formatting as stable as possible. Will
try to add couple of tests tomorrow

konste:
@forki, thank you for speedy response! It is always great to work with you!

forki:
@konste can you give me the app.config in question (which contains value = false) for a unit test?

konste:
[App.config.txt](https://github.com/fsprojects/Paket/files/50855/App.config.txt)

forki:
wow. even if we disable formatting it seems .NET always writes `value="false"/>` as `value="false" />`

konste:
I may look into it. Which exactly .NET Xml facility is employed here? XmlDocument? XDocument? 

But why rewrite configs which are not touched a bit at all?

forki:
that are two separate issues. one is we should not touched configs that we don't change logically. second is that we don't stuff outside of redirects

konste:
I understand that the second part is harder than it seems. Writing XML file is usually one call to a Save method (unless combination of reader and writer is used, which seems like overkill here), so we just need a way to preserve original formatting. So what .NET class is used to handle XML in Paket?

konste:
[...]

PR #1299, opened 2015-12-08
Make redirects off skip binding redirects

mrinaldi:
This PR makes `redirects: off` skip binding redirects completely, regardless the packages instruct otherwise as discussed in #1248.

It'll skip even when a package has a `redirects: forced`.


forki:
thanks so much.

PR #1311, opened 2015-12-15
NuGetV2.DownloadPackage: Add support for relative src link to package content

antymon4o:
Given that the nuget package feed result of a specific package defines the content of the package with a relative link (for example <content type="application/zip" src="Packages(Id='PackageName',Version='1.0.0.0')/$value" /> )
When NuGetV2.DownloadPackage creates the WebRequest to download the package contetns
Then it should add the SourceUrl as the base address for the download Uri.


forki:
thx a lot

antymon4o:
Thank you for the great project! :)

Issue #1265, opened 2015-11-27
Is it possible to create packages of differing versions with paket pack?


eiriktsarpalis:
I own a repo which contains multiple projects that depend on each other, each having its own `paket.template` file. The version of each nuget release is passed through FAKE using the parsed release notes file, and the package inter-dependencies are correctly registered using the `CURRENTVERSION` macro.

Consider now the following situation: I just added a new project to the repo which is mostly experimental. I would like to release it on nuget together with its siblings, but I would like to enforce a prerelease suffix on that package specifically. This seems to be impossible at the moment: if I attempt to hard-wire a version number on the specific `paket.template` it will get overriden by FAKE. At the same time, it seems that the `PaketPackParams` record in FAKE is global, as opposed to a per-template spec (and curiously enough, the `PaketPackParams.TemplateFile` property evaluates to null).

Any suggestions?

forki:
Mhm that's... Interesting ;-)
I have no good idea. But I'm open for changes...

eiriktsarpalis:
One idea would be to have the Packet.Pack lambda execute once for every parsed `paket.template` found in the repo. The `PaketPackParams` could thus be augmented with information such as dependencies, description, id, etc. In that case, I could fix my problem with something like
```fsharp
Paket.Pack (fun p -> 
    { p with 
        Version = if p.Id = "Experimental.Lib" then "0.0.1-alpha" else release.NugetVersion })
```

forki:
This sounds good from a fake standpoint. But how would this translate into
the paket cmd call

eiriktsarpalis:
Good point. I guess it would only be possible if the pack command could target/parse individual templates or if paket were used as a library by fake. Not sure what is best here.

forki:
Templates of type project are always processed together. Maybe we can
something into template instead?

eiriktsarpalis:
Perhaps it could be possible to pass separate version id's per project id. This would effectively add a separate command line option
```
paket pack version 1.1.2 --custom-version Experimental.Lib 0.0.1-alpha
```
which would override the default version number in case the `paket.template` id matches the first argument of the command. This could be then specified on the FAKE side by adding a `CustomVersions : Map<string, string>` property to the `PaketPackParams` record. Not as nice as the lambda approach, but certainly a less intrusive solution given the current state of things. Thoughts?

forki:
Yes that would work too.

eiriktsarpalis:
[...]

PR #1314, opened 2015-12-18
Implement support for specifying per-template versions in paket pack

eiriktsarpalis:
This PR attempts to address the problem described in #1265. It incorporates the possibility to supply specific version numbers on a per-template basis:
```
paket pack version 1.1.2 specific-version Experimental.Lib 0.0.1-alpha specific-version Another.Experimental.Lib 1.0-beta
```
This has the behaviour of overriding the version number in specified template id's and ensures that this is reflected in wherever `CURRENTVERSION` is being replaced. When running the above command, the following file:
```
id Another.Experimental.Lib
dependencies
     Core.Lib CURRENTVERSION
     Experimental.Lib CURRENTVERSION
```
should generate a package equivalent to the template below:
```
id Another.Experimental.Lib
version 1.0-beta
dependencies
     Core.Lib 1.1.2
     Experimental.Lib 0.0.1-alpha
```
Please review. I will be posting a corresponding PR to the FAKE paket components to reflect this change.


inosik:
I like this :+1:

I'm currently working on something, which would help us to keep the `dependencies` node out of `paket.template`, so we don't need to repeat what we might have in `paket.references`.

Look [here](https://github.com/inosik/Paket/tree/project-deps/constraints) for more.

forki:
thx. and sorry that I didn't fix that other issue yet.

eiriktsarpalis:
Thanks! I've updated my code to use the latest update. Here's what it looks like: https://github.com/mbraceproject/MBrace.Core/commit/807eb2ffb408ebfb8da50352bf73624e544130cb#diff-6074983490330a0aac1bd23c958232d7R105

PR #1323, opened 2015-12-21
Fixes #1321

### LINKED ISSUE #1321: paket pack ignores specific-version parameter when inserting dependencies making created packages unusable
Say I have a solution with 3 packages that are packed using specific-version for each package.
A (AssemblyVersion: 3.3.0.0, specific-version: 3.3.0-beta01)
B (AssemblyVersion: 3.1.0.0, specific-version: 3.1.0-beta01, DependsOn: A)
C (AssemblyVersion: 3.5.0.0, specific-version: 3.5.0-beta01, DependsOn: A)

In the created packages the dependency on A in B and C are wrong. They both specify the version: 3.3.0.0 for A instead of the correct 3.3.0-beta01

eiriktsarpalis:
Fixes issue as described in #1321

### DIFF
--- a/src/Paket.Core/PackageProcess.fs
+++ b/src/Paket.Core/PackageProcess.fs
@@ -10,11 +10,11 @@ open System.Collections.Generic
 open Paket.PackageMetaData
 open Chessie.ErrorHandling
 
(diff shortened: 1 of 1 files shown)


forki:
thx

Issue #1313, opened 2015-12-18
http dependencies on file system are not stored in paket-files directory


jruizaranguren:
According to the docs, the following example:

```F#
http http://www.fssnip.net/raw/1M/test1.fs src/test1.fs
```

is stored at the directory `paket-files\www.fssnip.net\src\test.fs`

I'm referencing a local file through:

```F#
http file:///absolute/path/in/windows/to/package.dll package/package.dll
```

But the file is stored at `C:/C_/package/package.dll`

Am I doing something wrong or it is just improvable behavior?

forki:
I guess you are the first person who tried file protocol.

mexx:
Can you try to use `localhost` as host in your url.
```
http file://localhost/absolute/path/in/windows/to/package.dll package/package.dll
```

jruizaranguren:
@mexx, i get the error `The given path's format is not supported.`

jruizaranguren:
Any hint on where to look in code for this? Thanks

jruizaranguren:
Solved in 2.39.6

PR #1325, opened 2015-12-22
a couple of tests to discuss the behavior for http file protocol, as issued in #1313

jruizaranguren:
The desired functionality is:

* file protocol works for http sources, i.e.:  `file://localhost/absolute/path/to/file1.txt`
* when three slashes are used `localhost` is inferred.

I am getting the following error that I have not been able to reproduce through a test: `The given path's format is not supported`. If I get some hints to create this test I may be able to fix it.
Thanks.

### DIFF
--- a/tests/Paket.Tests/DependenciesFile/ParserSpecs.fs
+++ b/tests/Paket.Tests/DependenciesFile/ParserSpecs.fs
@@ -374,6 +374,45 @@ let ``should read http source file from config without quotes with file specs``(
             Commit = Some "/raw/1M/1"
             AuthKey = None } ]
 
+[<Test>]
+let ``should read http source with file protocol and interpret three slashes as localhost`` () =
+    let config = """http file:///absolute/path/to/file1.dll"""
+    let dependencies = DependenciesFile.FromCode(config)
+    dependencies.Groups.[Constants.MainDependencyGroup].RemoteFiles
+    |> shouldEqual
+        [ { Owner = "localhost"
+            Project = ""
+            Name = "file1.dll"
+            Origin = ModuleResolver.SingleSourceFileOrigin.HttpLink "file://localhost"
+            Commit = Some "/absolute/path/to/"
+            AuthKey = None } ]
+
+[<Test>]
+let ``should read http source with file protocol and file specs`` () =
+    let config = """http file://localhost/absolute/path/to/file1.dll file1/file1.dll"""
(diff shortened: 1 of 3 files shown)


forki:
seems the tests are already failing

jruizaranguren:
Yes, they are. They just explain the proposed conventions. I wanted to know if you think that is desired behavior or not before change it.

forki:
I don't really understand the use case and why is file a special case of http?

jruizaranguren:
Well, I guess it should not be. But it is not currently working with Paket. 
* When I add `file://localhost/absolute/path/to/file1.txt` to `paket.dependencies` the file is downloaded but I get the error `The given path's format is not supported`
* When I use `file:///absolute/path/to/file1.txt` file is not found.
So I guess something must be made somewhere (I have not already found) in Paket in order to process correctly file:// sources.

forki:
I just tried with the following paket.dependencies file:
    
    http file:///c:/code/myProject/myFile.txt

this works on windows and is the correct format according to Wikipedia.

jruizaranguren:
When I use the format you wrote, such as: 
```
http file:///c:/Projects/POCs/index.txt
```

It is downloaded to `C:\c_\index.txt`, instead of being downloaded to a path relative to the `paket-files`:
directory:

image

Is this intended behavior?

According to the Wikipedia article, and also to the [original RFC](https://tools.ietf.org/html/draft-ietf-appsawg-file-scheme-03#appendix-B.2), the following alternatives should be valid:

```
http file:///c:/Projects/POCs/index.txt
http file://localhost/c:/Projects/POCs/index.txt
http file://hostname/c:/Projects/POCs/index.txt
```

Anyway it is perfectly reasonable to only allow for one of them. Although I think that it would be better and safer if files are downloaded to a path relative to paket-files.

Perhaps these are two different issues:

* Allow the different representations of http file-protocol (could be dismissed)
* Download the files specified with http file-protocol to a path relative to paket-files.

forki:
1) Allow the different representations of http file-protocol (could be dismissed)

Currently we oly support what the underlying .NET API is doing. We might allow more, but then we would need to internally map the URIs.

2) Download the files specified with http file-protocol to a path relative to paket-files.

Yep that seems to be a bug. Will check

forki:
please try 2.39.6

jruizaranguren:
It works. Thanks.

Issue #1320, opened 2015-12-19
ability to specify which sources to pull the dependencies from


ctaggart:
Paket looks for the nuget dependencies in all of the sources for a group. It would be great if we could limit the sources used to lookup the dependencies. This change would increase performance and reduce the unnecessary load that is put on NuGet feeds.

Using this as an example, I named the `source`s and added `from` lists.
https://github.com/ctaggart/SourceLink/blob/master/paket.dependencies

```
source nuget https://www.nuget.org/api/v2
source appveyor https://ci.appveyor.com/nuget/ctaggart-sourcelink

from nuget appveyor
  nuget SourceLink.Fake 1.1.0-b279

from nuget
  nuget FAKE
  nuget FSharp.Formatting
  nuget NuGet.CommandLine
  nuget RazorEngine
  nuget FSharp.Core
  nuget LibGit2Sharp
  nuget Mono.Cecil
  nuget UnionArgParser
  nuget System.Threading.Tasks.Dataflow
  nuget Argu
```

Performance related tickets this would help are #1317 & #1249.

forki:
Our first implementation did it like this, but unfortunately for transitive
dependencies we need to change the source.

ctaggart:
I'm not clear what wouldn't work here with transitive dependencies. May be we can work through a better example.

forki:
Sourcelink.fake might have a dependency on something that is on the other
feed.

ctaggart:
The logic I was imagining, but didn't write down allows for that. It doesn't, but let's pretend SourceLink.Fake had a dependency on [canopy](https://www.nuget.org/packages/canopy/). canopy then depends on FSharp.Core, Selenium.WebDriver, Selenium.Support, and SizSelCsZZZ. In the above example, I put down:
```
from nuget appveyor
  nuget SourceLink.Fake 1.1.0-b279
```
meaning that it should look for SourceLink.Fake in both the `nuget` feed and `appveyor` feed for it and its dependencies. After finding Canopy in that list of feeds, it would look for its transitive dependencies in that list of feeds unless it is a named dependency. In this example, FSharp.Core is a named dependency that should be looked up in only the `nuget` feed.

In summary, a named dependency has a list of sources to look its dependencies. Unnamed transitive dependencies inherit that list. Does that work? What are the edge cases that I'm missing?

forki:
yes that would work and is basically what we had earlier. but in most common scenarios basically everything went under the complete list

[...]

Issue #1317, opened 2015-12-19
really bad performance using AppVeyor feeds with > 2.17.3


ctaggart:
We have 60 dependencies in a `paket.depdendencies` with 55 of them from NuGet Gallery and 5 of them  our AppVeyor feed. Paket 2.17.3 took 45 seconds to do a `paket update`. Paket 2.18.0 took 9 minutes.

Paket 2.17.3 will tries 1 request per missing dependency to AppVeyor:

```
GET https://ci.appveyor.com/nuget/ourfeed/package-versions/Auth0?includePrerelease=true HTTP/1.1
Authorization: Basic 
Accept: application/atom+json,application/json
User-Agent: Paket
Host: ci.appveyor.com

HTTP/1.1 200 OK
Cache-Control: private
Content-Type: application/json; charset=utf-8
Server: Microsoft-IIS/8.5
X-AspNet-Version: 4.0.30319
X-Powered-By: ASP.NET
Date: Sat, 19 Dec 2015 05:26:07 GMT
Content-Length: 2

[]
```

Paket 2.18.0 and above makes 4 requests per missing dependency to AppVeyor:

```
GET https://ci.appveyor.com/nuget/ourfeed/Packages?$filter=Id%20eq%20'Auth0'%20and%20NormalizedVersion%20eq%20'1.11.3' HTTP/1.1
Authorization: Basic
Accept: application/atom+xml,application/xml
User-Agent: Paket
Host: ci.appveyor.com

HTTP/1.1 400 Bad Request
Cache-Control: private
Content-Length: 290
Content-Type: application/xml
Server: Microsoft-IIS/8.5
DataServiceVersion: 1.0;
X-AspNet-Version: 4.0.30319
X-Powered-By: ASP.NET
Date: Sat, 19 Dec 2015 05:29:04 GMT

<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<error xmlns="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata">
  <code></code>
  <message xml:lang="en-US">No property 'NormalizedVersion' exists in type 'Appveyor.Models.Package' at position 18.</message>
</error>
```

```
GET https://ci.appveyor.com/nuget/ourfeed/Packages?$filter=Id%20eq%20'Auth0'%20and%20Version%20eq%20'1.11.3' HTTP/1.1
Authorization: Basic 
Accept: application/atom+xml,application/xml
User-Agent: Paket
Host: ci.appveyor.com

[...]

PR #1327, opened 2015-12-23
Cache which source contains versions in GetVersions

forki:
This is should improve the second update run since we cache if a nuget feed doesn't have versions for a given package. It should be a good alternative to #1320
/cc @ctaggart


ctaggart:
This cut the `paket update` time in half to about 25 seconds my project that I was using in #1317. How does caching the sources work? What happens when you add the same package to another feed? That was a common scenario an my previous employer.

forki:
Just run "paket update -f" and it will find that new package on all feeds.

PR #1330, opened 2015-12-26
Observable.flatten bugfix

ForNeVeR:
Current implementation of `Observable.flatten` never calls `obs.OnComplete`, and that leads to a bug in Paket.VisualStudio (and possibly adds problems for another users of this API).

Instead of writing that method here (it could be difficult for a general case), I've decided to simply use `FSharp.Control.Reactive` instead.

This PR will close https://github.com/fsprojects/Paket.VisualStudio/issues/60.

Please note that you'll have to include `FSharp.Control.Reactive.dll` in a future builds of Paket.VisualStudio, and I've found that it's not trivial to achieve (because Paket.VisualStudio have no direct dependency on this package). In my local environment I got this fixed by adding reference to FSharp.Control.Reactive into a main Paket.VisualStudio assembly, but maybe there is some more conventional way.

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -3,6 +3,7 @@ source https://nuget.org/api/v2
 nuget Newtonsoft.Json redirects: force
 nuget Argu
 nuget FSharp.Core redirects: force
+nuget FSharp.Control.Reactive
 nuget Chessie
 
 github fsharp/FAKE src/app/FakeLib/Globbing/Globbing.fs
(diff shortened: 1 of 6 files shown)


forki:
thx for hunting down this bug. But the dependencies is very big, so I tried to fix it directly. Could you please review: afe98e4

thx

ForNeVeR:
@forki I got your point, I'll review (and test) that change tonight. Thanks for the answer.

ForNeVeR:
@forki yup that works very good (I've tried the latest releases Paket.VisualStudio version), and the code looks ok to me. Thank you!

Issue #1287, opened 2015-12-03
Include hashes in lock file, for NuGet packages too?


epsitec:
I am wondering if I am the only person who'd like to make sure that the packages that get restored after a clean checkout from source control are indeed the ones that were used when the lock file was commited.

I feel incomfortable at the idea that somehow, somebody could hack the NuGet repository and that I would end up with other NuGet packages as what I used when I tested my code (e.g. when switching from testing to production, and rebuilding the code before it gets into the customer's hands).

I have the same issue with NPM... but maybe I am just paranoid?

forki:
Why would we want to do this?

epsitec:
To avoid poisoning a build system, and having paket detect that there is a mismatch between the binaries which were used when the lock file was commited, and when the product finally got built.

forki:
Mhm, yeah there are build system which do this. It's probably not a bad
idea.
I'm not completely against it.

cdrnet:
Wasn't there the problem that at least in the public NuGet gallery, NuGet packages could actually change their hashes over time (since some metadata was changeable)?

forki:
The issue was that we can't trust the hash that was given from the api. But
in theory we could hash the zip locally and pin that hash.

epsitec:
Computing a local hash and pinning it would be a great feature indeed.

mlidbom:
Doing this by default could cause very significant friction in some cases. 

An example: 
Developer works on Project A and B. Project B references A through a nuget package. 
Developer locally(on his machine) bumps project A and updates the package reference in project B. 
Developer's local version of A.nupkg is hashed and set as correct in the lock file.
Developer pushes the new version of A and B.
The CI server builds A and publishes the new version of A.nupkg.
The CI server tries to build B but fails since the hash of A.nupkg built by the CI server does not match the hash in the lock file.

bacar:
Isn't that an example of Paket doing _exactly_ the right thing; it _should_ complain because the downloaded package _isn't_ the one that was used at the time - ensuring you get the right dependency seems to be exactly the goal of Paket. 

I think this will become more important with Paket 3 network share caches as espoused here: https://fsprojects.github.io/Paket/caches.html
For this to be useful I basically have to grant everybody write access to the network share, which means anybody could come and stuff up existing cached packages.

isaacabraham:
You could also use a local / network cache to provide a trusted repository of your nuget dependencies.

smoothdeveloper:
[...]

PR #1332, opened 2015-12-28
initial work around adding the hash of a nuget package to the lockfile

baronfel:
This PR begins to address #1287 by adding a Hash parameter to the ResolvedPackage concept and plumbing that through the lockfile.
The hash is stored as the Base64-encoded string of the SHA256 bytes of the underlying .nupkg for a particular nuget id's version, though of course this is negotiable.  This format was chosen for readability purposes mostly.

There's a hairy chunk of code in the install process where I have to save the lockfile after we create the install model but before we use the model to apply changes to the references files, because the lockfile isn't saved past this point and we want to be able to persist the hashes of the packages that we just downloaded.

I still need to add tests, but it's at least in a locally-working state.  Thoughts?

### DIFF
--- a/docs/content/dependencies-file.md
+++ b/docs/content/dependencies-file.md
@@ -235,6 +235,19 @@ If you want to influence the resolution of [transitive dependencies](faq.html#tr
 
 To override a `lowest_matching` option for a single NuGet package, you can use the package specific [lowest_matching option](nuget-dependencies.html#Lowest_matching-option).
 
+## Hash option
+
+This option opts in to Paket's package verification mechanism.  This mechanism verifies the Base64-encoded SHA256 hash of any Nuget packages downloaded from any source against a value that is recorded in the `paket.lock` file.
+
+To enable this feature, add the `hash: on` directive to either the top-level of your `paket.dependencies` file, or within a particular desired dependency group.
+
+	hash: on
+	source https://nuget.org/api/v2
+
+	nuget FAKE
+
+The valid values for this directive are `on`, `off`, or no directive.  If no directive is given, then the lock file will not be modified in any way.
+
 ## Comments
 
 All lines starting with with `//` or `#` are considered comments.
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -150,3 +150,7 @@ For example:
     set HTTP_PROXY=http://user:password@proxy.company.com:port/
     set HTTPS_PROXY=https://user:password@proxy.company.com:port/
     set NO_PROXY=.company.com,localhost
+
+## Can I use paket to verify the contents of the packages I use?
+
+Yes!  Paket provides an opt-in hash check mechanism to verify that packages that are used are the same after each package restore.  To opt into this, simply add the `hash: on` directive to your `paket.dependencies` file.  Then, perform a forced restore to cause your `paket.lock` file to be refreshed with the hash codes of your packages. See [the docs](dependencies-file.html) for more information.
--- a/integrationtests/Paket.IntegrationTests/HttpSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/HttpSpecs.fs
@@ -21,7 +21,7 @@ http file:///%s/library.dll library/library.dll""" (root.Replace("\\","/"))
 
 http file://%s/library.dll library/library.dll""" (root.Replace("\\","/"))
 
-    File.WriteAllText(Path.Combine(root,"paket.dependencies"),if isMono then monoDeps else deps)
+    File.WriteAllText(Path.Combine(root,"paket.dependencies"), if isMonoRuntime then monoDeps else deps)
 
     directPaket "update" "i001341-http-dlls" |> ignore
     
--- a/integrationtests/Paket.IntegrationTests/LoadingScriptGenerationTests.fs
+++ b/integrationtests/Paket.IntegrationTests/LoadingScriptGenerationTests.fs
@@ -54,7 +54,7 @@ let ``simple dependencies generates expected scripts``() =
       "include.nunit.fsx"
   |]
   
-  if not isMono then // TODO: Fix me
+  if not isMonoRuntime then // TODO: Fix me
     Assert.AreEqual(expectedFiles,actualFiles)
 
 let assertNhibernateForFramework35IsThere scenario =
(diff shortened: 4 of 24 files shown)


forki:
this sounds cool, but we need to make this backwards compatible.

baronfel:
cluttered the github view by rebasing off of latest master so that the CI checks would fire.

baronfel:
ok, this is more like it.  now the only failure is in the one test that checks that a framework restriction is maintained, and this test now breakes because we add a hash in.  The way the hash is parsed/tracked is part of the install settings, and previous versions of paket can still parse that section, due to how we parse all of the key value pairs and then extract out meaningful ones.

The only problem that I can see now is that if an older paket version reads and saves the lockfile then the hash will be wiped.  Not breaking, because then the new version will detect and write a new one, but I could see excessive thrashing of the lockfile causing folks angst when they have the file checked into source control.

Do you have any thoughts on how else we might ease some of that pain?

forki:
if the new version can read old lock files and doesn't fail on the missing hash then it's fine.
We also should allow to disable the hashing feature

baronfel:
Should I add this hashing toggle to the overall install options structure, with global redirects and content like so: `hash:on` or `hash:off` and then add handling for this option to the dependencies and lock file parsing/logic?

forki:
Yes I think we should add a toggle in paket.dependencies and disable it by
default. At least for now

baronfel:
Sounds great!  I'll do that, then add tests and docs and then ping you when we're all ready to go.

baronfel:
ok, this is ready for review, but there's a problem with checking the generated hashes between mono and the MS CLR.  investigating that....

forki:
Cool. Let me know when you fixed that hashing issue. I plan to pull this into v3

baronfel:
It'll probably be a couple days.  I need to setup a vm to really get to the bottom of it and my Monday/Tuesday are packed :(

baronfel:
still running into issues, but some major clues are that the test package that I'm downloading are reporting as different file sizes on Linux and Windows.  

These report the first three bytes, then the stream length, and then the computed hash.

windows:
```
> makeHash (FileInfo("C:/code/Paket/integrationtests/scenarios/hash-calculation-on/temp/packages/Owin/Owin.1.0.0.nupkg"));;
504b3
4571
val it : string = "K+iPB4EhXvkBZ41mz3xetApTOeA5qRR4+r9g8mdt8FQ="
```

osx:
```
504b3
3741
val it : string = "large hash that is different that I don't want to copy"
```

baronfel:
The Owin.1.0.0.nupkg downloaded by Paket during the test on OSX cannot be opened in the NuGet Package Explorer, but can be renamed to .zip, extracted, and contains all of the expected contents.  It's almost like it's missing ~1KB of metadata during the download.

baronfel:
Testing the hash generation code on OSX with the correctly-sized-package results in a test pass, so now I have to figure out why Paket on my OSX box is downloading a stunted package.

baronfel:
I'm not sure, because it downloads correctly from paket on windows, and from nuget.org directly.  It only seems like paket on mono is having a problem with this package so far.

maartenba:
Package definitely not broken.
Also FYI, not sure if you are using the hash from the API to compare, but NuGet uses SHA512.

mlidbom:
[ Comment mostly pasted from #1287 ]
Doing this by default could cause very significant friction in some cases.
So I would argue that the ability to turn this off is a must have and that it should probably remain off by default.

An example: 
Developer works on Project A and B. Project B references A through a nuget package. 
Developer locally(on his machine) bumps project A and updates the package reference in project B. 
Developer's local version of A.nupkg is hashed and set as correct in the lock file.
Developer pushes the new version of A and B.
The CI server builds A and publishes the new version of A.nupkg.
The CI server tries to build B but fails since the hash of A.nupkg built by the CI server does not match the hash in the lock file.

baronfel:
Going to try to rebase this off of v3 this weekend and see if I can try to find the mono issue again.

cdrnet:
@baronfel  Any luck with the rebase? I'd also be interested to opt-in to lock-file based hash validation in a few repos.

baronfel:
@cdrnet wow I completely forgot about this. I'm on vacation so I have the time, so I'll see if I can get it done today. At least the rebase portion.

baronfel:
Whew, rebased. Let's see what CI says and go from there.

baronfel:
CI suggests I need to add netstandard-1.6 compatible hash library usage to get the windows side to build, and then there's an odd dotnetcore build issue happening on Travis.

baronfel:
@forki should this go against the preview3 coreclr work then, because I'm having a hard time building against master with the dotnetcore-specific code.

baronfel:
went ahead and rebased off of the DNC branch. My current pain point is in Paket.Core.preview3/Utils.fs.  I'm trying to conditionally include System.Security.Cryptography.Algorithms, which is in the paket.references for that project, but the reference cannot be found.  I'm not quite sure how to `install` paket references in the brave new world :-/

baronfel:
Ok finally got dotnetcore working on my laptop and got the `Utils.fs` file to build. Let's see what CI says now.

baronfel:
Ok, CI looks decent.

The failures are 
Appveyor
1) Path too long exception? I don't think this is because of this PR.
Travis
1) Some sort of local download that does not have the expected path. Again don't think this was because of this PR
2) Here's the _actual_ failure.  This is the same issue we had that blocked me earlier. @cdrnet This is where any help/ideas you have would be appreciated.  I could also try the test with another package entirely.  All this test is doing is using Paket to install a package (Owin version 1.0.0) and see if the calculated hash of the downloaded package matched what's in the test. 


I actually have had a hard time getting to what should be the 'real' value for the hash.  I've tested a couple downloaded packages from nuget, and haven't yet been able to get Nuget's own metadata with the hash, the one I compute in Paket now, or OSX's native system utilities to generate the same hash.

FYI to generate the hash on osx I'm using `shasum -a 512 PATH_TO_NUPKG | awk '{print $1}' | openssl base64` to generate it.

cloudRoutine:
@baronfel do you want to try starting this over?

baronfel:
I could rebase it again, surely.  The core logic isn't very large at all, just calling out to the crypto libraries to compute hashes.  The underlying problem that I ran into with computing different hashes for the same package will likely still be there, though :(

matthid:
I changed the base branch because core3 is quite old and this PR prevents me from removing it...

baronfel:
Thanks!

matthid:
@baronfel is this still something you are interested in?

forki:
closing for inactiviy. Please send a new one if you are still interested

PR #1336, opened 2016-01-02
WIP: Allow to use full git protocol

forki:
- [x] port git helper from FAKE
- [x] parse git urls from dependencies file
- [x] parse git urls from lock file
- [x] clone/fetch repos in user folder cache
- [x] clone/fetch repos in paket-files folder
- [ ] docs


src/Paket.Core/LockFile.fs
```diff
@@ -119,15 +119,26 @@ module LockFileSerializer =
...
                         updateHasReported.Remove GistLink |> ignore
                         updateHasReported.Add GitHubLink
                     yield sprintf "  remote: %s/%s" owner project
                     yield "  specs:"
+                | GitLink url ->
+                    if not (updateHasReported.Contains(GitLink(""))) then
+                        yield "GIT"
+                        updateHasReported.Remove GitHubLink |> ignore
+                        updateHasReported.Remove GistLink |> ignore
+                        updateHasReported.Add (GitLink "")
```
mexx: Shouldn't also `HttpLink ""` be reported as removed?

src/Paket.Core/LockFile.fs
```diff
@@ -353,7 +367,7 @@ module LockFileParser =
 
                             { currentGroup with
                                 LastWasPackage = false
-                                SourceFiles = sourceFile :: currentGroup.SourceFiles }::otherGroups
+                                SourceFiles = sourceFile :: currentGroup.SourceFiles }::otherGroups
```
mexx: Is this trailing whitespace needed?

src/Paket.Core/RemoteDownload.fs
```diff
@@ -189,7 +239,35 @@ let downloadRemoteFiles(remoteFile:ResolvedSourceFile,destination) = async {
...
+                            let hash = Git.CommandHelper.runSimpleGitCommand repoFolder ("rev-parse head")
+                            hash = gitRepo.Commit
+                        with
+                        | _ ->
+                            // something is wrong with the repo - recreate it
+                            Directory.Delete(repoFolder,true)
+                            false
+                    else false
+
+                if isInCorrectVersion then
```
mexx: Shouldn't it be `if not force && isInCorrectVersion then` ? Just like for the other origins.

src/Paket.Core/RemoteDownload.fs
```diff
@@ -132,10 +147,41 @@ let rec DirectoryCopy(sourceDirName, destDirName, copySubDirs) =
...
 let downloadRemoteFiles(remoteFile:ResolvedSourceFile,destination) = async {
     match remoteFile.Origin, remoteFile.Name with
+    | SingleSourceFileOrigin.GitLink cloneUrl, _ -> 
+        let cloneUrl = cloneUrl.TrimEnd('/')
+        
+        let repoCacheFolder = Path.Combine(Constants.GitRepoFolder,remoteFile.Project)
+
+        // clone/fetch to cache
+        if Directory.Exists repoCacheFolder then
+            verbosefn "Fetching %s to %s" cloneUrl repoCacheFolder
```
mexx: Please note, there will be no trace message (non verbose) when the cache folder is already present.
With the other comment in mind, about forcing the download, the cache should be ignored in this case, right?

src/Paket.Core/LockFile.fs
forki: good catch

src/Paket.Core/RemoteDownload.fs
forki: yep

forki:
This is now in v3 branch and first prerelease is on nuget

Issue #1346, opened 2016-01-05
Strange Crypto Exception on some Windows machines


haraldsteinlechner:
On some machines i get a crazy windows encryption exception. The error occurs when protecting the password. 
Most likely this issue only happens if users only use a domain user but have no local admin user on the machine.
(have a local fix)

![paketerror](https://cloud.githubusercontent.com/assets/513281/12120568/977f84e8-b3d1-11e5-9e75-5c0d61302e68.png)

forki:
please send pull request

haraldsteinlechner:
https://github.com/fsprojects/Paket/pull/1347

PR #1347, opened 2016-01-05
ConfigFile: password encryption does not work on specific machine con…

haraldsteinlechner:
This pull request solves issue #1346

Note that i did not change DataProtectionScopes for decryption. This is is the minimal change i know which works (i think decrpytion ignores this). 
The docs do not say anything about this: e.g. https://msdn.microsoft.com/en-us/library/2fh8203k(v=vs.110).aspx

### DIFF
--- a/src/Paket.Core/ConfigFile.fs
+++ b/src/Paket.Core/ConfigFile.fs
@@ -55,7 +55,12 @@ let private getRandomSalt() =
 /// Encrypts a string with a user specific keys
 let Encrypt (password : string) = 
     let salt = getRandomSalt()
-    let encryptedPassword = ProtectedData.Protect(Encoding.UTF8.GetBytes password, salt, DataProtectionScope.LocalMachine)
+    let encryptedPassword = 
+        try 
+            ProtectedData.Protect(Encoding.UTF8.GetBytes password, salt, DataProtectionScope.CurrentUser)
+        with | :? CryptographicException as e ->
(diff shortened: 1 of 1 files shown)


forki:
mhm this doesn't work on travis!?

haraldsteinlechner:
ah ok. awesome. no solution works on all configurations. We could do a simple probing - handle exception and try in local machine mode if user mode fails. what you think?

forki:
sounds good. whatever works to make it work for more people is fine

haraldsteinlechner:
works now

src/Paket.Core/ConfigFile.fs
```diff
@@ -55,7 +55,12 @@ let private getRandomSalt() =
 /// Encrypts a string with a user specific keys
 let Encrypt (password : string) = 
     let salt = getRandomSalt()
-    let encryptedPassword = ProtectedData.Protect(Encoding.UTF8.GetBytes password, salt, DataProtectionScope.CurrentUser)
+    let encryptedPassword = 
+        try 
+            ProtectedData.Protect(Encoding.UTF8.GetBytes password, salt, DataProtectionScope.CurrentUser)
+        with | :? CryptographicException as e ->
+            traceWarnfn "could not protect password: %s\n for current user" e.Message
```
forki: i think a verbosefn might be enough here

forki:
thanks a lot

Issue #1348, opened 2016-01-05
Paket pack messes NPM type paths with @ char


Vidarls:
Example: 

`font-awesome@4.5.0`

becomes

`font-awesome%404.5.0`

I'm using octopus deploy to deploy a web application that uses some npm type packages and I use paket to package the project. 

From looking at paket code, this seems to be more a feature than a bug, as Escape URI string seems to be used:

```f#
let ensureValidName (target: string) =
        let escapedTargetParts = target.Replace("\\", "/").Split('/') |> Array.map Uri.EscapeDataString
        let escapedTarget = String.Join("/",escapedTargetParts)

        let uri1 = Uri(escapedTarget, UriKind.Relative)
        let uri2 = Uri(uri1.GetComponents(UriComponents.SerializationInfoString, UriFormat.SafeUnescaped), UriKind.Relative)
        
        uri2.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped)
```

ref: https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NupkgWriter.fs#L229

It does however give me some 404 headaches now..

forki:
ideas how to solve this?

forki:
maybe @ sign needs to be excluded. PRs welcome

Vidarls:
Probably a good idea to be as explicit as possible as long as no-one gives any generic solutions.
I'll try to put something together that will leave the `@` char alone. 
(Although I think this will also be a problem with international characters, like the scandinavian `æøå` and the german special double ss char. All of which are valid folder names in windows (at least `æøå` is), but here will be replaced with some hex code)

forki:
Thx

Vidarls:
Working now, thanx for express release and merge :)

PR #1351, opened 2016-01-06
Fixing problem with @ char in paths during pack (#1348)

Vidarls:
ref: #1348


forki:
the travis fail is unrelated.
thx

integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -17,4 +18,22 @@ let ``#1234 empty assembly name``() =
     with
     | exn when exn.Message.Contains("PaketBug.dll") -> ()
 
-    File.Delete(Path.Combine(scenarioTempPath "i001234-missing-assemblyname","PaketBug","paket.template"))
\ No newline at end of file
+    File.Delete(Path.Combine(scenarioTempPath "i001234-missing-assemblyname","PaketBug","paket.template"))
+
+[<Test>]
+let ``#1235 npm type folder names`` () =
```
forki: Awesome.

Vidarls:
ok, good. Was scratching my head severely now

Vidarls:
Quick question, what is the logic behind the # 1234 type naming in the integration tests? I just did the above test number +1, but now I am starting to think it might be references to github issues?

forki:
yep test name should be "#1348 ..."

forki:
this allows us to find the issue from the test (when we break it in the future)

Vidarls:
fixing...

Vidarls:
fixed

PR #1353, opened 2016-01-06
WIP: Allow to build git sources and use their nupks

forki:
Paket v3 alpha versions already support git repos as dependencies (see http://fsprojects.github.io/Paket/git-dependencies.html), but we should go one step further and allow to build the referenced repos. After the build we can use whatever NuGet package the build produced.

The paket.dependencies file might look like:

    nuget Newtonsoft.Json
    nuget FSharp.Core

    group Dev

        git https://github.com/fsprojects/Paket.git master build:"build.cmd NuGet", Packages: /temp/ OS: windows

        nuget Paket.Core

- [x] parse build command and nupkgs output folder from git source feed definition
- [ ] allow to use SemVer git tags
- [ ] restrict checkouts to the specified OS (if OS was specified) see up-for-grabs: #1355
- [x] execute the build command when on checkout of the git source feed
- [x] use output folder als NuGet source feed
- [x] docs

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,6 @@
-#### 3.0.0-alpha012 - 08.01.2016
+#### 3.0.0-alpha013 - 08.01.2016
 * Allow to reference git repositories - http://fsprojects.github.io/Paket/git-dependencies.html
+* Allow to run build commands on git repositories - http://fsprojects.github.io/Paket/git-dependencies.html#Running-a-build-in-git-repositories
 * Allow to use git repositories as NuGet source - http://fsprojects.github.io/Paket/git-dependencies.html#Using-Git-repositories-as-NuGet-source
 
 #### 2.41.1 - 08.01.2016
--- a/docs/content/git-dependencies.md
+++ b/docs/content/git-dependencies.md
@@ -24,6 +24,14 @@ If you want to restrict Paket to a special branch, tag or a concrete commit then
     git http://github.com/forki/AskMe.git 97ee5ae7074b // short hash
     git https://github.com/forki/FsUnit.git 1.0        // tag
 
+## Running a build in git repositories
+
+If your referenced git repository contains a build script then Paket can excute this scipt after restore:
+
+    git https://github.com/forki/nupkgtest.git build build:"build.cmd"
+
+This allows you to excute arbitray commands after restore.
+
 ## Using Git repositories as NuGet source
 
 If you have NuGet packages inside a git repository you can easily use the repository as a NuGet source from the [`paket.dependencies` file](dependencies-file.html):
@@ -42,3 +50,9 @@ The generated [`paket.lock` file](lock-file.html) will look like this:
       remote: https://github.com/forki/nupkgtest.git
       specs:
          (05366e390e7552a569f3f328a0f3094249f3b93b)
+
+It's also possible to [run build scripts](git-dependencies.html#Running-a-build-in-git-repositories) to create the NuGet packages:
+
+    git https://github.com/forki/nupkgtest.git build build:"build.cmd", Packages: /source/
+    
+    nuget Argu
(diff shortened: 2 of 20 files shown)


haf:
+1 on picking up packages from a folder!

inosik:
How would we define cross platform builds? For example `build.cmd` vs. `build.sh` or `MSBuild.exe` vs. `xbuild`?

forki:
yeah that is a good question. I assume we need another optional parameter "mono-build" that would be initialized with "build"

haf:
Better to have a flag that delimits the source to a specific kernel, e.g. MinWin or Darwin or Linux.

forki:
@haf could you propose a syntax?

haf:
`git https://github.com/fsprojects/Paket.git master build:"build.cmd, NuGet", Packages: /temp/, OS: windows`

forki:
so we would define multiple lines?

haf:
Yes, that makes it fairly declarative.

Krzysztof-Cieslak:
What if dependency doesn't generate nupkgs but just dlls (or anything else) ? Maybe it would be possible to add more flexible solution? While I don't care too much and like opinionated tooling it might be problem for some people.

forki:
@Krzysztof-Cieslak this will be possible with something like that:

   git https://github.com/fsprojects/Paket.git master build:build.cmd, OS: windows
   http file://.../paket-files/github.com/Paket/.../myDll.dll

but we might want to make this second line look nicer.

baronfel:
maybe some sort of output spec for the git repo: 
`git https://github.com/fsprojects/Paket.git master build:'build.cmd' OS:windows output: (Library <PATH TO LIB> | Packages <PATH TO PKG DIR>)` 
`git https://github.com/fsprojects/Paket.git master build:'build.cmd' OS:windows output: Library /bin/lib.dll` 
`git https://github.com/fsprojects/Paket.git master build:'build.cmd' OS:windows output: Packages /pkg/`

inosik:
What if we don't build the repo at all and just let the user decide what to do? You want to make a project reference? Okay. Want to build it yourself and just include the output as content? Cool. 

We should remember that external repositories might be anything, not just .NET projects but native libraries, JS frameworks or just icons as well.

Krzysztof-Cieslak:
Well, build command is optional - you can do this with current alpha git implementation:
```
git https://github.com/JS_CONTENT_USER/FANCY_JS_LIB.git master 
http file://.../paket-files/github.com/FANCY_JS_LIB/.../FANCY_JS_LIB.js
```

forki:
yes build command is optional.

inosik:
Yes, I already tried the new feature, but even if building is optional, we still need to care about different OSes, different build tools, output directories and so on.

forki:
anyone interested in implementing the OS filter? 
something along the lines:

    let isMatchingOperatingSystem (operatingSystemFilter:string option) : bool = ...

forki:
Progress: http://fsprojects.github.io/Paket/git-dependencies.html#Using-Git-repositories-as-NuGet-source is now in alpha version avaliable

forki:
so we have a first version that allows to run builds and can be used as NuGet source

theimowski:
how about referencing local git repositories - something between those lines:
``git ..\referenced-repo master``
would this be valuable e.g. when you want to test newer version of referenced component even before pushing it to a remote git repo? 

or maybe specify somehow an optional local path to cloned repository in question:
``git https://github.com/fsprojects/Paket.git master build:build.cmd, OS: windows, local: ..\referenced-repo``

one obvious caveat is the relative path, which everyone would have to use.
Another option would be to restrict the local path according to a specific convention, e.g. the directory resides one level up under name of the repo (the default behavior of ``git clone``)

forki:
you can reference local repos:

    git file:///C:\Users\Steffen\AskMe master

theimowski:
Ah ok, I somehow missed that. Can you use relative path rather than absolute? Would there be any value in associating such local repo with its remote destination?

PR #1357, opened 2016-01-08
NuGetV2: avoid revealing password also if more than one source is defined

cdrnet:
Minor regression on the fix from #1238 to protect personal passwords from being revealed in logs and the shell. In case of multiple sources it fell back to the original unprocessed `sources` and thus bypass the `ToString` override.


forki:
ouch. thx

cdrnet:
Great, thanks!

Vilmir:
Thanks for solving this! We are then going to adopt Paket in our company :+1:

PR #1360, opened 2016-01-08
[V3] Remove lock-dependencies functionality

inosik:
I would propose to remove the `lock-dependencies` functionality from Version 3 of Paket. I realized that this doesn't make much sense in it's current form and can be achieved with a custom `paket.template` while being much more fine grained.

If this is going to be accepted, I would prepare a PR for removing the complete section about this in the documentation.

### DIFF
--- a/docs/content/commands/pack.md
+++ b/docs/content/commands/pack.md
@@ -31,6 +31,11 @@ Or, you could run:
     [lang=batch]
     paket pack output nugets version 1.0.0 lock-dependencies
 
+<blockquote>
+Note that the `lock-dependencies` parameter will be removed in Version 3 of Paket.
+</blockqoute>
+
+
 Depending on which command you issue, Paket creates different version requirements of the packages you depend on in the resulting `.nuspec` file of your package:
 
 <table>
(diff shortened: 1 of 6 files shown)


forki:
I think couple of people are using this.

inosik:
I wasn't sure about this, because for me, it seems like a rather bad idea to have such restrictive constraints of my dependencies. But in case I want to, I can achieve the same with a `paket.template` file and `LOCKEDVERSION`. 

I'll keep this PR open, just in case you change your mind before release of v3.

pms1969:
@inosik  there is now the minimum-from-lock-file switch, which may suit your purposes better.

inosik:
@pms1969 Actually I only wanted to reduce the bloat of the pack function, since you can achieve the same thing with `paket.template`s.

I wrote the `lock-dependencies` functionality but never actually needed it. So I'm cleaning my own mess here 😄

pms1969:
@inosik Ahh, I understand now.  I do agree with you, pinning every dependency to a specific version wouldn't work in almost all cases.  Upgrading becomes a serious hassle if you have more than one of those components, as you're bound to get conflicts.

tsibelman:
I personaly don't see any  scenario where minimum-from-lock-file is not a best option, maybe it need to be default ?

forki:
Yes we could make it default for v3. Would accept pull request s.

Issue #1355, opened 2016-01-07
Implement operating system detection


forki:
For https://github.com/fsprojects/Paket/pull/1353 in Paket v3 we need a OS filter function along the lines:

    let isMatchingOperatingSystem (operatingSystemFilter:string option) : bool = ...

to make it interesting: the syntax for the filter string is still up for dicussion. So we need ideas here.

Volunteers?

kerams:
Just to make sure I understand this correctly, the function is to decide whether the OS we're running on matches an OS specified in the filter or returns `true` if input is `None`, correct?

What are the exact OSes that need to be distinguished?
Is it necessary to tell apart different versions of the same OS?
Can the input string specify multiple OSes?

forki:
Yes you understand correctly.
The issue is I don't really know yet.
Maybe we make it a bit easier and say string -> bool. This way we can at
least defer the decision if we want to match multiple OSes

PR #1361, opened 2016-01-08
Implement isMatchingOperatingSystem

kerams:
Thoughts on this? I decided to force the caller to provide the OS in order to make the function easily testable.
#1355

### DIFF
--- a/src/Paket.Core/Utils.fs
+++ b/src/Paket.Core/Utils.fs
@@ -126,6 +126,21 @@ let extractPath infix (fileName : string) : string option=
     else 
         Some(path.Substring(startPos + infix.Length + 1, endPos - startPos - infix.Length - 1))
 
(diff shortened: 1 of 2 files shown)


forki:
This looks like a good start. Now we need the part that detects the OS. We already have a isMono function, but we need more...

forki:
please take a look at https://github.com/fsprojects/Paket/pull/1366

kerams:
I thought that `Environment.OSVersion.Platform` would be enough.

forki:
yes that's what I used now.

forki:
thanks a lot!

kerams:
It might not be that simple after all :). http://stackoverflow.com/questions/10138040/how-to-detect-properly-windows-linux-mac-operating-systems

forki:
Please take a look at the referenced pull request. We already have a couple
of platform helpers there. Maybe you can make use of these as well.

PR #1369, opened 2016-01-10
Improve OS detection and matching

kerams:
It's annoying that you have to jump through hoops just because this is messed up in .NET/Mono.


forki:
:+1:

PR #1378, opened 2016-01-12
Fix for SolutoinDir in ProjectReference element

ErwinSteffens:
Proposal for fix for resolving $(SolutionDir) macro/variable in ProjectReference include attribute.
When proposed solution is accepted I will add some tests before merging.


ErwinSteffens:
fixes #1378

src/Paket.Core/Utils.fs
```diff
@@ -311,6 +311,8 @@ let askYesNo question =
 
 let inline normalizePath(path:string) = path.Replace("\\",Path.DirectorySeparatorChar.ToString()).Replace("/",Path.DirectorySeparatorChar.ToString()).TrimEnd(Path.DirectorySeparatorChar)
 
+let inline expandVariables(path:string) = path.Replace("$(SolutionDir)", Environment.CurrentDirectory + Path.DirectorySeparatorChar.ToString())
```
forki: This is usully only the Project dir

PR #1385, opened 2016-01-13
Functional Refactoring

cloudRoutine:
- added maybe builder
- cleaned up a few functions
- refactored record methods into module functions for pipelining


build.fsx
```diff
@@ -165,6 +165,21 @@ Target "RunTests" (fun _ ->
             OutputFile = "TestResults.xml" })
 )
 
+Target "QuickTest" (fun _ ->
+
+    !! "C:\Users\Jared\Github\Forks\Paket\src\Paket.Core\Paket.Core.fsproj"
```
Krzysztof-Cieslak: Hard-coded path? ;)

cloudRoutine: hahaha whoops, nice catch

forki:
why did you add the maybe builder? it's not used

cloudRoutine:
```fsharp

        let parseCondition (data:System.Text.StringBuilder) (input:string) index =
            maybe{
                if input.Length <= index then return! None else 
                data.Clear () |> ignore
                let! left, index = parseWord data input index false
                data.Clear () |> ignore
                let! comp, index = parseComparison data input index
                data.Clear () |> ignore
                let! right, index = parseWord data input index false 
                return (left, comp, right, index)
            }
```
https://github.com/cloudRoutine/Paket/blob/polish/src/Paket.Core/ProjectFile.fs#L298-L308

forki:
it's not in this diff right? And is it really woth it?

cloudRoutine:
It is in it, for some reason it's getting collapsed/hidden in the web UI. Honestly I thought I'd be using it more than I did.

forki:
so the question is if we really want to bring that in. TBH I don't think t that big benefit.

cloudRoutine:
this is what it was before

forki:
how about Option.map? that's already avalaible

cloudRoutine:
Is there anything else I need to address?

build.fsx
```diff
@@ -165,6 +165,21 @@ Target "RunTests" (fun _ ->
...
+
+    !! "src\Paket.Core\Paket.Core.fsproj"    
+    |> MSBuildRelease "" "Rebuild"
+    |> ignore
+
+    !! testAssemblies
+    |> NUnit (fun p ->
+        { p with
+            DisableShadowCopy = true
+            ToolPath = "packages/test/NUnit.Runners.Net4/tools"
```
forki: I think you can remove that line. the tool should be discovered

tests/Paket.Bootstrapper.Tests/Paket.Bootstrapper.Tests.csproj
```diff
@@ -58,4 +58,7 @@
       <Paket>True</Paket>
     </Reference>
   </ItemGroup>
-</Project>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
forki: what's this?

cloudRoutine: it gets added every time I run the `RunTests` target

forki:
thanks for helping

forki:
oups. what's that tests/Playlists/paket.tests.playlist thing?

cloudRoutine:
It's for the visual studio test runner so people can run just the tests for paket.core easily

forki:
btw: we're still using F# 3 compiler. 

    normalizePath lib |> FileInfo

is not working on my machine.

forki:
but don't worry. easy to fix for me

PR #1390, opened 2016-01-14
fix for projects that contained linked files.

pms1969:
some project link files were causing odd behaviour.  now fixed.


pms1969:
Sorry, just to be clear, this is a fix for the symbol source package creation.

forki:
thx

PR #1402, opened 2016-01-17
fix symbol sources creation on projects that contain linked files

pms1969:
included a small addition to the documentation that identifies include-reference-projects


pms1969:
not sure what's wrong with the travis build.  those files are definitely commited.  couldn't find a way to kick off the build again either.

Issue #1394, opened 2016-01-15
"Paket pack symbols" also puts the symbols in the non-symbol file


franknarf8:
Hello,

I noticed the 'symbols' flag of the 'pack' command changed behavior recently (it was either created lately or used to only output a single file, I can't recall properly). Anyway, I used to add my sources files manually in the paket.template file and I'm glad this new feature is now online.

I started using 'pack symbols' and I noticed that the .pdb files were also included in the non-symbol.nupkg file.

See repro example : https://github.com/franknarf8/paket-issue-repro/tree/4f8a1ff74bebc37c1dd99270bdaa74db37349c20

A.3.0.0-release.nupkg -> contains A.pdb
A.3.0.0-release.symbols.nupkg -> contains A.pdb + *.cs

Is it intended behavior that the debug symbols (.pdb) are also added to A.3.0.0-release.nupkg?

pms1969:
prior to those changes, the pdb's existed for the project that that the template was for.  This is different to nuget.  I didn't want to break existing behaviour, so for reference projects, I just made sure that the pdb's were also included, as I thought that would be the right thing to do in that situation.  If strict nuget compatability is sought, I think that the referenced project and associated pdb's shouldn't be included unless the command line include-referenced-projects option is set.

personally, I don't mind the PDB's being present, but it's not what nuget does.

Steffen...  thoughts?

tsibelman:
I opened a bug for this issue, I saw that then pdbs are present in main nuget it impossible to get sources from symbols nuget, looks like need to fixed.

pms1969:
Hi @tsibelman, are you saying that because the pdb's are present in the main nuget package, VS fails to download the source package?

tsibelman:
Yes it what I encountered during my testing.

tsibelman:
Here is a duplicate bug #1370

pms1969:
@tsibelman should now be fixed as of 2.45.0

franknarf8:
Everything works works fine now!
Some paket users may not notice the change though and expect the pdb to be in the main nupkg.

tsibelman:
@pms1969 thank you

pms1969:
@franknarf8, the way I worked it, was this:  if you did not request symbols, pdbs are added if your template file has the switch "include-pdbs true"; if you did request symbols, the normal package will not include pdb's of the project file regardless of whether you have "include-pdbs" set or not, and the symols package will always include the pdbs.

[...]

PR #1403, opened 2016-01-18
pdb's no longer being packaged in the non symbols package

pms1969:


forki:
Why that change?

pms1969:
https://github.com/fsprojects/Paket/issues/1394

basically, if the pdb's are in the non-symbols package, VS won't download the symbols package for debugging.

pms1969:
as far as I can tell, the Travis build was already borking on that test.

forki:
So now we put the pdbs in the normal package, but not into the symbols
package?

pms1969:
Other way around.

forki:
But not everyone is using symbols packages. What about the other case?

pms1969:
depends if you want strict nuget replication in that area or not.  Nuget doesn't package them in the non-symbols regardless.

Should only require a small tweak to fix for when there is no symbols requested.  Depends on what you'd like to do?

forki:
I think we should still put the pdbs in the normal package when we don't
create a symbols package. That's what nuget does, right?

pms1969:
No it doesn't, tho i'll confirm that.

In the case that you want to add them, you can still add it as a file dependency and it won't be removed. (I've just confirmed that that is true of that build)

pms1969:
Confirmed. Nuget does not include pdb's regardless of whether symbols have been requested or not.

forki:
Ok. Let's remove them then. But how about a template file property that
automatically adds them? Should be a good alternative for people so they
don't have to use explicit file references

pms1969:
not a bad idea?  so for project files only?
property: include-pdbs
default: false
??

forki:
yes that would be awesome

pms1969:
do you want that as part of this pull request or another?

forki:
please add it to this one.

pms1969:
done

pms1969:
travis failing on the same tests.  It has an ignore on it, so not sure why it gets run.

Issue #1341, opened 2016-01-05
Add a http file reference as project reference


jruizaranguren:
I have added a `dll` library downloaded as an `http` reference within a group. The `paket.dependencies`  file contains:

```
group Files
http file:///absolute/path/to/library.dll target/path/library.dll
```

Then in the `paket.references` file I use:

```
group Files
	File:library.dll
```

This is adding a content node in the references `ItemGroup` in the `.csproj`, such as:

```xml
    <Content Include="..\..\paket-files\files\localhost\target\path\library.dll">
      <Paket>True</Paket>
      <Link>paket-files/library.dll</Link>
    </Content>
    ...
    <Reference Include="System" />
    ...
```

This is just a link to the source file which I can not use to create a project reference.

Is there any way to add this file as a project reference?

forki:
not yet. but I think it's not that hard to do.
let me see if I can hack something together

jruizaranguren:
Thanks. If you give me some clues I might try to make the enhancement.

forki:
it would be usefule if you could generate a minimal repro that uses a sample dll (doesn't need to compile - dll might be 0 bytes)

jruizaranguren:
As a couple of tests or a folder with all required files (source file, solution file, project file)?

forki:
yes a folder with all required files would be nice. I would then put this into our integration test suite

jruizaranguren:
Here you have. http path must be changed!.

[i001341.zip](https://github.com/fsprojects/Paket/files/78340/i001341.zip)

forki:
please test the latest version.
I added your sample repo as integration test and implemented a fix. but Im not completely sure if it works. so please see it as an alpha version that needs further testing.

jruizaranguren:
Starting with a greenfield folder, it succesfully creates a project reference to the dll. But there are some undesirable behaviors:

* It copies the dll to a new local folder below the project folder: `paket-files/library.dll`. This is not needed because the library is already downloaded below the `paket-files` directory. Only the project reference should be created.
* When the reference is deleted in `paket.references`, it succesfully deletes the reference, but the mentioned new local file `paket-files/library.dll` is not deleted.
* When `paket-files/library.dll` is found below the project directory, the installation fails. So forced updates of the library will fail.
* If we already had the library referenced as a link in a `content` item, this is not deleted when the new reference is added (minor issue to be documented).

forki:
[...]

PR #1406, opened 2016-01-18
improve http-dependencies documentation #1341

jruizaranguren:
Try to improve documentation as requested in https://github.com/fsprojects/Paket/issues/1341

### DIFF
--- a/docs/content/commands/update.md
+++ b/docs/content/commands/update.md
@@ -65,3 +65,10 @@ If you want to update a single group you can use the following command:
   `--redirects`: Creates binding redirects for the NuGet packages.
 
   `--no-install`: Skips paket install --hard process afterward generation of [`paket.lock` file](lock-file.html).
+
+## Updating http dependencies
+
+The update command has no effect on http dependencies. If you want to update a file you need to use the [`paket install` command](paket-install.html) with `--force [-f]` option.
+
+Using [groups](groups.html) for http dependent files can be helpful in order to reduce the number of files that are re-installed.
+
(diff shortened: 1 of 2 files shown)


docs/content/commands/update.md
```diff
@@ -65,3 +65,10 @@ If you want to update a single group you can use the following command:
   `--redirects`: Creates binding redirects for the NuGet packages.
 
   `--no-install`: Skips paket install --hard process afterward generation of [`paket.lock` file](lock-file.html).
+
+## Updating http dependencies
+
+The update command has no effect on http dependencies. If you want to update a file you need to use the [`paket install` command](paket-install.html) with `--force [-f]` option.
```
forki: If that's true then this is clearly a bug. @sergey-tihon is that correct? I mean you use https://github.com/sergey-tihon/Stanford.NLP.NET/blob/master/paket.dependencies#L13

jruizaranguren: I do not see it as a bug. It is not straightforward to determine if a file is outdated through http unless you allways download it and store a hash of it. You would need a system based on http headers and I doubt it could be universally reliable. 
It would be a bit overkill to download every file to test for updates. I find it a good trade-off to be aware of that and need to force the install.

forki: If "Update" doesn't download but "install" does, then I classify this as bug.

forki: can you check if "update -f" works?

jruizaranguren: Yes it does!!! :smile:  Wrong docs :disappointed:

sergey-tihon: Sorry, always did it through `paket install`. When I manually edit `paket.dependencies` and then execute `paket install` - i feel that everything is under control ;)

jruizaranguren:
Minimum change that might work.

forki:
awesome. thx

PR #1407, opened 2016-01-18
Find packages is now case-insensitive #93

dcorriveau-omniscient:
I am F# virgin so I'm not sure if the code is "optimal", but it seems to work. I've created a unit test to poke our private nuget server and with my fix it does return all packages that contain the words, case insensitive.

### DIFF
--- a/integrationtests/Paket.IntegrationTests/paket.references
+++ b/integrationtests/Paket.IntegrationTests/paket.references
@@ -6,3 +6,4 @@ group Test
   NUnit
   NUnit.Runners.Net4
   File:FsUnit.fs .
+  FAKE
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -22,6 +22,7 @@ group Test
 
   source https://nuget.org/api/v2
 
+  nuget FAKE
   nuget NUnit.Runners.Net4
   nuget NUnit ~> 2
   github forki/FsUnit FsUnit.fs
--- a/paket.lock
+++ b/paket.lock
@@ -42,6 +42,7 @@ GROUP Test
 NUGET
   remote: https://www.nuget.org/api/v2
   specs:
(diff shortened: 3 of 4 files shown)


forki:
The changes look the same like in the other pr?

dcorriveau-omniscient:
Sorry, I messed up the PR (I don't do that often so I wasn't sure how to do it). I'll create another branch and do another PR on that branch with only the good changes.
Originally I only had the right change for the case-sensitive, but since I did the PR on my master branch and I changed other stuff since then I guess those changes are also included in the PR even if they are done after the original PR... ?

forki:
Yes. Always use a new branch for PRs. You should never use master (that
breaks things when you have more than one).

forki:
And thanks for contributing!
On Jan 18, 2016 21:05, sforkmann@gmail.com wrote:

Yes. Always use a new branch for PRs. You should never use master (that
breaks things when you have more than one).
On Jan 18, 2016 21:04, "dcorriveau-omniscient" <notifications@github.com>
wrote:

Sorry, I messed up the PR (I don't do that often so I wasn't sure how to do
it). I'll create another branch and do another PR on that branch with only
the good changes.
Originally I only had the right change for the case-sensitive, but since I
did the PR on my master branch and I changed other stuff since then I guess
those changes are also included in the PR even if they are done after the
original PR... ?

—
Reply to this email directly or view it on GitHub
<https://github.com/fsprojects/Paket/pull/1407#issuecomment-172639083>.

dcorriveau-omniscient:
Looks like you were able to pick partial commits anyway ? So I won't have to make another pull request using another branch (for this time) ?

forki:
I am not sure what happened. I merged the other pr. 
Please send new pr with the actual fix

PR #1408, opened 2016-01-18
Fake references missing for IntegrationTests

dcorriveau-omniscient:
Not sure if I'm doing something wrong but I didn't have FAKE references and so I couldn't compile the project.

### DIFF
--- a/integrationtests/Paket.IntegrationTests/paket.references
+++ b/integrationtests/Paket.IntegrationTests/paket.references
@@ -6,4 +6,3 @@ group Test
   NUnit
   NUnit.Runners.Net4
   File:FsUnit.fs .
-  FAKE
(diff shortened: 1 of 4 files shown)


forki:
What do you mean with "fake references"?

forki:
Ah I understand. But please take the existing fake from the build group. No changes in dependencies or lock file needed. Only in the paket.references

dcorriveau-omniscient:
FakeLib doesn't work for the IntegrationTests project (exclamation mark next to the reference and I can't compile).
Edit: Alright, I'll try the paket.references only

dcorriveau-omniscient:
Alright, it should work with the Build group now.

forki:
Lgtm. Thx

PR #1410, opened 2016-01-18
Case insensitive find packages for NugetV2

dcorriveau-omniscient:
I think I figured how to do pull requests :P


forki:
Ah yes. That makes sense. Thx

PR #1417, opened 2016-01-20
fixed pack package dependencies for dependent projects

pms1969:
changed the doco around include-referenced-projects
fixed dependency resolution for project dependencies


forki:
does it still exclude stuff that has it's own template file?

pms1969:
no.  I'll add that in now.

pms1969:
Just noticed a problem with ProjectFile.findCorrespondingFile

I think the indentation for findInDir is all wrong, and as a result, it never does that bit of searching.

I can fix that, but is the purpose of that function to find the paket.template and paket.references files for the project?  If so, should it actually be tree walking up the directory stack to find them?  I would imagine that they should be present in the same directory only to be considered the associated file.

pms1969:
Working in the following way.  

* Projects are recursively traversed. 
* Any projects with their own paket.template are added as dependencies
* Any other projects have their list of dependencies added.

I'd like to do a bit more testing before hitting go on this, but need to be able to sync up with my laptop.

forki:
cool. let me know when you are ready

pms1969:
should be ready.

You may want to consider putting this into the 3.0 branch, as there's a breaking change
that change being that dependent project dll's are not added to the package now unless the switch included-dependent-projects is set.

What it does now do tho, is walk the project dependency tree recursively if that switch is set.  It does it for dependencies, sources and dll's, which imho is an improvement of nuget, as it only goes one level down.

forki:
> that dependent project dll's are not added to the package now unless the switch included-dependent-projects is set

but we never wanted to do this anyway, right? I mean we epected that this did not happen. Or am I wrong here?

pms1969:
Direct project dlls had already been added, so my previous commit didn't change that behaviour. But you are correct, the behaviour in that commit is what we want.

pms1969:
Sorry, this commit, is the correct behaviour.

pms1969:
No edit on mobile. :-(

forki:
thx

forki:
can youn please send a PR for docs update? Thx

pms1969:
the original commit had some document changes.  Tho I'll add some extra wording to make sure it's clear that it's recursive, and that projects with their own project template files are ignored.

forki:
Do we have the parameter documented? Where exactly?

pms1969:
right at the very end.
http://fsprojects.github.io/Paket/paket-pack.html

pms1969:
hmm, can't find that in the solution any more.

Issue #1414, opened 2016-01-19
Question about 'pack'


dcorriveau-omniscient:
I'm trying to create Nuget packages using Paket and I'm having a few issues. For one project, when I use a template and set it to "project", the output assembly isn't added to the package (maybe I should look at the code to see how it infers what should be the output assembly).

So since it didn't work for me I thought I'd switch to a "file" template. I managed to make it work by specifying the assembly and the info but I'm wondering why can't paket infer the basic info from the AssemblyInfo, just like it does for a "project" template ? For example, why do I need to specify the id, version, etc. ? I'd like to be able to specify the files I want in lib, but I would like paket to use default values from AssemblyInfo if they aren't declared. I'm guessing it's a design choice but I'm not sure why it's working that way ? Also, is there something special about the "copyright" field, since title, id, company, etc. can be inferred from the AssemblyInfo but "AssemblyCopyright" isn't used for the Copyright info ?

Thanks.

forki:
Inference happens only in type project.

dcorriveau-omniscient:
I know, but I'd like to understand the design choice behind that, since that suggests my workflow might not be right.

forki:
You can pack everytthing into the file type. We cannot know what stuff. Just look at packages like FAKE or the Paket package itself. We only have tools in these - no libs. They might not even be .NET libs.

Even if we have .NET libs in it - which one would we take? 

It just no made for that.

forki:
The interesting question is: why does project type not work. Maybe you ran into a bug.

dcorriveau-omniscient:
Ok so your main usage for a file type are packages that contain misc stuff, that aren't .net projects most of the time. That makes sense.

I'll have to debug the project type then, as it seems that's what would be best suited for me.

forki:
Yes. Also content packages that do not contain any binary data at all

pms1969:
For me, project types only really work when the paket.template file is named after the csproj.  I.e. I have a project called a.csproj with an a.csproj.paket.template file.

Might also just be my useage.  Also, you cannot place the template name on the command line.  You must call paket either from a super directory with no template specified, or from the directory containing the csproj file, also without specifying the template name.

dcorriveau-omniscient:
Not sure if I fully understand the templatefile argument. I thought it was used to specify a template file so that it would limit the packaging to that project only (the one that corresponds to the template file).

[...]

PR #1418, opened 2016-01-20
Added pack template test

dcorriveau-omniscient:
Related to https://github.com/fsprojects/Paket/issues/1414#issuecomment-173169406
The test is creating a Nuget package and as you can see if you unzip it, it doesn't include the output assembly (my test isn't very accurate because it's checking the file size but you get the point).


integrationtests/scenarios/i001376-pack-template/before/packages/MySql.Data/MySql.Data.nuspec
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2011/08/nuspec.xsd">
```
forki: scenarios/i001376-pack-template/before/packages/ should be excluded, right?

dcorriveau-omniscient: Yea.

forki:
can yuou please rebase / squash that. so that we loose the history of that packages thingy

dcorriveau-omniscient:
I tried something, let me know if the history is better.

forki:
nope ;-) as you can see here it's now 2 commits more

forki:
https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

forki:
I assume you forgot to force push.

If you pull again and push, then you merge the old stuff in

dcorriveau-omniscient:
Sorry, my doppelganger took control of my account and kept adding commits. Fortunately I managed to fix things up :P

forki:
ha! look much better now :+1:

forki:
fixed in https://github.com/fsprojects/Paket/pull/1420
thanks for creating a repro case

PR #1424, opened 2016-01-22
What files should be committed or ignored

citizenmatt:
Added an FAQ to make it clear what files should be committed or ignored from source control.

Complete newbie at paket, so not sure I've got it right :)

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -65,6 +65,22 @@ There is no reasonable way to make this work – and even NuGet.exe can't do it 
 
 Instead we encourage the .NET community to use a declarative install process and we will help to fix this in the affected packages.
 
+## What files should I commit?
+
+Paket creates a number of files in your solution folders, and most of them should be committed to source control. To be clear, these are the files that should be committed to source control:
+
+* [`paket.dependencies`](dependencies-file.html) - specifies your application's dependencies, and how they should be fulfilled.
+* All [`paket.references`](references-files.html) files - each project will have a `paket.references` file that specifies which of the dependencies are installed in the project. Each of these files should be committed to source control.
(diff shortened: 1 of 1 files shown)


docs/content/faq.md
```diff
@@ -65,6 +65,22 @@ There is no reasonable way to make this work – and even NuGet.exe can't do it
...
+
+* [`paket.dependencies`](dependencies-file.html) - specifies your application's dependencies, and how they should be fulfilled.
+* All [`paket.references`](references-files.html) files - each project will have a `paket.references` file that specifies which of the dependencies are installed in the project. Each of these files should be committed to source control.
+* [`paket.lock`](lock-file.html) - records the actual versions used during resolution. If it exists, Paket will ensure that the same versions are used when restoring packages. It is not strictly necessary to commit this file, but strongly recommended (see the following question for details why).
+
+The following files can be committed, but are not essential:
+
+* [`.paket/paket.bootstrapper.exe`](getting-started.html) - this is a small, rarely updated executable that will download the latest version of the main `paket.exe`. If you prefer to place `paket.exe` in the path, or use a [Paket plugin for your editor](editor-support.html), then you don't need to commit this file. However, not committing it also requires that other users of your solution have `paket.exe` available.
```
forki: others and build servers too

forki:
Awesome!

docs/content/faq.md
citizenmatt: I didn't see any docs about how build servers would use it - is it just an assumption for scripts?

forki: Many users have a build.sh or buld.cmd with:

    .paket/paket.bootstrapper.exe
    .paket/paket.exe restore
    *run whatever build tool*

paket itself has this as well

forki:
Very cool. I will also add some words about Paket.targets which allows you to restore in VS  without external build script.

forki:
Thanks so much

PR #1426, opened 2016-01-23
If Paket bootstrapper fails, suggest common fix

rmunn:
One of the most common reasons for the Paket bootstrapper to fail on Linux is if the Mono installation is brand-new -- because out of the box, Mono does not trust *any* X.509 root certificates. Therefore, GitHub's SSL certificate won't be trusted and Paket won't be downloaded.

The fix most often recommended is to run:

    mozroots --import --sync

which imports the entire root SSL certificate store trusted by the Mozilla project. This is over 100 certificates, so not everyone will want to run this command. (And we certainly shouldn't be running it for them.) But for someone who is new to F# and Mono, this suggestion might save them hours of frustrated Google searches.

Note that we try to only print this message if it looks like it is actually true for the user's situation, by using Mono's certificate manager to count the number of trusted X.509 certificates. We compare to 1 rather than 0 because the word "X.509" will appear once in the header of certmgr's output even on a system with no trusted root certs.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -21,6 +21,18 @@ else
   mono .paket/paket.bootstrapper.exe prerelease
   exit_code=$?
   if [ $exit_code -ne 0 ]; then
+    certificate_count=$(certmgr -list -c Trust | grep X.509 | wc -l)
+    if [ $certificate_count -le 1 ]; then
+      echo "Couldn't download Paket. This might be because your Mono installation"
+      echo "doesn't have the right SSL root certificates installed. One way"
+      echo "to fix this would be to download the list of SSL root certificates"
+      echo "from the Mozilla project by running the following command:"
+      echo ""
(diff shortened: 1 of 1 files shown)


forki:
OK this is really nice, but it will only help Paket contributors. We should put this directly into the bootstrapper if possible.

Or at least also into https://github.com/fsprojects/ProjectScaffold/blob/master/build.sh

rmunn:
Yes, that makes a lot of sense. I figured the `build.sh` files I was seeing in other projects were copied from Paket. But if they were copied from the ProjectScaffold template, then that's the best place to put this. I'm of two minds about putting it into the bootstrapper; if it doesn't greatly increase the bootstrapper's compiled size, then that makes sense. But if it makes `paket.bootstrapper.exe` even 2.5% larger, that's probably too much and the check should go into `build.sh` instead.

Issue #1434, opened 2016-01-26
paket pack with enabled include-referenced-projects flag fails with NRE


ar3cka:
Find the build log here https://ci.appveyor.com/project/ar3cka/journalist

forki:
I can reproduce

forki:
@pms1969 I fixed it in https://github.com/fsprojects/Paket/pull/1435, but I'm not completely sure why that was in there. Any ideas?

forki:
@ar3cka I released a new paket prerelease which allows you to test if the behaviour is correct. I mean the NRE should be gone, but you should look if the packages are built correctly.

pms1969:
ok, I'll check it out.

pms1969:
looks good.

ar3cka:
I've checked. Package contains all assemblies. Thx.

PR #1435, opened 2016-01-27
Fix NRE in #1434

forki:


pms1969:
that bit of code was to ensure that referenced project were included as dependencies in the package, not as dll's, if, and only if the project had it's own template that happened to be a project template.

pms1969:
was it barfing on core.Id.Value?

forki:
But why was that recursive? And we do add nuget dependencies at another
place, right?

pms1969:
was recursive because project a can reference project b which can reference project c.  project c has a paket.template, and you'd need that to be included as a package dependency

forki:
Mhm I guess that makes some sense for this setting. Yes the package id is
null.

pms1969:
The dependencies that are added elsewhere are only direct dependencies, they don't include dependencies for projects that become packages themselves

forki:
how about 3b3a73d then?

pms1969:
looks good

forki:
I don understand "~~referenced-project" - where is that handled?

pms1969:
yeah, I was unsure that was the right thing to do with that, but I kind of hit the limits of my F#.

It's handled right at the bottom of that file.

forki:
what do you think about  5783de4?

forki:
ah never mind. I think I now understand what you want to do.

forki:
Like  f46872a ?

Issue #1440, opened 2016-01-29
Conditional ItemGroups on .NET version get in the way


jackfoxy:
Not sure if this is entirely Paket's doing. There must be a use case for setting different ItemGroup .NET version conditionals for different versions of a package dll, but in my use case (and I would think this is the most common case) I am only every building whichever .NET version I specified for the project in Visual Studio. At best the conditionals just work and are clutter, but frequently Paket Update does not update them correctly. Most recently when I upgraded projects to .NET v4.6.1 Packet seemed to only know about v4.6 and set condtionals only as high as v4.6. (Yes, I tried setting net461 in the dependencies file.) Needless to say, nothing would build. It's also aggravating when there is only one version of a dll, and conditionals are built up for the only version that can build. (Hope this makes sense, taking a long weekend and might not be able to respond for a couple days.)

forki:
Mhm ww have the framework option that you can set on top of the dependencies file and that should work. Do you have a concrete repro case where it didn't?

forki:
Installing all versions in conditionals is the default since this allows you to change your target framework whenever you like and also allows you to build for multiple target frameworks without reinstalling packages. 
When you set a target framework filter we install only a subset of the possible assemblies. This reduces clutter in the csproj files, but also makes target framework switching harder. The target framework filter even influences the package resolution. 

Setting net461 as you described above should have worked to install the libs for you target framework. That's why I asked for a repro case.

jackfoxy:
All of my problems over the last 2 months, and those of others on our team over many more months have resulted in significant manual rework of projects, so I cannot easily give you a test case. I just don't remember how to set things up. Instead I think I can make some recommendations below.

>>Installing all versions in conditionals is the default since this allows you to change your target framework whenever you like and also allows you to build for multiple target frameworks without reinstalling packages.

Something I have never wanted/needed to do, and none of the solutions we build at our company target multiple frameworks. This is useful for some solutions in the world, but I am sure the overwhelming majority do not need to do this. </commentary>

I got off to a bad start with this issue report. Let me start over:

This is not just my experience, but I think I can speak for our whole team, which is F# focused, almost no C# whatsoever.

1) We are not interested in managing multiple .NET versions in our builds. Ever. If we ever care we can build it in.

2) There are only 3 use cases for Paket to ever modify a project file. This should be the default behavior of Paket.Update. If it cannot be the default behavior, there should be an easy to understand option to make it behave this way.
[...]

PR #1451, opened 2016-02-02
Automatic framework detection - references #1440

forki:
#### Automatic framework detection

If you use only one target framework in your whole repository, then paket can detect this for you and limit the installation to that specific target framework. You can control this in the [`paket.dependencies` file](dependencies-file.html):

    framework: auto-detect

    nuget Example >= 2.0 // only the target framework that is used in projects

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,7 @@
 #### 2.49.0-alpha001 - 03.02.2016
 * Automatic framework detection - http://fsprojects.github.io/Paket/dependencies-file.html#Automatic-framework-detection
-* Work around auth issues with VSTS feed - https://github.com/fsprojects/Paket/issues/1453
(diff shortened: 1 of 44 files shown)


Krzysztof-Cieslak:
From Slack:
> probably it would be nice if it worked also if multiple frameworks are used in solution, giving conditionals for only those frameworks which are defined in any project

forki:
@Krzysztof-Cieslak so we would install the union of all the target frameworks?

Krzysztof-Cieslak:
Yes, so let's say we have in solution one project with 3.5 and second one with 4.0 - for all projects we define conditional references for both 3.5 and 4.0.

forki:
sounds good

forki:
If we don't find a project then we would not restrict the resolution, right?

Krzysztof-Cieslak:
Yes, that sounds reasonable

PR #1448, opened 2016-01-31
added pack switch minimum-from-lock-file 

pms1969:
which adds the minimum version for a packages dependencies from the built against version of the package.

includes integration test.

PR #1452, opened 2016-02-03
added pack switch minimum-from-lock-file 

forki:
replaces #1448


pms1969:
Works a treat.  Thanks.

PR #1465, opened 2016-02-08
Install Paket with a command line like Chocolatey

rflechner:
Hi,

When using Paket, I noticed that it could be repeated to install in new projects.
So I think it could be useful to install it as easily as Chocolatey (https://chocolatey.org/)
I propose to give only one command line who :
 - creates .paket folder
 - downloads paket.bootstrapper.exe and paket.exe
 - run paket.exe init
 - add paket.exe to .hgignore or .gitignore if they are present

If are intersted by this pull request, we must change urls in index.md to target the main repository instead of my fork.
Using Google url shortener, we can reduce size and number of lines to copy.

paket_home2
paket_home

Regards


forki:
We have a choco package. Is that if any help?

cdrnet:
If you really have to do things like this, please at least do *not* use a URL shortener, that's a security nightmare...

wesley david s photo

rflechner:
whoops, sorry.
I didn't know the power shell tool.

I found cool to copy command "curl https://raw.githubusercontent.com/rflechner/Paket/EasyInstall/bootstrappers/scripts/install.sh | /bin/sh" from the website for a first setup (on macos).
But effectively, it could be a bad idea.

I close my pull request :)

Thank you for your time.

paket_install_mac02

forki:
so does the choco tool work for you?

rflechner:
On Windows:

It seems the Chessie dependency is missing.

I used "choco install paket.powershell", but the nupkg (in C:\ProgramData\chocolatey\lib\Paket.PowerShell) doesn't have the dependency.

On MacOs:

May be a homebrew package could be intersting ?

capture-paket

Issue #1392, opened 2016-01-15
StackOverflowException resolving NuGet package without version constraint from large paket.dependencies


viktor-svub:
StackOverflowException resolving "our.proprietary.component"
* Paket.exe 2.44.1, version 2.21 works but does not handle prereleases
* paket.dependencies NuGet, declared without any version constraint
* there are about 180 declared dependencies, all NuGet, all custom feeds
* probably also transitive dependency of some of the other component/s
* SO is thrown in \src\Paket.Core\Domain.fs, line 30, ckeckout cf9c622
  (this.GetCompareString().CompareTo(that.GetCompareString()))
* stack trace @ https://gist.github.com/viktor-svub/c1563f0c2b98562f67ee

forki:
can you please try to find the last working version?

viktor-svub:
OK, the issue appears between 2.40.14 (latest working) and 2.40.15 (throws SO). 
* the full command crashing with exception is `paket.exe update --no-install`
* we have 4 custom NuGet feeds, often containing different versions of same packages
* majority of the 180 immediate dependencies are `prerelease`
* majority of all the dependencies are using `@~>` resolution
* the generated `paket.lock` is about 360 lines long

forki:
diff --git a/src/Paket.Core/SemVer.fs b/src/Paket.Core/SemVer.fs
index 444e688..d466672 100644
--- a/src/Paket.Core/SemVer.fs
+++ b/src/Paket.Core/SemVer.fs
@@ -144,6 +144,8 @@ interface System.IComparable with
                     | None, None -> 0
                     | Some p, None -> -1
                     | None, Some p -> 1
+                    | Some p, _ when p.ToString() = "prerelease" -> -1
+                    | _, Some p when p.ToString() = "prerelease" -> 1
                     | Some left, Some right -> compare left right
 
             | _ -> invalidArg "yobj" "cannot compare values of different types"


that's the only change. weird

forki:
but I assume the real issue is that our resolver algorithm is still recursive. Need to implement BFS qith a queue...

forki:
@viktor-svub in the latest version I inline the tryToImprove function. This will "free" halve of the stack in your sample and should help to process significantly larger graphs. Does this work for you?

@mrinaldi do you see a chance that we can remove the recursive step function and turn it into something that manages a queue or a stack internally?

mrinaldi:
IIRC yes.
[...]

Issue #1432, opened 2016-01-26
Prerelease package range causes stack overflow in PackageResolver


rkachowski:
We've been using a paket based package system internally to manage nuget packages within our organisation. In recent versions we've noticed a recurring error which forces us to pin to a lower version (we are using 2.39.10, but more investigation is needed to check whether more recent versions are stable)

When using the latest (2.47.2 release and also current master d0846c4 ) we get error when attempting to run `paket update`. From a brief look it appears that the error is due to recent changes in the prerelease handling code.

We have two situations in our build, one *stable* version where we wish to pull in only stable releases, and a *prerelease* version, where all dependencies are pulled in on the latest development/prerelease version.

the paket.dependencies file for a prerelease looks similar to
```
source http://private_repo
nuget package_name >= 0.0.0-prerelease
```

In this situation it looks like `getCompatibleVersions` is only returning non prerelease versions and so the resolver gets stuck in a loop when it recurses on [step](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/PackageResolver.fs#L475).

Attached is a sample stack trace
[stacktrace.txt](https://github.com/fsprojects/Paket/files/105376/stacktrace.txt)

forki:
sounds like a duplicate of #1392 - is there a way for me to reproduce?

(something like giving me temporary access?)

rkachowski:
i'll need to get back to you with a reduced test case...

flowsprenger:
This is a small reproducible testcase
https://gist.github.com/flowsprenger/3d1b0dd7f6168b9e31b9

Running setup_packages.sh with Paket 2.39.10 it will yield:
Resolving packages for group Main:
- Paket.Test.D 1.0.0-prerelease
- Paket.Test.B 1.0.0
- Paket.Test.C 1.0.0-prerelease
- Paket.Test.A 1.0.0

Later versions will yield:
```
Paket version 2.50.2.0
Resolving packages for group Main:
 - Paket.Test.D 1.0.0-prerelease
 - Paket.Test.B 1.0.0

Unhandled Exception:
System.StackOverflowException: The requested operation caused a stack overflow.
  at Microsoft.FSharp.Core.LanguagePrimitives+IntrinsicFunctions.TypeTestGeneric[Double] (System.Object source) [0x00000] in <filename unknown>:0
[...]

PR #1470, opened 2016-02-11
Reproduce stackoverflow from #1432

forki:


forki:
Ok the fix in 98faf34 is really worrying me.

So this suggest we do something strange and confuse Set.remove. I looked at the implementation and it's using CompareTo and no Equals.

During the debugging of the integration test I saw a set with 3 elements where we wanted to remove one element (which was also the first in the set). CompareTo was only called twice and not with the first element of the list (at least the debugger didn't step in there).

/cc @dsyme @mexx

forki:
Ok d6e4503 seems to be the correct fix. Some how we had a case where we made a > b and b > a
This confused Set.remove since it is relying on proper CompareTo.

mexx:
@forki `Set` requires the contained type to be ordered, as far as I can see from a quick look on the implementation. Is the underlying object type ordered?

forki:
@mexx yes I fucked up the order as written above.

forki:
And was so close: https://github.com/fsprojects/Paket/issues/1392#issuecomment-171967067

PR #1474, opened 2016-02-15
Readme: Fix invalid link to "http-resource".

zapu:
The link used to lead to `https://github.com/fsprojects/Paket/blob/master/11` because of invalid syntax used.


forki:
Thx

PR #1478, opened 2016-02-17
fix the debugger label for Groupname

baronfel:
tiny change so that debuggerdisplay of groupnames works instead of throwing a 'member not found 'Item'' exception.


forki:
Thanks

PR #1479, opened 2016-02-17
fix group transitive dependency checking

baronfel:
if a group isn't found in the lockfile then its transitive dependencies should be the empty set.


baronfel:
fixes #1476

forki:
Very cool. Thanks

Issue #1477, opened 2016-02-17
Paket holds on to files previously restored (via install)


haraldsteinlechner:
In my build script i use paket's install mechanism in order to install and restore packages. Afterwards i override some of the installed packages with packages, just compiled from source. 
(similar to cabal sandbox add-source)
After some debugging i could not figure out what function is holding on to extracted files.
concretely my pseudocode looks like:

```
let p = Paket.Dependencies.Locate(Environment.CurrentDirectory)
p.Install(false,false)
// here some dll's in packages folder are kept open, thus i get accessdenied exception in continuation.
restoreSourceDependencies()
```

my current version of paket is:  Paket.Core (2.48.2)

before further debugging, do you have any ideas what parts of the install procedures could hold on to file handles?

any ideas?
thanks harald

haraldsteinlechner:
ok it seems to be related to assembly binding redirect stuff.
I have absolutely no clue why this happens (according to msdn docu) but this commit "fixes" this issue for me:
https://github.com/haraldsteinlechner/Paket/commit/09d17f32f02a4ac1cd9dfa1ae56f2c71e2253dc1

I tested the situation separately in fsi.exe and found that indeed System.Reflection.Assembly.ReflectionOnlyLoadFrom holds on to the file after returning. Note that i cannot reproduce the problem with copyShadowAssemblies enabled. 
Now comes the really odd part: i invoke my FAKE script already with shadow assemblies enabled. 

After all, not understanding the problem completely, we could simply load the file as demonstrated in my commit but i don't know what effects this could have.

isaacabraham:
@haraldsteinlechner can you just send the link to the MSDN doc? And (ideally) a failing test for this (if possible and not a lot of work!)

Cheers

haraldsteinlechner:
what i meant with docu is that: https://msdn.microsoft.com/en-us/library/system.reflection.assembly.reflectiononlyloadfrom(v=vs.110).aspx says nothing about holding on to fileHandles

PR #1480, opened 2016-02-17
Prevent paket holding locks on assemblies during binding redirects

haraldsteinlechner:
solves issue https://github.com/fsprojects/Paket/issues/1477

### DIFF
--- a/src/Paket.Core/InstallProcess.fs
+++ b/src/Paket.Core/InstallProcess.fs
@@ -213,7 +213,8 @@ let private applyBindingRedirects (loadedLibs:Dictionary<_,_>) isFirstGroup crea
                             match loadedLibs.TryGetValue key with
                             | true,v -> v
                             | _ -> 
-                                let v = Assembly.ReflectionOnlyLoadFrom library
+                                let ass = File.ReadAllBytes library
+                                let v = Assembly.ReflectionOnlyLoad ass
                                 loadedLibs.Add(key,v)
                                 v
 


forki:
@haraldsteinlechner @isaacabraham is this good to go?

forki:
What do we need from the assembly? We have a assembly meta data reader here in paket that doesn't load the assembly at all.

haraldsteinlechner:
imo my commit would be good to go. 
The only problem i could imagine is performance.
I'd recommend to merge the commit now, but leave the issue open and look for another option instead of loading the assembly.

isaacabraham:
@forki it was to get assembly version and the signed key (if it exists). We never had the metadata reader at the time of the BR code so if that can also get these attributes maybe it would be better to replace this PR with one that uses that instead?

In the interim this code looks like it should be fine (although I haven't tested it myself) - it's just an alternative way of loading data on the assembly. The fact that it has to load the entire assembly into memory first could have some implications in terms of memory though?

forki:
mhm merging this broke couple of releated integration tests. not sure why the build was green.

haraldsteinlechner:
hmmm. crazy. also for my build all worked fine. checking again

forki:
It's 4 integration tests. And if I revert your commit they are working
again.

haraldsteinlechner:
my initial commit was green because it only ran on travis i suppose

haraldsteinlechner:
ok. i failed in finding a good, fast and robust solution to this problem.
1) Loading assemblies in the default AppDomain always locks the file.
2) We could use a fresh appdomain for loading and unloading those assemblies. However, using appdomains to do this is not straightforward (especially since passing data around requires serialization etc)
3) We could use mono.ceceil for loading assembly metadata, but i suppose this is problematic as well (additional dependency)-

I don't know why there is a difference between mono and .net but one thing i know for sure: assembly binding semantcis is subtle and danger zone.

So, fixing our windows build (and appveyor CI), is not that simple. Revert and look for another solution not involving assembly loading (which is also SLOOW)?

isaacabraham:
Also AppDomains are dead in CoreCLR world.

forki:
Ok I will try to use the meta data reader that came from the F# compiler
itself. We already use it for pack.

isaacabraham:
Sounds good. If we can rationalise this stuff to use a single mechanism that's a win.

forki:
Ok tried a different workaround in c57e735 - can please test with latest?

haraldsteinlechner:
i also found a different workaround: https://github.com/haraldsteinlechner/Paket/commit/777a4157023352b8ff37c18cfbd87bbd1305c675

the main problem was the caching of loaded libs

forki:
cool please send as PR

haraldsteinlechner:
i updated my fix https://github.com/haraldsteinlechner/Paket/commit/ce21a161030f3c8dc84654973d5ebfeab483e90f 

However, i did not send a pull request yet since i think your version is more robust when considering same assemblies with different versions etc. (my version builds on computing a key for assembly paths, yours uses more .net assembling binding resoution magic)

haraldsteinlechner:
i finally went with my solution, since i had some weired test errors (laziness in seq). i hope we can close down this boring thing.... (see my new PR, waiting for CI)

Issue #1416, opened 2016-01-20
paket remove does not delete packages content


jruizaranguren:
When a dependency is removed from `paket.dependencies` or alternatively when `paket remove nuget nuget_id` is called, all the previously downloaded files stay in drive.

This content is garbage. It would be great if `remove` get rid of all package content in `packages` or `paket-files`.

ivankoster:
I would actually propose to also implement garbage removal on install / update commands.
If I change the packet.dependencies file and perform an install command, I'd like the packages that are no longer referenced to be removed from disk.

cdrnet:
Downloaded packages in the `./packages` folder can also become garbage passively by VCS operations like `git checkout`, where the lock file gets updated. Dropping downloaded packages as part of the install/update process (based on the changes it applies to the lock file) would therefore only reliably keep the `./packages` folder clean if that folder is actually checked into the VCS as well.

Do we assume that Paket "owns" the `./packages` folder, so it could technically garbage collect based on the lock file? If not, can it reliably find out which of the packages within were originally downloaded and put there by Paket?

forki:
Yes we assume packages folder is owned by paket.

forki:
comes with v3

PR #1482, opened 2016-02-18
Garbage collection

cdrnet:
I've started on a proposal how we could implement garbage collection to clean up packages no longer needed in the `packages` folder (and if possible the same for `paket-files`). Addresses #1416.

Since the packages folder is typically excluded from version control, we do not always have explicit information about a package being removed. What we do know is what packages are currently extracted in the packages folder, and what packages we actually expect to be there according to the lock file.

The proposed algorithm essentially goes through the subdirs of the packages folder, tries to imply group and package names, and drops all of these folders which we identify as a extracted package but are *not* known by the lock file (ignoring the version).

Questions:
* Is it safe to assume that the name of the extracted nuspec file matches the package id, or do we need to parse it to be sure? We could also interpret the folder name instead, which we do construct ourselves (with or without version).
* When/where would we want to do this? Should it be explicit? Only on demand? Currently it is rather aggressive: always after add, remove and install, unless --no-install is set, and always after restore. I expect this to be too aggressive in practice.

### DIFF
--- a/src/Paket.Core/AddProcess.fs
+++ b/src/Paket.Core/AddProcess.fs
@@ -32,7 +32,7 @@ let private add installToProjects addToProjectsF dependenciesFileName groupName 
 
         if installAfter then
             InstallProcess.Install(options, hasChanged, dependenciesFile, lockFile)
-            GarbageCollectionProcess.DeleteUnusedPackages(Path.GetDirectoryName(dependenciesFileName), lockFile)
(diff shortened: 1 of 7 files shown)


forki:
I think GC is very important, but can you please send this against the v3 branch? I think this feature might come with some complications and therefor we should play it safe and add it to Paket 3.0

cdrnet:
Of course, I'll open a new PR on v3 (rebased)

forki:
thx

Issue #1471, opened 2016-02-12
Incremental build with MsBuild


Vilmir:
We have recently updated our product codebase to use Paket instead of Nuget for managing our package dependencies.
Our developers are quite pleased by this move, but we discovered a cumbersome behavior of MsBuild in incremental builds.

Our build script builds all our csproj in a single MsBuild task. It is used as an incremental build, meaning that for most of the situation we do not clean anything before building. This was working well with NuGet. But with Paket, csproj may not be rebuilt while they shall.

It happens when someone updates an existing library package to a new version, without modifying anything in a given project using this lib. Paket updates the paket.lock file and does not modify the csproj using the library, because the unzipped path does not change.

Visual Studio handles this case well, it understands that the referenced assembly did get an update and rebuilds the csproj.
MsBuild does not handle this case, and does not rebuild the csproj. It seems that MsBuild does not consider referenced assemblies as inputs to decide whether or not to rebuild.
The assembly produced by the csproj still uses the old, now depreciated library assembly. If this assembly is called, it fails with a dependency not found error.

I have created a small project here to repro the problem with the library Autofac:
https://github.com/Vilmir/IncrementalBuildCheckWithPaket

Repro steps:
* Open Developper Command Prompt for VS 2015
* msbuild.exe build.proj
* Run bin/Debug/IncrementalBuildCheck.exe it works
* ".paket/paket.exe" update nuget Autofac version 3.5.2
* msbuild.exe build.proj
* Run bin/Debug/IncrementalBuildCheck.exe, you shall see an exception:
```
System.IO.FileLoadException: Could not load file or assembly 'Autofac, Version=3.0.0.0, Culture=neutral, PublicKeyToken=17863af14b0044da' or
 one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131
040)
File name: 'Autofac, Version=3.0.0.0, Culture=neutral, PublicKeyToken=17863af14b0044da'
   at IncrementalBuildCheck.Program.TestLibrary()
   at IncrementalBuildCheck.Program.Main(String[] args) in C:\Users\11mvazquez\Desktop\IncrementalBuildCheckWithPaket\Program.cs:line 12
```

forki:
That's an interesting point, but I'm not sure what we can do here. The goal was always to keep the csproj as stable as possible. 
What do you suggest?

cdrnet:
An approach that might help here could work along the lines of

    paket.exe restore
[...]

PR #1485, opened 2016-02-22
Restore: optional --touch-affected-refs to touch refs affected by a restore

cdrnet:
Proposal how  #1471 could be addressed, to help incremental build tools like MsBuild to detect the need to rebuild selective projects.

I'm not convinced `Project.save true` is the right approach though, since it could also affect the content of the project file. Maybe we should specifically add a `Project.touch` instead which guarantees the project file content remains exactly the same?


forki:
Project.touch sound s like a good change

PR #1492, opened 2016-02-29
Reworked test/Fix for #1477 (locked files after restore)

haraldsteinlechner:
test and fix for: https://github.com/fsprojects/Paket/issues/1477
seealso: https://github.com/fsprojects/Paket/pull/1480


forki:
thx

PR #1493, opened 2016-03-01
ProjectFile.save with forceTouch to only modify the last write time without content if unchanged

cdrnet:
This is a follow up to #1485 which introduced the optional --touch-affected-refs
parameter on restore. In the original implementation this caused the
project file to be saved forcefully, which could in some cases cause the
file content to change slightly (e.g. minor xml formatting).

With this change, save with forceTouch=true only saves the content if it has actually changed. Otherwise it only sets the last-write-time to now.

### DIFF
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -941,6 +941,9 @@ module ProjectFile =
                 project.ProjectNode.AppendChild analyzersNode |> ignore
             )
 
+    let touch (project:ProjectFile) =
+        if File.Exists(project.FileName) then
+            File.SetLastWriteTimeUtc(project.FileName, DateTime.UtcNow)
 
     let save forceTouch project =
         if forceTouch then 
@@ -1305,7 +1308,9 @@ type ProjectFile with
     member this.RemovePaketNodes () = ProjectFile.removePaketNodes this 
 
     member this.UpdateReferences (completeModel, usedPackages, hard) = ProjectFile.updateReferences completeModel usedPackages hard this
-         
+
+    member this.Touch () = ProjectFile.touch this
+
     member this.Save(forceTouch) = ProjectFile.save forceTouch this
 
     member this.GetPaketFileItems () = ProjectFile.getPaketFileItems this
--- a/src/Paket.Core/ProjectFile.fs
+++ b/src/Paket.Core/ProjectFile.fs
@@ -941,16 +941,17 @@ module ProjectFile =
                 project.ProjectNode.AppendChild analyzersNode |> ignore
             )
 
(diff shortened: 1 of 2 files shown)


src/Paket.Core/ProjectFile.fs
```diff
@@ -941,6 +941,9 @@ module ProjectFile =
                 project.ProjectNode.AppendChild analyzersNode |> ignore
             )
 
+    let touch (project:ProjectFile) =
+        if File.Exists(project.FileName) then
```
mexx: It changes the logic previously present.
When the file wasn't present, it was created by the `save true project` call.
Please adjust accordingly.

cdrnet: It does not change any behavior as before the touching was only exposed indirectly through routines where the file was known to be present.

But I agree, `touch` is typically expected to create the file if not present. Since it is exposed directly now, we better adapt it to follow that expectation.

mexx: You maybe right about the way how it was ever accessed, but my comment was based on the how the logic is written in the save function and how it changed in the touch. I mean `touch` was intended to be equivalent to `save true`, right? No matter in which context is ever called.

cdrnet: No - touch is not intended to change the file content, only to make the file appear as if it had changed. Just like the touch tool on unix. The primary use case is to help msbuild incremental build detect it needs to rebuild a project.

Maybe it was a mistake to expose it as member?

mexx: I was a little bit incorrect in my wording, `touch` was intended to **replace** the `save true` call.

But you're right, why should it be there in first place?
The project file should know the own state, whether it's changed or not, right? So `save true` could actually utilize this knowledge to decide whether to simply change the last access date or write out the contents.

cdrnet: On second thought - maybe it would make more sense to slightly change the behavior of `save` instead of adding touch, along the lines of:
```
   let save forceTouch project =
        if Utils.normalizeXml project.Document <> project.OriginalText then 
            verbosefn "Project %s changed" project.FileName
            project.Document.Save(project.FileName)
        elif forceTouch && File.Exists(project.FileName) then
            File.SetLastWriteTimeUtc(project.FileName, DateTime.UtcNow)
```

cdrnet: Seems we've come to the same conclusion concurrently ;)

cdrnet:
Should I squash the commits together?

forki:
Did you look at the build?

forki:
some tests timed out.

cdrnet:
Thanks!

PR #1511, opened 2016-03-11
Added ability to include partials in the paket.template

magicmonty:
You can now include partials via `#load "<filename>"`

This is meant for having a base template and then just generating the version info and the release notes for example as include file via FAKE. So I don't have to generate everything via FAKE.

### DIFF
--- a/docs/content/template-files.md
+++ b/docs/content/template-files.md
@@ -196,3 +196,9 @@ This only works for paket.template files of type `project`.
 ### Comments
 
 A line starting with a # or // is considered a comment and will be ignored by the parser.
+
+### Includes
+
+A line with the format `#load "<filename>"` will be replaced with the content of the given file.
+
+The file must exist relative to the root file. If the file is not found, the line will be deleted.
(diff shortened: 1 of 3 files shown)


forki:
Version info and release notes can be easily replaced by using corresponding parameters of paket pack.

magicmonty:
We need to replace things like Authors etc internally also

forki:
Yes but why don't we add new parameter?

magicmonty:
IMHO this solution is more generic. If we add a parameter for every entry we don't need a template file anymore. 
* Also what is with multi line release notes?
* What is, if I want to add something to the description (e.g. "built by xy")
* Setting the copyright year automatically

magicmonty:
I just had a discussion wit a collegue this morning, 
Maybe it is better to make a FAKE helper, which can read the `#load` command and make the replacements there. This would be compatible with the existing paradigm in Paket to take just one file and use it as it is (if the `#load` was not processed, it would be considered as a comment).

forki:
Yes a fake helper sounds more useful. But please take a look at fake's
nuget pattern replacement. That worked very well in practice. I think it
would be better than #load

magicmonty:
OK, thanks for the hint, I'll have a look

inosik:
[This](https://gist.github.com/inosik/64707bfa6c944868f617) is what I do in some of my projects. It's actually the same as the NuGet stuff in FAKE.

Issue #1524, opened 2016-03-17
Resolving dependencies takes too long


Stift:
We have around 90+ direct dependencies and like ~200 dependencies in total in our project and since v2.52.9 the resolving takes 'forever'. It seems for me, that the fix of #1520 has caused that problem. Before an `paket install` took like 7-10s and now we are waiting +15min at least (I had to stop this and did not measure the correct timing).

forki:
yeah looks like a massive perf issue. I will investigate
/cc @isaacabraham

forki:
I just reverted that change. Can just please check if perf is restored? Seems I need to investigate deeper

forki:
pushed 2.52.12 - I think I verschlimmbessert it in 2.52.11

Stift:
Tested it with 2.52.12. and it's back to normal. Thx.

forki:
OK now I need to find a proper fix for #1520 
@Stift I guess I need to push another version soon and ask you again if perf is still good ;-)

Stift:
Just do so. I'm glad to test it.
But today is St. Patricks Day and I'm off and out in a few hours:
![](http://i.giphy.com/10DIbtLvhlDdra.gif)

forki:
2.52.13 will be online in 2min. Please give it a try

Stift:
that will fit

forki:
so it's online

Stift:
I'm already testing it. I get some warnings that I cannot specifcally say what they are about:

    Could not resolve package Npgsql:
     - Dependencies file requested: 2.2.5-custom
     - <customPackage1> 1.0.11044 requested: >= 0
     - <customPackage2> 1.0.11149 requested: >= 2.2.5
     - <customPackage2> 1.1.10087 requested: >= 2.0.0 < 3.0.0

Also some of the numbers do not fit to the package itself. Is something mixing up here?

forki:
shit. soo I'm not done

Stift:
Also after a second run of `install` downloads now all packages again and this again takes a while.
First run: 26s
Second run (downloading?): +5min

Stift:
sorry for that

forki:
ok this one is really driving me nuts.

I released another version. Can you give me your deps file by any chance?

Stift:
I'll send it to you, but it contains many of our own packages and we use also a Klondike server as backend. i'll send it to you via mail.

forki:
[...]

PR #1530, opened 2016-03-18
Improve warning text on conflict

Stift:
While having some issues (#1524) I stumbled upon some warning in Paket which I found misleading. This PR should just improve these and give a better direction.


forki:
Now we need Elm-like messages that say what to do ;-)

PR #1538, opened 2016-03-22
Correct src folder structure for packing with symbols

kevinbosman:
Packing a NuGet package with symbols generates the incorrect src folder structure within the symbols.nupkg file, which prevents the source from being found when querying the symbolsource server.

This PR adds an integration test with a few examples in a new scenario, and fixes the problem.

The fixed symbols package has been tested to be working correctly on our local ProGet symbols server from Visual Studio 2015.


forki:
don't worry about the broken test. I'm playing with something

kevinbosman:
Cool thanks, it was timing out on Travis only, so I kicked off another build...

forki:
sorry didn't see that. there was indeed a broken test here. "#1538 symbols src folder structure" doesn't work on mono. could you please look again?

kevinbosman:
Will do...

>

PR #1541, opened 2016-03-23
prevent packing of sources from doubling up on the project directory …

pms1969:
…when adding files in the root of the project

Example. we are packing project A with source tree like so
```
projectA
    + SubDir
        + class1.cs
        + class2.cs
    + class3.cs
    + class4.cs

```
when packed it should look exactly the same, but currently the following directory structure is produced
```
projectA
    + SubDir
        + class1.cs
        + class2.cs
    + projectA
        + class3.cs
        + class4.cs
```

This is obviously incorrect.

### DIFF
--- a/src/Paket.Core/PackageProcess.fs
+++ b/src/Paket.Core/PackageProcess.fs
@@ -81,7 +81,9 @@ let private convertToSymbols (projectFile : ProjectFile) (includeReferencedProje
                        | None -> compileItem.Include
             
             let tld = Path.GetFileName(compileItem.BaseDir)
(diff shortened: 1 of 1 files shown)


kevinbosman:
See also #1538 which addresses the same problem

pms1969:
so it does.  you can probably ignore this then.

forki:
so #1538 is solution for #1541 and https://github.com/fsprojects/Paket/pull/1540 ?

kevinbosman:
No, not #1540.
#1538 only addresses src folder structure when packing with symbols.

PR #1543, opened 2016-03-23
ensuring that dependent dll's are not included in the package when us…

pms1969:
…ing include-referenced-projects

as per requested in (https://github.com/fsprojects/Paket/pull/1540).  Separate PR in order to keep the history clean.


pms1969:
I'm having a great deal of trouble running any tests locally at present.  Probably just a work thing.  I'll keep an eye on the build progress.

pms1969:
Is that a relevant failure on Travis?

forki:
nope. that's unrelated

forki:
@kolektiv @inosik would this still work for you?

inosik:
I haven't been using this parameter anyways, so I think it should be fine.

forki:
ok let's try it and see what happens.

PR #1547, opened 2016-03-24
removing transitive dependencies from dependencies list

pms1969:
brings the dependency list in line with exactly what nuget produces (as far as I can tell).

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -243,11 +243,38 @@ let findDependencies (dependenciesFile : DependenciesFile) config platform (temp
 
          yield! getPackages project]
     
-    match allReferences with
+    //System.Diagnostics.Debugger.Launch() |> ignore
+
+    // filter out any references that are transient to the 
+    let distinctRefs =  allReferences 
+                       |> Seq.distinct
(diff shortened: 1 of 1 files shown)


forki:
Do we really want that? Is it needed?

forki:
(I mean probably want that - but what are we breaking currrently?)

pms1969:
I guess it's more of a completeness thing than anything else.  it declutters the output, and makes the dependency list that is generated by paket the same as nuget.

in a project of any size, (and I have a few at work), the depency list is relatively large.  a good chuck of these dependencies are transitive.  it leads to larger lock files I guess, and I need to justify the difference to the people I work under ;-)

forki:
ok will look into it

forki:
Could you please review b014e55 and 11a4308? Thanks

pms1969:
all looks good.  just ran trunk against my solution here, and it worked as expected.

forki:
Do we influence the packaged dlls with it? I mean for packages that don't use dependencies but bundle their stuff we still need to package all deps. Otherwise that would create a bug right?

pms1969:
that code just has an impact on the dependencies list, not on the packaged dlls

Issue #1329, opened 2015-12-24
[Feature Req] Possibility to specify a nupkg cache location


CumpsD:
This issue is to discuss a Twitter idea

# Initial idea
https://twitter.com/CumpsD/status/680011065250443264
https://twitter.com/CumpsD/status/680011137853820930

# Background story
Paket is great because it gives me repeatable builds with versions I specify. I can safely sleep at night knowing if my project builds today and collects dust for 5 years, it will still build with Paket.

However, I don't always belief things to stay around forever, especially with central package repositories. Things happen. Who can guarantee the packages I use will be there 5 years from now? or 10 years? If I remember correctly, `npm` had this discussion once, which is why Proget and other proxies exist which cache packages so you keep them even if they are gone from the central place.

# Idea
What if I could just tell Paket to store all `nupkg` files it downloads in a cache, of which I can specify the location. No need for proxy setups or anything, simply run paket, and have it store to a fileshare or something and that's your "backup". Later when restoring, Paket first checks your usual feeds, with the existing logic, and if it doesnt find anything (NuGet down, package removed?) it can fall back to your offline cache.

_Clarification:_
This cache location should be possible to specify per invocation, so you can have different ones per repository for example. Cmdline argument or `paket.dependencies` option?

Discuss!

CumpsD:
Additionally, if I tell paket to store the `nupkg` files in the repository itself, and I use Git Large File Storage to deal with `nupkg`, it would be nice too :) Then they are all in the repo as a backup, without exploding the repo size.

amazingant:
FWIW, at work my package source is a network share so that I don't have to worry about packages disappearing. I've had a few packages that randomly pulled old versions that I was depending on. This means I have to manually go out and look for updated versions of packages, so I wouldn't mind if Paket gained the ability to, say, use that network share as the "cache" location per the above suggestion.

CumpsD:
Pinging @forki for feedback about the idea :) Before working on a PR

forki:
Yes I think additional caches would be a nice feature for the v3 branch. 

* They should be configured in the dependencies file. 
* clear-cache command would need options

forki:
Ok after node drama I think we shold make that a thing.

I see two variants. 

1) We specify a network share and store all package versions
2) we specify a folder inside the repo and store only the latest nupkgs
   * so you can commit that folder and still gitignore the packages folder

forki:
I propose the following syntax:

    source https://nuget.org/api/v2
[...]

PR #1549, opened 2016-03-24
Implement additional caches

forki:
as described in #1329


src/Paket.Core/Cache.fs
```diff
@@ -2,11 +2,47 @@
...
+    | CurrentVersion
 
 type Cache = 
     { Location : string
-      CacheType : CacheType }
\ No newline at end of file
+      CacheType : CacheType option }
+
+    static member Parse(line : string) =
+        let sourceRegex = Regex("cache[ ]*[\"]([^\"]*)[\"]", RegexOptions.IgnoreCase)
```
mariusschulz: There's no need for the first and last character class here: `"cache[ ]*\"([^\"]*)\""` should be fine.

adamchester:
Just FYI I gave this a few quick tests and it seems to behave as I expected. I wondered the first time if the packages missing from the local cache deserved a "something went wrong" message.

forki:
Thx. Will continue this later today. There is a lot missing

src/Paket.Core/Cache.fs
forki: it's copied from nuget packages sources and I might add passwords here

forki:
> I wondered the first time if the packages missing from the local cache deserved a "something went wrong" message.

what does this mean?

adamchester:
I'm not in front of a computer right now, so I can't get the exact message, but it gave a message that indicated something was wrong when the package wasn't in the cache. It seemed a little excessive to say something was "wrong".

Issue #1260, opened 2015-11-26
Referencing XAML resources


sergey-tihon:
When I you reference to [MahApps.Metro.Resources](https://www.nuget.org/packages/MahApps.Metro.Resources/)

Paket add following code into project file
```
    <Content Include="Resources\Icons.xaml">
      <Paket>True</Paket>
    </Content>
```
In the same time NuGet add it like 
```
    <Page Include="Resources\Icons.xaml">
      <Generator>MSBuild:Compile</Generator>
      <SubType>Designer</SubType>
    </Page>
```

As a result when MSBuild does not compile `xaml` file and application crashes in the runtime when tries to load resources.

sergey-tihon:
One more note, project previously used NuGet, it contains `<Page>` node, so Paket added duplicate files with `<Content>` nodes.

As well as Paket adds
```
<Content Include="Resources\Entypo.ttf">
      <Paket>True</Paket>
</Content>
```
even if NuGet already added this as
```
<Resource Include="Resources\Entypo.ttf" />
```
(leads to duplication in file explorer)

forki:
I assume our xaml support is pretty poor (read as non-existent). We could
need some help here.

sergey-tihon:
Not sure that I know how to support all cases...
 
- Migration from NuGet
- Install dependencies correctly
- Track/Update such dependencies

Could please mark this issues and `up-for-grabs` and mention some files that should be changed according to you feelings :smirk:

forki:
I assume the fix is needed in https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/ProjectFile.fs#L447

forki:
ping

PR #1550, opened 2016-03-24
Better WPF support

sergey-tihon:
Attempt to add support of [new Build Action (Resource & Page) specific to WPF projects](https://msdn.microsoft.com/library/aa970494%28v=vs.100%29.aspx?f=255&MSPPError=-2147217396) 
Related issue https://github.com/fsprojects/Paket/issues/1260


sergey-tihon:
@forki It seems that it works, at least in my cases.
Should I add integration tests on something else?

forki:
That would be great. (otherwise I will probably  break it) 

There are many many integration tests that you can copy. Just look at bit around

sergey-tihon:
This PR is ready for merge after review

forki:
Awesome!

cat

Issue #1552, opened 2016-03-28
Inconsistent condition generation 


sergey-tihon:
Almost each time when I call `paket.exe install` it change my `paket.lock` file from
> CommonServiceLocator (1.3) - framework: >= net40, monoandroid, portable-net45+wp80+wpa81+win+MonoAndroid10+xamarinios10, xamarinios, winv4.5, winv4.5.1, wpv8.1, wpv8.0, sl50

to
> CommonServiceLocator (1.3) - framework: >= net40, monoandroid, portable-net45+wp80+wpa81+win+MonoAndroid10+xamarinios10, xamarinios, winv4.5, wpv8.0, sl50

and vise versa.

What actually change most of my csproj files from
[...]

PR #1554, opened 2016-03-28
Test "#1552 install mvvmlightlibs again"

sergey-tihon:
This test should fail, I hope it simplify troubleshooting


forki:
It fails. So what's the plan? Should I merge it with your other pull request?

sergey-tihon:
@forki This PR should be a repro for #1552 (just to simplify investigation)
You may close\decline it when understand what is wrong here

forki:
Ah got it

PR #1556, opened 2016-03-29
Improved consistency test for #1552

sergey-tihon:


forki:
this test is not broken on my machine.

forki:
ok it fails on travis.

sergey-tihon:
As well as on my machine
image

forki:
trying to look deeper

forki:
I can now reproduce by using -f

PR #1557, opened 2016-03-29
New repro for #1552

sergey-tihon:


forki:
new release under way

PR #1560, opened 2016-03-29
install mvvmlightlibs first time with wpav8.1 framework in lock file

sergey-tihon:
This test work on my machine and produce lock file with `wpav8.1`
image

### DIFF
--- a/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
@@ -377,11 +377,16 @@ let ``#1552 install mvvmlightlibs again``() =
         |> normalizeLineEndings |> shouldEqual expected
 
     prepare scenarioName
-    ["install -f"
-     "update -f"
-     "install"
-     "update"]
-    |> List.iter lockFileShouldBeConsistentAfterCommand
+    let commands =
+        ["install -f"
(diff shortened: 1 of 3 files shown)


sergey-tihon:
Seems that on travis paket create lock file with `wpav8.1` - https://travis-ci.org/fsprojects/Paket/builds/119346651#L2424
@forki You should be able to reproduce it on you machine as well

sergey-tihon:
image

forki:
ok I get the same error as https://travis-ci.org/fsprojects/Paket/builds/119373387

    Expected string length 18715 but was 18660. Strings differ at index 2594.
    Expected: "...argetFrameworkIdentifier) == 'WindowsPhoneApp') Or ($(Targ..."
    But was:  "...argetFrameworkIdentifier) == '.NETCore') Or ($(TargetFrame..."

but it's an error in the csproj - not in the lock file!?

forki:
OK now I can reproduce. I needed to change it to " File.ReadAllText newLockFilePath" instead of LockFile.Load and ToString

sergey-tihon:
I think that issue with project files is another one.

>I needed to change it to " File.ReadAllText newLockFilePath" instead of LockFile.Load and ToString
Why resource can be different?

forki:
ok new version was released

Issue #1537, opened 2016-03-22
VS2015 Update 1: MVC build fails with error MSB4064: The "Features" parameter is not supported by the "Csc" tas


mlidbom:
There are a few twists to when you encounter this bug.
I eventually found that it occurs every time if: 

- Packages are restored before build starts.
- The obj folder is missing
- The bin folder is missing

I've attached a zip with a repro solution. 
Run build-fails.bat to see the build fail.
Run build-succeeds.bat to see the build succeed.

**Edit: Seems the problem occurs with VS2015 Update 1. I'm afraid this likely means that every paket user with an MVC project will start to see their builds failing when they update.**

Edit2: The exact error from the build log: 

```
2>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.CSharp.Core.targets(88,11): error MSB4064: The "Features" parameter is not supported by the "Csc" task. Verify the parameter exists on the task, and it is a settable public instance property.
2>C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.CSharp.Core.targets(67,5): error MSB4063: The "Csc" task could not be initialized with its input parameters. 
```

[PaketBugRepro.zip](https://github.com/fsprojects/Paket/files/184100/PaketBugRepro.zip)

inosik:
Strangely enough, both scripts work on my machine.

And actually, it shouldn't make any difference, since `paket.targets` already does `paket restore` before building.

There is somebody else having the same issue at [StackOverflow]( http://stackoverflow.com/questions/34561218/visual-studio-community-2015-update-1-build-issue) and at the Roslyn repo (dotnet/roslyn#9538).

mlidbom:
That is strange indeed. 
I'm running in in the "Developer Command Prompt for VS2015"
Are you using the same version?

inosik:
I don't have VS 2015 on this machine, only up to 2013.

So this is some MSBuild f**kery going on here, not an Paket issue.

Stift:
I only can reproduce it when I explicitly use `C:\Program Files (x86)\MSBuild\14.0\bin\msbuild.exe`

mlidbom:
It may well be msbuild related, but before paket modified the csproj file it worked just fine.
I on the other hand don't have VS2013 on this machine. Will try to see if I can borrow one to see if I can verify that it works on VS2013.

mlidbom:
[...]

PR #1561, opened 2016-03-30
Property Definitions: placed after non-paket imports if they directly follow the top property groups

cdrnet:
Related to #1537

If in a project file there are non-paket imports right after the top property groups, then place the property definitions right after them (instead of before them). This makes sure that imports like `Microsoft.CSharp.targets` can be moved up before them, to make sure`TargetFrameworkIdentifier` is actually defined. Note that it does *not* move `Microsoft.CSharp.targets` up manually if it is further down (which it is by default).


cdrnet:
Should I rather have created/forked a new integration test instead of changing existing ones?

forki:
Yeah more tests are probably better

forki:
but I think it's fine

cdrnet:
Note that in the quite common case that `Microsoft.CSharp.targets` is further down this change will only allow to manually fix it by moving it up (and make the fix survive paket operations), but will not actually fix it itself. 

Should we try to be smarter? Being smart could easily break other stuff unfortunately.

forki:
I think we should wait and see what happens - also in regards to #1562

PR #1567, opened 2016-04-04
Remove --hard flag

forki:
## Background

    --hard: Replaces package references within project files 
              even if they are not yet adhering to the Paket's conventions 
              (and hence considered manually managed).

### Sample

in default mode paket 2.0 will not change references to libraries if they are not already controlled by paket. A common case is "file/ enw F# project" creates a fsproj file that already references FSharp.Core. If you then add FSharp.Core as nuget package, then Paket will not replace the reference unless you specify --hard parameter. 

## Suggestion

I suggest we remove the --hard flag in paket v3 and thread all calls as if --hard would have been set.

- [x] removed flag from all commands 
- [x] fixed tests
- [x] update docs

We will not create a --soft flag to restore old behaviour, but if you really need to keep that existing reference you can manually add a 

    <paket>false</paket>

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,9 +1,10 @@
-#### 3.0.0-alpha095 - 31.03.2016
+#### 3.0.0-alpha096 - 04.04.2016
 * Allow to reference git repositories - http://fsprojects.github.io/Paket/git-dependencies.html
 * Allow to run build commands on git repositories - http://fsprojects.github.io/Paket/git-dependencies.html#Running-a-build-in-git-repositories
 * Allow to use git repositories as NuGet source - http://fsprojects.github.io/Paket/git-dependencies.html#Using-Git-repositories-as-NuGet-source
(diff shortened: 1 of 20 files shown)


forki:
This is obviously breaking, but it was a big source for confusion. 

I'm only suggesting this for v3 so I'm relaxed about it like this guy:

pizza

isaacabraham:
This is good, it's what I think people expect anyway as the default.

agross:
:+1:

tsibelman:
It fine to remove it in my opinion. But what would happen if  I try to use it ? Will I gen an error or just it will be ignored ? I prefer not to have an error this way I don't need to update my scripts.

forki:
With this pr we would throw an error: unknown parameter --hard

tsibelman:
Can this be a warning instead ?

forki:
Since this is a major release change I would prefer not to take that debt.

isaacabraham:
Agree with @forki - this a major release with a breaking change that should be documented as part of the release.

Issue #1568, opened 2016-04-04
Paket Push doesn't make use of the stored credentials for a given source.


baronfel:
if we specify a source url on the command line, should we be able to fetch the credentials from a stored set of credentials like we do for install/restore?

forki:
that would probably make sense, but most use apikey!?

baronfel:
I agree, but we do allow to set api tokens in the config already so couldn't we fetch from there?

forki:
Yes. If you construct a repro sample, then it will be easy to make that
work

baronfel:
Oh I was going to send a PR a bit later with a test and doc updates, if that's ok.

forki:
Even better. ;-)

PR #1570, opened 2016-04-04
Apply any saved api key credentials to the push operation

baronfel:
Addresses #1568 by roping in a call to the Paket config for a token for the given url and deferring to that only when there is non command-line or env variable for the token.


forki:
thanks

PR #1580, opened 2016-04-07
Update: support combining filter with specific version

cdrnet:
Update supports updating all packages matching a regex. It also supports providing a specific version to update to. What it did not support before is combining these two, i.e. updating all packages matching a regex to the same specific version. This PR adds the missing piece to support this combination.


cdrnet:
NB: the matching logic seems somewhat risky to me, there is a chance for collisions with argument/option keywords. Isn't "nuget" required to be the first (index 0) and the package name the second token (index 1) for nuget dependencies, provided we specify `StringSplitOptions.RemoveEmptyEntries`?

forki:
yeah. can you fix that as well?

cdrnet:
One integration test timed out ("1371 without download fail"). Could this be related to my change?

forki:
probably not. that one is flaky.

forki:
thanks a lot

PR #1588, opened 2016-04-10
change from wget to curl in install.sh script

jfellien:
Cause the OSX does not contains `wget` native, I changed the download of paket.exe in install.sh script to curl. In my tests (existing and not existing paket.exe) everything works fine.


forki:
monkey

PR #1603, opened 2016-04-12
Refactoring Bootstrapper to introduce better coverage and testing

Stift:
I have completely refactored the bootstrapper. The initial problem why I started this is just that this simple tool got rather complex during the last year. In the beginning it was just ~80 lines, but more and more features were integrated which introduce also bigger complexity. Therefore, I refactored it to have some test in place and also removed some bugs which I discovered btw. I think this gives some benefit to all users of paket, as everyone checks in the bootstrapper and rely on its correct functionality.

The following items are tested from now on:
- parameter evaluation
- download strategies (github, nuget, cached)
- download strategy resolution
- bootstrapping behaviour

I also improved the CachedDownload strategy which before behaved like a bootstrapper on its own and now is only a strategy, and rely that the bootstrapper will do his job.

After refactoring we should have a test coverage of ~90%, before there was nearly none. The only downside is, that my refactoring increased the size of the bootstrapper by ~5kB, which is excusable if I look on the plus side.

Besides that the following bugs were removed
- When you call `paket.bootstrapper.exe prerelease` and after that `paket.bootstrapper.exe`, the second call does not use the latest version
- minor issue with selecting versions in nuget strategy when prereleases were not ignored effectively

The following features were also added.
- non evaluated command line arguments are printed on console
- with the parameter --help a help page is now displayed with short explanation of parameters

I'm sorry for the bigger PR, but for me it seemed like the best way to succeed.


forki:
clap

src/Paket.Bootstrapper/DownloadStrategies/GitHubDownloadStrategy.cs
```diff
@@ -0,0 +1,131 @@
...
+            }
+            return versions;
+        }
+
+        public void DownloadVersion(string latestVersion, string target)
+        {
+            var url = String.Format(Constants.PaketExeDownloadUrlTemplate, latestVersion);
+            ConsoleImpl.WriteDebug("Starting download from {0}", url);
+
+            //using (var httpResponseStream = WebRequestProxy.GetResponseStream(url))
```
forki: could you please remove commented code. Hurts my OCD ;-)

Stift:
Yeah. Sorry for that. Didn't want to hurt you.

vasily-kirichenko:
@Stift great huge job! I wish we could write in in F#, in FP style, but we cannot.

Stift:
@vasily-kirichenko Thx.
I also would like to do that instead of doing this csharpiness stuff, but the Fsharp.Core still blocks us from doing so.

forki:
I think it's fine for now, but it would make a cool case study to refactor this further to get rid of mocks and interfaces in C#. 
But on the other hand I'm not sure if that brings us anything. This the bootstapper and I'm super happy that someone takes time to make that thing work well.

Stift:
You can get rid of interfaces and mocks, but then you use delegates all over the place and pass them around. At least, the code is still readable. As said before, I just wanted to have tests for now and increase the coverage on that tool. 
I'm absolutely happy if someone wants to join and make this cool case study (with me). Now that we have testing in place this should be even easier.

forki:
yes primary goal is to make that thingy useable and get everything under test. well done!

PR #1605, opened 2016-04-13
recognize .netstandard framework

theimowski:
Apparently, .netstandard is going to replace all `dnxcore` monikers as described here: https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#deprecated-monikers

### DIFF
--- a/src/Paket.Core/FrameworkHandling.fs
+++ b/src/Paket.Core/FrameworkHandling.fs
@@ -77,6 +77,7 @@ type FrameworkIdentifier =
     | DotNetFramework of FrameworkVersion
     | DNX of FrameworkVersion
     | DNXCore of FrameworkVersion
+    | DotNetStandard of string
     | MonoAndroid
     | MonoTouch
     | MonoMac
@@ -94,6 +95,7 @@ type FrameworkIdentifier =
         | DotNetFramework v -> "net" + v.ShortString()
         | DNX v -> "dnx" + v.ShortString()
         | DNXCore v -> "dnxcore" + v.ShortString()
+        | DotNetStandard v -> "netstandard" + v
         | MonoAndroid -> "monoandroid"
         | MonoTouch -> "monotouch"
         | MonoMac -> "monomac"
@@ -131,6 +133,7 @@ type FrameworkIdentifier =
         | DotNetFramework FrameworkVersion.V5_0 -> [ DotNetFramework FrameworkVersion.V4_6_1 ]
         | DNX _ -> [ ]
         | DNXCore _ -> [ ]
+        | DotNetStandard _ -> [ ]
         | Silverlight "v3.0" -> [ ]
         | Silverlight "v4.0" -> [ Silverlight "v3.0" ]
         | Silverlight "v5.0" -> [ Silverlight "v4.0" ]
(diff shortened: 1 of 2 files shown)


dsyme:
I don't know if this is related, but I tried a ``paket update`` on this and it didn't resolve:

```

source https://dotnet.myget.org/F/dotnet-core/api/v3/

nuget NETStandard.Library 1.5.0-rc3-24013
```

giving

```
C:\misc\repro-54>.paket\paket.exe update
Paket version 2.60.2.0
 - NETStandard.Library is pinned to 1.5.0-rc3-24013
Paket failed with:
        Could not find versions for package NETStandard.Library on any of ["http
s://dotnet.myget.org/F/dotnet-core/api/v3";
 "http://api.nuget.org/v3/index.json"; "https://www.nuget.org/api/v2"].
```

I don't know if it is driver error, e.g. incorrect feeds, or that the myget feeds are not coughing up the lists of available versions, or the same as the bug being fixed here.

forki:
source https://www.myget.org/F/dotnet-core/api/v3/index.json

brings:

image

forki:
source https://www.myget.org/F/dotnet-core/api/v3/index.json
    
    nuget NETStandard.Library 1.5.0-rc3-24013-00

seems to work.

dsyme:
@forki Thanks! silly me :)

forki:
I think we will make that more robust. Nobody really knows how these new
endpoints are really meant to be used. (or at least I don't know).

theimowski:
I removed the [WIP] prefix - I think we can merge it and fix whatever comes next

forki:
I added couple of other things to this, but Paket 2.61 now works with .NETSTANDARD. At least with Kestrel in https://github.com/CumpsD/kestrel-classic/pull/1

theimowski:
Cool thanks!

forki:
would love to hear if it works for you.

theimowski:
Will see in next days but hopefully it does :+1:

src/Paket.Core/FrameworkHandling.fs
```diff
@@ -222,6 +225,7 @@ module FrameworkDetection =
                 | "dnx451" -> Some(DNX FrameworkVersion.V4_5_1)
                 | "dnxcore50" | "netplatform50" | "netcore50" | "aspnetcore50" | "aspnet50" | "dotnet" -> Some(DNXCore FrameworkVersion.V5_0)
                 | v when v.StartsWith "dotnet" -> Some(DNXCore FrameworkVersion.V5_0)
+                | v when v.StartsWith "netstandard" -> Some(DotNetStandard "1.5")
```
dsyme: Does this gloss over the differences between "netstandard1.0", "netstandard1.3" and "netstandard1.5" and does it mater?  thanks

forki: Yeah I extended it to understand all published standard versions.

theimowski:
checked my use case - works fine. The remaining concern is indeed the unofficial prelease source feeds, which are required to pull down latest packages. Eventually the packages are to be published to the official feed, but till then it would be good to somehow guide users to work with prerelease feeds - not sure how to do this though

forki:
what concern do you mean? Adding endpoint to deps file should be enough, right?

theimowski:
Yes its enough but just as you stated above it's not obvious at once which feeds should be added

PR #1606, opened 2016-04-13
Paket is a dependency manager

inosik:
Minor thing. @forki removed "dependency" with d2c6fb9.


forki:
Shit. Is that the only place where I fucked up? I did search and replace.

inosik:
There are four `paket.template`s in the repository, this was the only one which had this. The other changes from d2c6fb9 look OK to me.

forki:
thanks

Issue #1598, opened 2016-04-12
Remove command - restrict install to specified grouped


theimowski:
### Description

When running `paket remove nuget XXX group YYY`, install process which is invoked afterwards, should be restricted to group `YYY` only

### Repro steps

Please provide the steps required to reproduce the problem

see #1596 

### Actual behavior

Currently, full install is performed after `remove`

### Known workarounds

use `--no-install` switch

### Related information 

see #1596

PR #1612, opened 2016-04-16
remove command: restrict install to the specified group only

theimowski:
didn't manage to fix the recursive resolver, but here's the smallest #fsharpx contribution possible ;)

references #1598


forki:
xoxo

PR #1617, opened 2016-04-17
Add new option --max-file-age 

magicmonty:
for temporarily preventing a delay while checking the new file version.

Use case:
The internet on my workplace is occasionally very slow. My build scripts call the bootstrapper on every call (in case paket.exe is missing). This leads usually to a hit to "GetNewVersion", which sometimes slows the build process down, as I build very often via command line.
With this command line parameter it is possible to delay this check for a new version so it will be checked for example only every 4 hours. If the paket.exe exists and the last write time of the paket.exe is not older than the given threshold, the bootstrapper will not check the version, so the network hit is not there anymore.


inosik:
I don't think that this should be in the bootstrapper, at least not in this form. I don't know what the better solution would be, though.

I solved this problem by splitting up the build script into a `build.cmd` and a `fake.cmd`. The `build.cmd` calls the bootstrapper, `paket restore` and then calls `fake.cmd %*`. The `fake.cmd` sets up the environment for FAKE, e. g. the default encoding, and then calls `FAKE.exe`.

I also noticed that nuget.org is faster for me than GitHub, so `--prefer-nuget` is hard-coded into `build.cmd`.

Also, the bootstrapper [supports custom NuGet feeds](http://fsprojects.github.io/Paket/bootstrapper.html). If you have an internal feed, you could use that as well.

magicmonty:
I have a similar setup. I've split it up in `paket.cmd`, `fake.cmd` and `build.cmd`. For me the `paket.cmd` always calls the bootstrapper. But if I just want to operate on paket (multiple calls, (e.g. `paket outdated`, `paket add ...`, `paket install`) I don't always want this hit on GetNewVersion. So I think, this could be useful anyway. 

If I had found a reasonable way in batch files to check for the last write date of the `paket.exe`, I would have done it here, but I havent found one yet (without an additional helper tool).

magicmonty:
Closing, because I accidentally rebased

magicmonty:
New updated pull request is #1697

PR #1628, opened 2016-04-19
Add support for MonoAndroid22 and MonoAndroid23.

davidschwegler:
We use some libraries that have minimum targets of MonoAndroid22 and MonoAndroid23, so Paket currently doesn't pick them up.


forki:
thx

PR #1631, opened 2016-04-21
Add monoandroid43 to be matched as Some MonoAndroid

dmunch:
Moniker monoandroid43 appeared in latest version of Xamarin.Android.Support.Design (23.3.0)


forki:
thx

Issue #1621, opened 2016-04-18
binding redirects should be created when two projects depends on each others but are referencing different assemblies


smoothdeveloper:
### Description

I faced situation where NUnit test projects targetting different framework but referencing one another would throw runtime error in test runner.

This is because assembly version is different between each framework for a given NUnit release.

### Repro steps

`paket.dependencies`

```
redirects: on
source https://www.nuget.org/api/v2

nuget NUnit = 3.0.1
```

have projectA depend on NUnit and compiled for .NET 4.0

have projectB depend on NUnit and projectA and compiled for .NET 4.5, projectB needs to call a member from projectA which depends on NUnit.

I'm adding a test in integrationtests.

### Expected behavior

My assumption is that `paket install` could detect this situation and generate correct binding redirect, but I'm unsure if it should go to that extent (changing the target framework again might break until next `paket install`).

### Actual behavior

This is what I get running the test from projectB

```
System.IO.FileLoadException : Could not load file or assembly 'nunit.framework, Version=3.0.5813.39032, Culture=neutral, PublicKeyToken=2638cd05610744eb' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

### Known workarounds

Write the redirect manually.

### Related information 

* Windows 7 64bit
* branch: master
* .NET 4 & 4.5

forki:
integration test in #1622 

/cc @isaacabraham @mrinaldi 
any ideas?

isaacabraham:
Sorry - can you provide the expected dependency redirects and what you actually get? Cheers

smoothdeveloper:
@isaacabraham the test has the expected redirect in the assert. Right now, no redirection is being written.

smoothdeveloper:
I'm facing the same with another dependency, it fails to compile in this case somehow.

It seems common practice to have different file versions for different framework which is somehow annoying.

Apparently, pinning the framework in paket.dependencies for that particular dependency doesn't seem to make the project point on correct assembly, I'll see if I can make another integration test for this.

In this case, the workaround is to remove the dependency and add it manually.

forki:
@mrinaldi sent a PR in #1634

PR #1634, opened 2016-04-22
Generates binding redirect for references targeting different profiles

mrinaldi:
See issue #1621


forki:
As always: huge thanks. :heart:

Issue #1633, opened 2016-04-22
requirement that restore be from the install source breaks CI builds


mlidbom:
### Description

When developing projects that refer to each other through nuget packages the strict requirement that a package must be restored from the same source that it was installed from breaks builds on the CI server and other developers machines.

**It is possible to work around with a different workflow, but that becomes very costly in time when you are talking about dozens of projects depending on project A instead of 1 as in the repro steps below.**
### Repro steps
1. Developer works on Project A and B. Project B references A through a nuget package. 
2. Developer locally bumps project A from 1.0 to 1.1 and updates A in B.
3. paket.lock now has A 1.1 installed from c:\DEV_FEED
4. Developer pushes the new version of A and B
5. The CI server builds A and publishes the new version of A.nupkg. to http://OFFICIAL_FEED
6. The CI server tries to build B but fails since A.nupkg does not exist in c:\DEV_FEED on the CI server

```
Paket failed with:
        The package A 1.1 can't be found in c:\DEV_FEED.
Please check the feed definition in your paket.dependencies file.
```
### Suggested solution: Development sources

paket.dependencies

```
source http://OFFICIAL_FEED development c:\DEV_FEED 
```

When updating-to/installing a package from c:\DEV_FEED it would be entered as installed from http://OFFICIAL_FEED in paket.lock
When restoring paket would fall back to packages in c:\DEV_FEED if package is missing in http://OFFICIAL_FEED
### Alternative solution: Try each source

set an option in paket.dependencies
allowRestoreFromAnyFeedTryFeedsInOrderDefined true

With that option set paket would check for the existence of a package in each feed in order when restoring. 
You would define the official feeds first and the development/local feeds later so that official packages would always be preferred.

forki:
yes overwriting sources on dev time is on our agenda. But I think it should be in a different file (to allow you to gitignore it).

I stil need an diea how the format would look like

mlidbom:
How about something like this?

paket.dependencies
```
source OFFICIAL_SOURCE
```
paket.sourceoverrides
```
source OFFICIAL_SOURCE fallback LOCALLY_BUMPED_PACKAGES prefer EXPERIMENTING_DEBUGGING
```
When installing/updating paket.lock would always register the packages as coming from the feed in paket.dependencies

When restoring paket would look up a requested package in this order
1. prefered
2. paket.dependencies
3. fallback

mlidbom:
[...]

PR #1636, opened 2016-04-23
paket.local file override

theimowski:
I created an integration test for the most basic scenario: in `paket.lock` you have nunit 2.6.3, and there is nunnit 2.6.4 in local dir. During restore `paket.local` should override the nunit to be fetched from local dir and as result the nuspec should have version 2.6.4.
Let me know if this makes any sense.

References #1633


theimowski:
I'm wondering how to attach to the restore process - my idea is that a temporary `paket.local.lock` couold be created to run resolver once again as there could potentially happen version conflicts.
Other solution is to assume the overriding package has the same dependencies, but not sure if we will shoot ourselves in the foot with such approach.

mlidbom:
Interesting how sometimes it is only once you start seeing implementation details that you realize how differently you imagined something :)

I never imagined that the version of a package used would change. I only ever wanted to allow for packages to be restored from a different feed than the one they were installed/updated from. I have no problem with having to run paket update to switch to a new version of a package. 

It seems we have been talking about very different features. No wonder how little sense our arguments made to each other :)

theimowski:
The reason I've used a different version of package for this test was to empasize that package was restored from overriden source as defined in `paket.local` rather than original nuget source.
TBH I'm still not sure the use case you [desribed](https://github.com/fsprojects/Paket/issues/1633#issuecomment-213778323) is something paket should support - this kind of "dependency cascading" seems a bit cumbersome and could be very confusing for users.
The original idea as desribed by @forki in [his talk](http://forki.github.io/PaketIntro/#/47)  ([video here](https://skillsmatter.com/skillscasts/7389-paket-3-embracing-git-for-dependency-management)) from what I understood aims to target a **much simpler use case**, where the "local source override" relates to one package referencing another.

mlidbom:
I sort of thought that the "original idea" discussed in the issue was mine, since I created the issue and described the scenario in some detail ;)

I looked through the slides you linked and found nothing that seemed relevant. The video seems to be a generic introduction to paket and it is quite long. Is there a specific slide or part of the video that you are referring to? I'd prefer not to have to watch an hour of video about things I already know to hopefully catch the relevant part...

>TBH I'm still not sure the use case you desribed is something paket should support - this kind of "dependency cascading" seems a bit cumbersome and could be very confusing for users.

Actually, from my usage so far of paket it seems that paket already works this way, just like nuget, in this respect except for when restoring. When installing or updating it installs packages from the first feed in which the highest version it can find exists. It is only when restoring that paket restricts the package to a single feed(The feed it was installed from). I would be perfectly happy if this restriction was simply removed. But I assume that it was put in for some reason so I suggested changes that would let us work around it while leaving the restriction in place.

mlidbom:
Regardless of our other discussion I think you need to define very clearly which algorithm is used to select the package for the feature that you are imagining to become clear/understandable/easily-documented.

Examples of what I mean:
1. The highest version found in either feed with the "override" feed preferred if both contain the package version.
2. The highest version found in the "override" feed regardless of what exists in the official feed.
3. The version specified in paket.lock with package from the "override" feed preferred(prefered).
4. The version specified in paket.lock with package from the "official" feed preferred(WIP/fallback).

I don't think this is the feature that I'm looking for or one that I need. It does not solve the problem in the issue that I posted unless your answer is (4.) which it does not seem to be.

theimowski:
> Is there a specific slide or part of the video that you are referring to?

Yes, I specially mean [this slide](http://forki.github.io/PaketIntro/#/47) ("git dependencies use case - Bugfixing"), which is being descirbed 41:00 in video.
So the idea, as I imagine it, is to allow for development-time only replacement of source feed **for a specific package** with another source - be it nuget feed, or git repository, possibly accessible only on local development environment.
Regarding restore functionality, I think that this is by design that Paket pins to and then restores resolved package from only one specific source. The rationale here is to provide reproducible builds for everyone.

The new functionality, as I see it, would allow to break this rule, but only for temporary dev purpose of testing referenced package with new implementation.
I'm still not sure how to attach that to the current restore process, but currently there are 2 ideas that come to my head:

1. Upon restore, if any override is defined, recompute the package resolution with respect to the overrides (and for example save it in gitignored `paket.local.lock`) so that you can already test locally against new future version of referenced package.
2. Upon restore, if any override is defined, simply restore the highest version of package from overriden source (don't know yet how potential conflicts could be resolved)

theimowski:
@forki - do you want to comment on that? Is there something I'm missing?

forki:
@theimowski I think we are on the same page here. Paket should never restore from different location than what's given in lock file. There is only one exception and this temporary on dev machine during debuggig. Everything else will lead to doom.

theimowski:
For now, I've dropped the idea to allow different version of package from overriden source. 
See the integration test - i'm just manipulating LockFile internal package resolution to override source for a specifc package name.
This is probably a good start - wondering if different versions should be supported as well?

forki:
> It is only when restoring that paket restricts the package to a single feed(The feed it was installed from)

yes that is super important and one of the very basic pricinples of Paket. We restore exactly the stuff from lock file in order to make builds reproducible. If you don't want that then consider to run install during CI builds. we do that in many projects. We have one CI build which builds the pinned versions nd one that uses update to see if new versions o our dependencies would break us.

mlidbom:
>We restore exactly the stuff from lock file in order to make builds reproducible.

I agree 100% with the goal of making builds reproducible. 
However: c:\DEV_FEED is not reproducible. It is likely to be different on every machine that runs the build. I believe that the behavior I propose addresses this better than the one you propose. 

Let me explain what I mean. Consider a typical example from our everyday work: 
4 Developer Da-Dd collaborate on a large new feature for package Pa that is needed by the 4 solutions Sa-Sd. In order to do this they bump Pa, build a local package and update Sa-Sd to the new version of Pa.

Now what happens when the new version of Pa is published?
Unless Da-Dd all remember to remove the override and/or remove the package from their "override feed", these developers will each build Sa-Sd against the wrong Pa package. Each with it's own essentially random content depending on when they happened to build their Pa package last.

With my proposed behavior Da-Dc will all build Sa-Sd against the published version of Pa, thereby minimizing the risk of confusing differences between workstations and CI.

And of course, on the CI server c:\DEV_FEED does not exist so the CI server will always use only the published packages giving us fully reproducible builds on the CI server.

**This setup gives us fully reproducible builds in CI and the closest thing I know of on workstations. All without having to ever run the risk of developers forgetting to add or remove any manual overrides on their workstations.**

What's not to like? In what scenario would this behavior be less desirable than the one you propose?

mlidbom:
>consider to run install during CI builds

Problems with that. 

- Builds would still fail for other developers unless they change version controlled files by running install. 
- Install sometimes changes files in a way that breaks builds. Examples we've run into:
 - Two different nuget packages both install versions of type definitions for typescripts from definitely typed. One must be removed from the project for it to build. Running install will restore both files to the csproj and build will fail.
 - The latest version of a package containing less files has a bug in the less that we patch manually locally. Running install will overwrite the local changes and the build will fail.

matthid:
> What's not to like? In what scenario would this behavior be less desirable than the one you propose?

From your initial request:

> When updating-to/installing a package from c:\DEV_FEED it would be entered as installed from http://OFFICIAL_FEED in paket.lock

I think the main problem with this is that you now have a lockfile in an invalid state (as the package doesn't actually exist). With overriding you don't have this problem. The only thing your developers need to change in your proposed workflow is to don't update versions (so they actually need to do less work). In fact I would even go further and say that if a version is in the override feed that doesn't actually exist paket should trigger a warning and remove it from the local override folder. To prevent this kind of workflow. If a new version is released users will switch to it automatically with the next update. 

As a developer I don't want to bump the version again just to be able to continue my debugging session with a modified package. Just because a new package was released and paket is picking it up behind the scenes.

> Builds would still fail for other developers unless they change version controlled files by running install.

Not sure what you mean, in your proposed workflow they will potentially fail because of missing nuget packages.

> Install sometimes changes files in a way that breaks builds. Examples we've run into: 

That sounds like either errors in packages (-> redeploy a fixed version) or paket (-> report a bug / feature request).
There is also `content: None` to exclude files (maybe handle one package manually as a workaround for now).

Just some thoughts on this...

theimowski:
A couple of points:

> However: c:\DEV_FEED is not reproducible.

That's right, but keep in mind we want to advise to gitignore file with overrides (`paket.local`)

> Unless Da-Dd all remember to remove the override and/or remove the package from their "override feed", these developers will each build Sa-Sd against the wrong Pa package. Each with it's own essentially random content depending on when they happened to build their Pa package last.

IMO, the concept of "locking" dependencies in lock file aims to prevent such cases - I like the idea that you have to be explicit about version when pushing to remote, but still be able to break this rule for debugging/ development purposes

mlidbom:
@theimowski @matthid @forki 

OK so it seems pretty obvious that we are not going to agree on this. Most likely because of different needs due to different current situations and past experiences. Most users of paket/nuget do not produce and consume 100+ of their own packages in 30+ solutions/repositories.

But we have tried several workflows over the years and this one has proven to be, by far, the best way of working for us. We've been using it for years with little friction(Besides nuget created merge hell). I know from **very painful personal experience what happens in my organisation when we try the flows that you promote.** Using them is not an option that works for us.

I sincerely hope that you can agree that not everyone has the same needs. That when I find the time to make a pull request that implements the behavior we need you will not reject it out of hand for being "wrong".

pms1969:
I think as paket gets more popular, demand for this feature (or something very similar) will become higher.  We need something that allows us to add an additional source for discovering new "developer" versions of packages, and in our case a "buyer beware" attitude is ok.  You could always caveat the option/feature with a warning message "there be dragons here!", and if it screws up a build, no one could claim that they haven't been warned.

In our case, lock files aren't committed to the vcs.  The reasoning is unimportant, it's just how we have it, and even if we did commit them, it would make no difference on our build servers. That's because, in our environment, we control all the 3rd party dependencies, by hosting our own nuget server.  And of the 130 solutions that we have, producing however many consumable packages, we always want to be at the latest available version.  It's historic, and a starting point for us to reduce the number of solutions, and packages, but it's what we have.  We control the upgrade of our 3rd party dependencies, and when they are upgraded, they too must be upgraded in all our solutions.  There's no point for us in having one solution on version 7.0 of newtonsoft.json, and another on 8.0.  Although we have 130+ solutions, it is all part of 1 monolithic system.

For me, being able to control an additional source from the command line is preferred, but not an immovable requirement.  Would it be sensible/feasable to allow updating from this/these additional sources, but not changing the lockfile to reflect the additional sources version of those files.  If code then gets committed to the vcs that relies on different versions of the packages, then as I said before - "Buyer beware - you were warned".  I appreciate it's not a "pure" paket workflow, but there is demand for it, and not having the feature, may well cause some headaches for some, and lead to either them implementing some god awful hacky work around, or force them to look elsewhere for a solution, which I think would be detrimental all round.

src/Paket.Core/LocalFile.fs
```diff
@@ -0,0 +1,105 @@
...
+        | DevNugetSourceOverride (p,s) -> 
+            let groups =
+                lockFile.Groups
+                |> Map.map (fun _ g -> { g with Resolution = overrideResolution (p,s) g.Resolution } )
+            LockFile(lockFile.FileName, groups)
+        | DevGitSourceOverride   (p,s) ->
+            let owner,branch,project,cloneUrl,buildCommand,operatingSystemRestriction,packagePath = 
+                Git.Handling.extractUrlParts s
+            let restriction = VersionRestriction.Concrete (defaultArg branch "master")
+            let sha =
```
theimowski: @forki to overwrite LockFile model, I need a sha of git commit - is this the right way to do this? looks like it's working for a remote git repository

theimowski:
I managed to pass an integration test with overriden remote git repo - any feedback is welcome!

src/Paket.Core/LocalFile.fs
```diff
@@ -0,0 +1,105 @@
...
+                | None   ->
+                    failwith "paket.local: only git repositories as NuGet source (with 'Packages: ...') are currently supported"
+
+            let source groupName =
+                let root = ""
+                let fullPath = remoteFile.ComputeFilePath(root,groupName, packagesPath)
+                let relative = (createRelativePath root fullPath).Replace("\\","/")        
+                LocalNuGet(relative, None)
+
+            let groups =
```
theimowski: how to deal with groups? for now all groups are affected by the override

Vilmir:
Sorry for not really helping the pull request, but my team would benefit from this feature, we highly appreciate the effort and I wanted to give a :+1:

theimowski:
I really appreciate that. Would be cool if you shared how you imagine the workflow to look like. In is "first version" we plan to make it following:
* add override to paket.local
* commit to dependent git repo / update local nupkg file
* paket restore in depending project
* restore overrides the official dep with local one for debugging/ testing
* when done, release new version of dependent package and Paket update to that version in depending project

Vilmir:
Sure!
We have a huge codebase that we have started splitting with smaller services. The services are delivered as NuGet packages into our legacy codebase and have their own development life cycle.
This implies that services can be developed and tested in isolation with a high level of confidence.
As this process is new for our dev team, it is still hard/too expensive for them to get this level of confidence in isolation.
We have realized recently that they spent lots of time integrating prerelease packages several time per day manually. Sometimes they even commit them by mistake into the repository.

theimowski:
TODO list for 1st iteration:

- [ ] override nuget with local directory
- [ ] override nuget with local git repo with specified Packages 
- [ ] override remote git repo with local git repo

src/Paket.Core/LocalFile.fs
theimowski: will support main group only for now

theimowski:
I think this is ready for initial review - still need to see why CI fail on new integration tests (they seem to pass on my machine)

forki:
this is just beyond awesome!

geil

theimowski:
thanks mate! I think it broke Travis CI Integration tests - I started fixing that, but didn't finish yet. [here](https://github.com/fsprojects/Paket/blob/v3/build.fsx#L190) is a line that we'd probably want to remove - I did it for narrowing down integration tests that were run

PR #1638, opened 2016-04-23
make exit code 0 when help is explicitly requested

baronfel:
This just patches a tiny inconsistency.  When --help is passed for sub-commands the exit code is zero, but when it is passed for the main paket exe we get 1.

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -354,6 +354,9 @@ let main() =
 
             handler command args
             ()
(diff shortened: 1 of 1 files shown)


baronfel:
The build failure here looks like something I'm seeing consistently across several of these pending PRs, I think there's something wrong in paket right now that's independent of these PRs

forki:
We have build issues after nunit 3 conversion. It's a know problem. I hope we solve that soon.

PR #1643, opened 2016-04-24
fix debuggerdisplay commands that were broken and add comments

baronfel:
Found a few broken debuggerdisplays while working on the script generation feature so I thought I'd fix them up.
Also added a few comments to some domain types are a result of me asking what the particular type was supposed to represent.


smoothdeveloper:
@baronfel can you make PR on v3 branch instead?
@forki could you merge those recent PR in v3 branch?

Thanks!

Issue #1648, opened 2016-04-27
Allow Pack to pin only project references


jam40jeff:
### Description

Currently, the pack command allows for specifying `lock-dependencies` to force all generated dependencies to be specific (picked up from the `paket.lock` file) rather than use the version specification from the `paket.dependencies` file.

However, the version specification of generated project reference dependencies is also controlled by this setting.  If `lock-dependencies` is specified, the project reference versions are pinned (=), but if it is not specified, they are minimums (>=).  A new setting could be introduced (`pinned-project-references`) to allow for the project references to be pinned while still using the version specification from `paket.dependencies` for any external dependencies.
### Repro steps

This is a feature request.
### Expected behavior

The two settings `lock-dependencies` and `pinned-project-references` would work together as follows:

Neither `pinned-project-references` nor `lock-dependencies` specified:
External dependencies: Version specification picked up from `paket.dependencies`.
Project references: Version specification is minimum (>=).

`pinned-project-references` is specified, but `lock-dependencies` is not specified:
External dependencies: Version specification picked up from `paket.dependencies`.
Project references: Version specification is pinned (=).

`lock-dependencies` is specified (`pinned-project-references` does not matter):
External dependencies: Version specification picked up from `paket.lock` (pinned).
Project references: Version specification is pinned (=).
### Actual behavior

Either external dependencies and project references must both be pinned, or neither be pinned.
### Known workarounds

There is no known workaround.
### Related information

N/A

jam40jeff:
Related to #1360.

If `lock-dependencies` is removed, then whether or not project references are pinned would only be affected by this setting.

I believe it still makes sense even if `lock-dependencies` is removed, because dependencies generated for project references should be automatic and not have to be specified in the nuspec file, but this setting would be the only way to determine whether the behavior is pinned (=) or minimum (>=).

jam40jeff:
I decided to change this option's name to `pin-project-references` to match the verb nature of the other options.

inosik:
I would just advocate KISS here and simply use `paket.template`, `CURRENTVERSION` and the `specific-version` parameter.

One other reason I think `lock-dependencies` is not very useful, is because my build script calls `paket pack` in the root directory. Now, if `lock-dependencies` (or the new `pin-project-references`) would be set,  **every** generated package gets its dependencies locked. In most cases (or actually, always), this is not what I want. Because otherwise, they wouldn't even be separate packages.

jam40jeff:
[...]

PR #1649, opened 2016-04-27
Allow Pack to pin only project references (Pull Request)

jam40jeff:
See #1648.

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -133,7 +133,7 @@ let addFile (source : string) (target : string) (templateFile : TemplateFile) =
     | IncompleteTemplate -> 
         failwith (sprintf "You should only try and add files to template files with complete metadata.%sFile: %s" Environment.NewLine templateFile.FileName)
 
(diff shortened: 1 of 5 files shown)


tsibelman:
Hi @jam40jeff It will be very helfull for as if you could finish this feature.

jam40jeff:
I have completed this feature and am using it in my own custom build of Paket.  However, I was unable to get the pull request approved.  Adding another voice of support to this issue can't hurt though.  I still believe it is a very valuable option and would like to see it included in Paket so I don't have to keep synchronizing my custom build with each release.

tsibelman:
Why this pull request was not aproved ? I do not see anything here, did you got some feedback by some other chanell ?

jam40jeff:
The discussion on #1648 is the reasoning for the pull request not being approved.  I implemented the simple switch which allows you to use exact versioning for project reference dependencies rather than minimum versioning.  The rest of that discussion was about how to make it so other schemes could be used, but I still think that would just complicate things without adding much benefit.

The two options (exact with this switch, minimum without it) are the only two I have needed.  I use minimum when I am guaranteeing backwards compatibility between projects within a solution and exact when I am not.

tsibelman:
I read the disscusion but fail to see anyone objecting to this change, @forki can you tell us if you see somee problem with this ?

forki:
can you please rebase this. Then I will review. Thx

forki:
thanks!

tsibelman:
Thank you guys, you made my day.

PR #1667, opened 2016-05-08
Blacklisting install.xdt and uninstall.xdt files

wallymathieu:
If we look at the code for how other transformation files are handled, then we see some blacklisting logic in paket. The xml transformation files mentioned in #614 are not present in the blacklisting though.


forki:
thanks for taking care of this!

wallymathieu:
No problem. It looks like there are some failing tests after this merge. I'll see if I can fix some of them when I have time.

Issue #1663, opened 2016-05-05
Framework resolution strategy for Google.Apis.Oauth2.v2


sergey-tihon:
### Description

Cannot use `Google.Apis.Oauth2.v2` with `Paket` for `Net45` project
Paket does not setup references properly in this case

### Expected behavior

![image](https://cloud.githubusercontent.com/assets/1197905/15045191/f4704a88-12e1-11e6-8850-d10e8a6e8164.png)

```xml
<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="BouncyCastle" version="1.7.0" targetFramework="net45" />
  <package id="Google.Apis" version="1.11.1" targetFramework="net45" />
  <package id="Google.Apis.Auth" version="1.11.1" targetFramework="net45" />
  <package id="Google.Apis.Core" version="1.11.1" targetFramework="net45" />
  <package id="Google.Apis.Oauth2.v2" version="1.12.0.454" targetFramework="net45" />
  <package id="log4net" version="2.0.3" targetFramework="net45" />
  <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net45" />
  <package id="Zlib.Portable.Signed" version="1.11.0" targetFramework="net45" />
</packages>
```

### Actual behavior

**paket.dependencies**
```
source https://nuget.org/api/v2

nuget Google.Apis.Oauth2.v2
```

**paket.lock**
```
NUGET
  remote: https://www.nuget.org/api/v2
  specs:
    BouncyCastle (1.8.1)
    Google.Apis (1.10) - framework: portable-net40+sl5+win8+wp8+wpa81
      Google.Apis.Core (>= 1.10)
      log4net (>= 2.0.3) - framework: >= net40
      Zlib.Portable.Signed (>= 1.11)
    Google.Apis.Auth (1.11.1) - framework: portable-net45+win8+wp8+wpa81
      BouncyCastle (>= 1.7) - framework: >= net45
      Google.Apis.Core (>= 1.11.1)
    Google.Apis.Core (1.11.1) - framework: portable-net40+sl5+win8+wp8+wpa81, portable-net45+win8+wp8+wpa81
      Microsoft.Bcl (>= 1.1.10) - framework: net10, net11, net20, net30, net35, net40, net40-full
      Microsoft.Bcl.Async (>= 1.0.168) - framework: net10, net11, net20, net30, net35, net40, net40-full
[...]

PR #1674, opened 2016-05-10
Paket should import build targets from packages in dependency groups

sergey-tihon:
Not sure it is directly related to #1663
But after `paket.exe update` with version `2.64.2`, paket removed by build targets from project file
image

### DIFF
--- a/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
@@ -300,8 +300,8 @@ let ``#1505 should install conditionals``() =
 [<Test>]
 let ``#1663 should install google apis``() = 
     install "i001663-google-apis" |> ignore
-    let newFile = Path.Combine(scenarioTempPath "i001505-conditionals","MyClassLibrary","MyClassLibrary","MyClassLibrary.csproj")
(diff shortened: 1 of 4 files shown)


inosik:
The problem is the group name "Build". Other names seem to work fine.

Seems to be the same kind of problem as we had with "Lib"-named packages.

sergey-tihon:
@inosik How group name affect Paket behavior?

inosik:
It expects `.targets` and `.props` files to be directly in `packages/Foo/build/` or in a framework specific sub-directory.

Now, if a group is named "Build", you get a path like `packages/build/Foo/build/Foo.targets`. I assume Paket checks only if `Foo.targets` is in `packages/build/` or in `packages/build/{framework}/`, and if it isn't, it considers it as ignorable.

I tried installing the package from your screenshot and Paket installed it correctly when the group wasn't named "Build". With "Build", it didn't import the `.targets` file.

sergey-tihon:
Just wonder how it worked before )

forki:
@inosik so it's a string trimming issue., right?

inosik:
I think so. We had something similar with packages which ended with "Lib", but I can't find the issue right now.

forki:
@sergey-tihon  if we have a repro I will try to fix it

sergey-tihon:
@forki This PR is repro

sergey-tihon:
It works!
Thanks again @forki !

PR #1683, opened 2016-05-11
Bye bye project.json support

forki:
fuck


forki:
damn it. I should have committed a project.lock.json file. then this would be much more on the side of deletion. Missed opportunity..

PR #1694, opened 2016-05-18
Fixed some bugs in NupkgWriter

vik-borisov:
- nuspec file
if the attribute "version" is empty it will not be written.
- coreProp file
fixed creator element format
fixed version element value (removed "Some()")
- removed whitespaces


forki:
wow. thanks for taking care of this.

PR #1697, opened 2016-05-21
Add new option --max-file-age

magicmonty:
for temporarily preventing a delay while checking the new file version.

**this is a reapply of PR #1617, as that one was accidentally rebased**

Use case:
The internet on my workplace is occasionally very slow. My build scripts call the bootstrapper on every call (in case paket.exe is missing). This leads usually to a hit to "GetNewVersion", which sometimes slows the build process down, as I build very often via command line.
With this command line parameter it is possible to delay this check for a new version so it will be checked for example only every 4 hours. If the paket.exe exists and the last write time of the paket.exe is not older than the given threshold, the bootstrapper will not check the version, so the network hit is not there anymore.

### DIFF
--- a/docs/content/bootstrapper.md
+++ b/docs/content/bootstrapper.md
@@ -1,10 +1,10 @@
 # paket.bootstrapper.exe
 
-Downloads the latest stable paket.exe. By default, the bootstrapper caches downloaded versions of paket.exe for the current user across all projects. If the requested version is not present in the cache, it is downloaded from github.com. If the download from github.com fails it tries to download the version from nuget.org instead. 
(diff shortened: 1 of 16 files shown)


smoothdeveloper:
Great idea @magicmonty, seems you have good amount of tests as well.

Can we maybe add a note in the docs of the website with recommended value?

magicmonty:
thanks for the hint, just added a line in the bootstrapper docs

magicmonty:
BTW, why is the build failing on Travis? This seems to have nothing to do with my changes.

forki:
I merged this into v3 since we already closed v2 for new features. 
Thank you very much for this contribution.

magicmonty:
Thanks for the hard work too. Paket is awesome

PR #1704, opened 2016-05-25
Fix small typo and use constants

theimowski:
"The group was not found" ==> "The group was found"


forki:
Thx a lot. Details are important.

theimowski:
:+1:

Issue #1600, opened 2016-04-12
Remove command - remove group when removing last dependency


theimowski:
### Description

When package XXX is the only dependency in group YYY, then  running `paket remove nuget XXX group YYY` should remove the group as well

### Repro steps

see #1596 

### Actual behavior

Currently, the group becomes orphaned

forki:
avaliable in v3

PR #1706, opened 2016-05-26
Feature/#1600 remove empty groups

shortsn:
Hi, this implements the feature requested in #1600.


forki:
That looks good. I assume the error is unrelated. Will taker deeper look on Monday

shortsn:
Ok, cool. Tell if i have to change something.

forki:
Thank you very much.
I merged this into v3 since we already closed v2 for new features.
So this will be available in the next prerelease of v3.

PR #1709, opened 2016-05-28
Mark packages with 'paket.overriden' file

theimowski:
This is to allow revert from override package to the original one


forki:
thanks buddy

Issue #1682, opened 2016-05-11
Git dependencies - build command cache invalidation


theimowski:
### Description

When git repo dependency is restored and target branch already checked out inside `paket-files`, the build command is not invoked even though there's missing nupkg 

### Repro steps

Dependencies file:

```
git https://github.com/theimowski/Argu.git paket_src_override build:"build.cmd NuGet", Packages: /bin/

nuget Argu
```

2. `$ paket install`

3. remove `paket-files\github.com\theimowski\Argu\bin` directory

4. `$ paket restore`

### Expected behavior

`build.cmd NuGet` should be invoked once more, because the target `bin` directory is missing

### Actual behavior

Paket fails with:

```
Paket version 3.0.0.0
found: C:\sandbox\paket\paket.dependencies
Parsing C:\sandbox\paket\paket.dependencies
19a8144eb45d052b5c1a7402baa5a61e38f0a66d
C:\sandbox\paket\paket-files\github.com\theimowski\Argu is already up-to-date.
Paket failed with:
        Could not find a part of the path 'C:\sandbox\paket\paket-files\github.c
om\theimowski\Argu\bin'.
StackTrace:
     at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileSystemEnumerableIterator`1.CommonInit()
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String origina
lUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`
1 resultHandler, Boolean checkHost)
   at System.IO.Directory.EnumerateFiles(String path, String searchPattern, Sear
chOption searchOption)
   at Paket.NuGetV2.findLocalPackage(String directory, PackageName packageName,
SemVerInfo version) in C:\github\Paket\src\Paket.Core\NuGetV2.fs:line 301
   at Paket.RestoreProcess.ExtractPackage@53-16.Invoke(Unit unitVar) in C:\githu
b\Paket\src\Paket.Core\RestoreProcess.fs:line 73
   at Microsoft.FSharp.Control.AsyncBuilderImpl.callA@851.Invoke(AsyncParams`1 a
rgs)

....
```

### Known workarounds

* remove `paket-files\github.com\theimowski\Argu` directory - this forces Paket to redownload git repo
* run restore with `--force` fla

theimowski:
Fixed in #1712

PR #1712, opened 2016-06-01
[WIP] Fix for #1682

theimowski:


src/Paket.Core/LockFile.fs
```diff
@@ -179,6 +179,10 @@ module LockFileSerializer =
                     | None -> ()
                     | Some command -> yield "      build: " + command
 
+                    match file.PackagePath with
+                    | None -> ()
+                    | Some path -> yield "      path: " + path
+
```
theimowski: @forki is it ok to add this info to the lock file? It's being used in `RemoteDownload` module

forki:
thanks a lot!

src/Paket.Core/LockFile.fs
forki: do you parse it again?

theimowski: Well I don't think it was being parsed before. Only the build command was

forki: but now we do? or is it just for reference?

theimowski: ah sorry didn't understand the question. Yes It's parsed [here](https://github.com/fsprojects/Paket/blob/v3/src/Paket.Core/LockFile.fs#L336-L341) so that it can be later read [here](https://github.com/fsprojects/Paket/blob/v3/src/Paket.Core/RemoteDownload.fs#L291)

theimowski:
cheers!

PR #1713, opened 2016-06-01
#1627 throw on invalid group names

shortsn:
Throws an ArgumentException on prohibited group names -> requested in #1627
- lib
- runtimes


forki:
Since this is breaking I merged it into v3.
Thanks a lot for this contribution!

PR #1717, opened 2016-06-04
generate-include-scripts: fix edge case of nuget package with mscorlib dep

smoothdeveloper:
FSI can't have that reference added more than once and it doesn't need it AFAIK

With this change, we do not generate `#r "mscorlib"` in F# scripts anymore.


forki:
Thx

PR #1728, opened 2016-06-15
Don't allow empty string as authors and description metadata

inosik:
Paket produces invalid metadata for packages, which only have `type project` in their `paket.template` file.

Those packages look ok, but cause problems further down the road, i. e. when trying to upload them or when using test adapters installed via Paket. Probably when trying to install them with NuGet, too, because the non-descriptive error message in VS comes from NuGet.Core.

/cc @biehlermi


forki:
could you pelase take a look at the broken test? thanks

inosik:
Missed that one, should pass now.

inosik:
An integration test for `paket update` timed out on Mono, not related to this.

forki:
Thanks. This is now in v 3.1

PR #1733, opened 2016-06-16
Ignore case in aliases dll names

### LINKED ISSUE #1732: Alias dll names must be specified as lower case
### Description

Contrary to what the [documentation specifies](https://fsprojects.github.io/Paket/references-files.html#Library-aliases) library aliases dll names must be lowercase, it mean that the example in the documentation doesn't work.

The correct version being :

```
PackageA
  alias a1.dll Name2,Name3
  alias a2.dll MyAlias1
Dapper
NUnit
```

### Repro steps
1. Apply what the documentation specifies :

   ```
   PackageA
     alias A1.dll Name2,Name3
     alias A2.dll MyAlias1
   Dapper
   NUnit
   ```

2. Run `paket install`

### Expected behavior

Aliases are created when the exact dll name is specified.

### Actual behavior

Doesn't create aliases in the project file.

### Known workarounds

Please provide a description of any known workarounds.

### Bla bla

Ideally the case would be completely ignored.

Also the current code matches on substrings & that might be a litle bit surprising (`alias c foo` will alias any dll containing the letter c)

[ProjectFile.cs](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/ProjectFile.fs#L625) :

```fsharp
                    let l = lib.ToLower()
                    let aliases =
                        aliases
                        |> Seq.tryFind (fun kv -> l.Contains(kv.Key))
                        |> Option.map (fun kv -> kv.Value)
```

i'll send a PR with a fix

vbfox:
The previous code was expecting the passed in name to be a lowercase substring of the dll file name. It's now expected to be the full name (not a substring) but allow the case to differ.

Fixes #1732


vbfox:
Oops binary files, i'll remove & rebase.

vbfox:
Actually let's reopen this PR, I don't know what's the policy @forki 

The files are the same as the previous test so don't really take space in git, but I don't know why the files were present in the other one. To accelerate the tests ? to ensure that it's working if nuget remove the old versions ?

(I have no problem changing the PR to not have them)

forki:
they are there to keep things stable.

forki:
thanks!

PR #1734, opened 2016-06-16
Add condition to AfterBuild target to unbreak nCrunch

thinkbeforecoding:
NCrunch doesn't run well with the target that is generated with some projects.
Adding a condition to the target disable it when running under ncrunch, avoiding to break compilation.


smoothdeveloper:
@thinkbeforecoding could you add a failing (without the change) integration test as well?

Thanks!

forki:
there is actually already a broken test. so @thinkbeforecoding can you please try to fix that one?

thinkbeforecoding:
I'm discovering the IntegrationTests so I'll try to see where it breaks

forki:
I'm already on it

thinkbeforecoding:
I think it's in a test where the condition has been added and the test verify equality with exising project. The project should then be updated.

PR #1735, opened 2016-06-16
fix #1731, config now removes trailings slashes

### LINKED ISSUE #1731: Credentials do not work if ending with a slash
If the URL using `paket config add-credentials` ends with a slash, these credentials are ignored (without a warning). This is especially misleading, since the source entry in the `paket.dependencies` __can__ end with a slash. So when simply copying the entries from the `paket.dependencies` into the command line, it simply does not work.

### Steps to reproduce

- Create a project
- create a `paket.dependencies` pointing to a private feed and a dependency
- run  `paket config add-credentials` using the feed URL including a trailing slash
- run `paket install` (potentially cleaning the cache beforehand)

### expected behaviour

- command finishes successfully

### observed behaviour

- The command fails (401), Fiddler shows that all requests to the feed happen unauthenticated

### notes

Reproduced in 3.0.0 and the last 2.x version. We use a Nexus a private feed, but I don't suppose that this has an impact, because the error happens before any request is created.

### Suggestion

Either:
1) add-credentials should fail with an error upon trailing slashes (ugly)
2) add-credentials should silently drop any trailing slashes (perhaps with a warning)
3) trailing slashes in paket.config are removed upon reading that file (still would allow two entries for the same source, one with and one without trailing slash)
4) Document the constraint on the webpage

I prefer 2), since it would also keep the paket.config clean. Additionally, reading the paket.config should at least trigger a warning (or better: fail) if trailing slashes are encountered (from older versions)

pauxus:
Simple solution for #1731. 

not (yet) implemented a check for defective paket.config

### DIFF
--- a/src/Paket.Core/ConfigFile.fs
+++ b/src/Paket.Core/ConfigFile.fs
@@ -202,4 +202,4 @@ let askAndAddAuth (source : string) (username : string) =
             username
 
     let password = readPassword "Password: "
-    AddCredentials (source, username, password)
+    AddCredentials (source.TrimEnd [|'/'|], username, password)


forki:
LGTM thanks

PR #1738, opened 2016-06-18
Make sure Guid in project reference is parsed well

Krzysztof-Cieslak:
Follow up on #1729. Previous PR hasn't handled all cases. Added tests to make sure everything is OK this time.


forki:
thanks buddy

Issue #1722, opened 2016-06-14
Feature Request: Add Batch option for add-credentials


pauxus:
Currently, add-credentials always asks credentials from the console.

In our developer machine setup, we automatically configure configuration files for various build systems / tools like maven, gradle, npm, nuget all using the same user credentials.

We do this by reading the credentials from the user once and after that executing config commands of the tools in sequence (npm config auth, nuget source...)

However, Paket does not allow to specify username and password directly on the command line, which makes it a lot harder to automatically configure it using that system.

I suggest providing a --batch flag:

paket config add-credentials <url> --batch <username> <password>

alternatively: 

paket config add-credentials <url> --username <username> --password <password>

forki:
Yes this sounds doable.
I'm making this up-for-grabs

PR #1742, opened 2016-06-20
revert accidential changes

pauxus:
This reverts accidential changes to project file (were part of my testing setup for #1722).

Sorry for that...


forki:
that's not really important. I overwrite this basically evry day. thanks anyways.

PR #1754, opened 2016-06-24
Duplicate Paket.SemVer.Parse into PublicAPI.fs

amazingant:
As per the discussion on 79b57ac8177566c5f06ba595636f64c91583a7b8, only `PublicAPI.fs` is treated as a public API; added a function in this file that calls `Paket.SemVer.Parse` and has the same compiled signature that `Paket.SemVer.Parse` had prior to 79b57ac8177566c5f06ba595636f64c91583a7b8.


forki:
thanks!

PR #1764, opened 2016-06-27
Added handling for cache not being accessible

### LINKED ISSUE #1758: Using network cache with invalid credentials fails restore
I have a network cache declared that I am not on the domain for. Periodically my credentials age out, and my build fails like so:
```
  ==> Trying again
Downloading Microsoft.AspNet.WebPages 3.1.1
Something went wrong while downloading Microsoft.Bcl 1.1.10
Message: The user name or password is incorrect.
```
It would be preferable for it to just log a warning, and restore from the main feed in that situation.

xavierzwirtz:
Fixes #1758

I moved the directory creation to when the cache is loaded from the dependencies file. It tries to create the directory if it doesn't exist, and doesn't include the cache in the feed list or cache list if it fails.

### DIFF
--- a/src/Paket.Core/Cache.fs
+++ b/src/Paket.Core/Cache.fs
@@ -45,3 +45,11 @@ type Cache =
 
         settings
              
+    member this.IsAccessible =
+        try
+            let targetFolder = DirectoryInfo(this.Location)
+            if not targetFolder.Exists then
+                targetFolder.Create()
+            true
+        with
+        | _ -> false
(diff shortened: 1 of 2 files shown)


xavierzwirtz:
This broke the `Paket.DependenciesFile.ParserSpecs.should read config with caches` test. It is no longer returning any caches that are not accessible.

forki:
yeah we should not do this during parsing.

xavierzwirtz:
Where would be the best place to do it? It feels like it needs to be done for every case where the cache would be accessed, the earlier in the process the better. Otherwise, in cases where there are a lot of dependencies there will be a slow call out to the network for every restore. What about an optional argument to `DependenciesFile.ReadFromFile`, defaulting to true, that ignores caches that are not accessible?

forki:
I think it should be done differently. We "just" need to make the cache calls more robust. we should not filter them out.

forki:
Maybe additionally we can now blacklist caches during the lifetime of one
paket call. But I think it was important to first make the calls safe

xavierzwirtz:
Thanks! This is already a lot better. I am able to run `paket restore` now without being vpn'd into the main office.

PR #1774, opened 2016-06-29
Add note about shell execution of powershell

baronfel:
At work our build process uses a powershell script instead of a cmd script, and it took a bit of tinkering to figure out how to invoke that script as the build command of a git dependency.  This change just adds a note to the docs so that others hopefully won't have to look as hard!


forki:
thanks

PR #1775, opened 2016-06-30
Make NuGet OData search case-insensitive

trondd:
The original code would use the package name as the search argument, which casued the following URL to return no packages:

https://www.nuget.org/api/v2/Packages?$filter=Id%20eq%20%27Log4Net%27


forki:
thanks

trondd:
Did you notice that the Mono build was red?

forki:
Yes. That's unrelated, right? Need to investigate tomorrow.

PR #1781, opened 2016-07-01
paket config should be able to run everywhere

pauxus:
Currently, `paket config` runs into an error when not executed from within a project. However, a project is not necessary for this, since the command only accesses a user configuration file. This PR simply creates a dummy `Dependencies` instance to fix this.

This is especially usefull in combination with #1722 to allow an automatic setup of a developer machine (enterprise environment)


forki:
thanks!

PR #1786, opened 2016-07-02
Small refactor of generate-include-scripts internal implementation.

smoothdeveloper:
Now scripts are being constructed as a list of ScriptPiece (a simple DU) and then rendered into string written in the resulting file.

This is needed so I have better infrastructure to work on a later feature related to include scripts.


smoothdeveloper:
Effortless refactoring, thanks to acyclic dependency being norm in F# codebase and awesomeness of VFPT :)

@matthid / @baronfel if you don't mind a quick look at this PR?

I'm going to use that to generate additional script for each group.

All allegedly impacted tests are passing locally.

forki:
thanks!

PR #1791, opened 2016-07-05
Fix detection of framework version with spaces

### LINKED ISSUE #1790: NuGet packages with lib folder `.NetFramework 3.5` can't be referenced
### Description

When adding a NuGet reference to a package containing only a folder named `.NetFramework 3.5` in the `lib` directory (e.g. [MarkdownDeep.NET](https://www.nuget.org/packages/MarkdownDeep.NET/)) the project file does not get a reference added to it.

### Repro steps

1. Add the package MarkdownDeep.NET (or any other package with folder structure `lib/.NetFramework 3.5`).

### Expected behavior

The project should have a reference added to the package, with the condition `$(TargetFrameworkVersion) == 'v3.5'`

I would assume that the folder named `.NetFramework 3.5` should be treated in the same way as a folder named `net35`.

### Actual behavior

No reference is added to the project.

Reproduction of what happens when the package is added to an empty scaffolded project:
https://github.com/danielrbradley/PaketMarkdownDeepRepro/commit/3c5f37e35fd468db984731d636ceb84a68de37e5

### Workaround

Rebuild the original NuGet package using a fixed folder name, and host privately.

### Related information 

This has been tested when adding the package on both C# and F# projects.

This was found while running an automated conversion from NuGet.

danielrbradley:
Fixes #1790

Fixes compatibility with packages such as MarkdownDeep.NET which have the
internal folder structure of

    ./lib/.NetFramework 3.5/MarkdownDeep.dll

Resolve this issue by adding an alias to remove spaces in the framework string.

### DIFF
--- a/src/Paket.Core/FrameworkHandling.fs
+++ b/src/Paket.Core/FrameworkHandling.fs
@@ -98,7 +98,8 @@ module KnownAliases =
          ".netportable", "portable"
          "netportable", "portable"
          "0.0", ""
-         ".", "" ]
+         ".", ""
+         " ", "" ]
         |> List.map (fun (p,r) -> p.ToLower(),r.ToLower())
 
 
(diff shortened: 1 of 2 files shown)


forki:
thanks!

Issue #1794, opened 2016-07-06
Convert from nuget: Local package source give false error


edbenson:
### Description

When you run convert-from-nuget , if you have a local nuget package source, you will see a false warning message "Local NuGet feed doesn't exist"

### Repro steps

Create a folder to contain local nuget sources, e.g. C:\LocalNugetSource\ 

Add this as nuget local package source; either in Visual Studio > Tools > NuGet package manager > Package Manager Settings > Package Sources, or by editing file: %AppData%\NuGet\NuGet.Config

```
  <packageSources>
  ...
    <add key="Local Package source" value="C:\LocalNugetSource\" />
  </packageSources>
```

now run 
convert-from-nuget

### Expected behavior

We expect no warnings in respect of the local nuget package source

### Actual behavior

Incorrect warning message: "Local NuGet feed doesn't exist: [folder]"

### Known workarounds

This is a warning only - you can safely ignore the message. But if you are new to Paket it's misleading.

### Related information 

* Operating system: Windows 7
* Branch: master
* .NET Runtime: Windows .NET 4.x

forki:
should be fixed

johannesvorster:
I'm using version 3.33.3 of paket.exe. When I run the convert-from-nuget command I get the 'Local NuGet feed doesn't exist' error, even though the folder exist.

forki:
Can you please give a bit more info? 

Cc @theimowski

johannesvorster:
Good day Steffen


Is it all right with you if I try to find the problem myself?


Regards


Johannes Vorster

E: johannes.vorster@outlook.com<mailto:elconi.johannes@gmail.com>
<http://za.linkedin.com/in/johannesv>


________________________________

forki:
sure ;-)

PR #1795, opened 2016-07-06
Fix issue #1794: Fix error with local sources and run convert-from-nuget

edbenson:
Simple fix: File.Exists should be Directory.Exists.

I have tested this locally and this fixes the error.


edbenson:
I would have liked to add a test but I could not find any tests for convert-from-nuget

forki:
cool. thanks

PR #1802, opened 2016-07-13
Fix incorrect CopyRuntimeDependencies.ProjectFile causing 'Could not find paket.dependencies'

vladnega:
The `CopyRuntimeDependencies.ProjectFile` member was incorrectly set to the `outputPath` variable, instead of the `projectFile` one. This caused the `Dependencies.Locate` function to search in the wrong folder for the `project.dependencies` file.


forki:
thanks!

Issue #1806, opened 2016-07-13
Pack localized assemblies.


JohanLarsson:
Automate generation of [this](https://docs.nuget.org/create/creating-localized-packages) from a template with `type project`

forki:
thanks for doing this

PR #1807, opened 2016-07-13
Pack localized assemblies.

JohanLarsson:
An attempt at #1806


src/Paket.Core/ProjectFile.fs
```diff
@@ -1337,6 +1338,18 @@ type ProjectFile with
 
     member this.FindReferencesFile() = this.FindCorrespondingFile Constants.ReferencesFile
 
+    member this.FindLocalizedLanguageNames() =
+        let pattern = @"Properties\\Resources.(?<language>\w+(-\w+)?).resx"
+        let includeattribute = "Include"
+        this.ProjectNode
+        |> getDescendants "EmbeddedResource"
+        |> List.filter (fun n -> n |> hasAttribute includeattribute)
```
vasily-kirichenko: why not point free?

JohanLarsson: What does it mean?

theimowski: means `(fun n -> n |> hasAttribute includeattribute)` is the same as `(hasAttribute includeattribute)`

JohanLarsson: Thanks.

forki:
Awesome! Only one important thing left: docs. ;-)

Could you please add a small bit about this auto-packing of satelitte assemblies?

src/Paket.Core/PackageMetaData.fs
```diff
@@ -167,17 +167,39 @@ let findDependencies (dependenciesFile : DependenciesFile) config platform (temp
...
+                    if File.Exists fileName then
+                        let satelliteTargetDir = Path.Combine(targetDir, language)
+                        yield (FileInfo(fileName), satelliteTargetDir)
+                    else
+                        failwithf "Did not find satellite assembly for (%s) try building and running pack again." language 
+            }
+
+        let template =
+            satelliteDlls 
+            |> Seq.fold (fun template dll -> addFile (fst dll).FullName (snd dll) template) template
```
smoothdeveloper: you could decompose dll here:

```fsharp
 |> Seq.fold (fun template (dllFile, targetDir) -> addFile dllFile.FullName targetDir template) template
```

src/Paket.Core/PackageMetaData.fs
```diff
@@ -167,17 +167,39 @@ let findDependencies (dependenciesFile : DependenciesFile) config platform (temp
...
+        let projects =                 
+            if includeReferencedProjects then 
+                project.GetAllInterProjectDependenciesWithoutProjectTemplates() 
+                |> Seq.toList 
+            else 
+                [ project ]
+
+        let satelliteDlls =
+            seq {
+            for project in projects do
```
smoothdeveloper: (minor/style) not sure of general convention, but CE block should be indented (or put `seq {` at end of previous line)

src/Paket.Core/PackageMetaData.fs
```diff
@@ -167,17 +167,39 @@ let findDependencies (dependenciesFile : DependenciesFile) config platform (temp
...
+                let satelliteAssemblyName = Path.GetFileNameWithoutExtension(project.GetAssemblyName()) + ".resources.dll"
+                let projectDir = Path.GetDirectoryName(Path.GetFullPath(project.FileName))
+                let outputDir = Path.Combine(projectDir, project.GetOutputDirectory config platform)
+                for language in project.FindLocalizedLanguageNames() do
+                    let fileName = Path.Combine(outputDir, language, satelliteAssemblyName)
+                    if File.Exists fileName then
+                        let satelliteTargetDir = Path.Combine(targetDir, language)
+                        yield (FileInfo(fileName), satelliteTargetDir)
```
smoothdeveloper: parens are uneeded:

`yield (FileInfo filename),  satelliteTargetDir`

maybe even check if remaining are needed

src/Paket.Core/ProjectFile.fs
```diff
@@ -1337,6 +1338,18 @@ type ProjectFile with
 
     member this.FindReferencesFile() = this.FindCorrespondingFile Constants.ReferencesFile
 
+    member this.FindLocalizedLanguageNames() =
+        let pattern = @"Properties\\Resources.(?<language>\w+(-\w+)?).resx"
+        let includeattribute = "Include"
+        this.ProjectNode
+        |> getDescendants "EmbeddedResource"
+        |> List.filter (hasAttribute includeattribute)
+        |> List.map (fun n -> n.Attributes.[includeattribute])
```
smoothdeveloper: maybe define tryGetAttribute

```fsharp
let tryGetAttributeValue name node = 
    if hasAttribute name node then
        Some node.Attributes.[name].Value
    else
        None
```

then you can replace List.filter/List.map/List.map with

```fsharp
|> List.choose (tryGetAttributeValue"Include")
```

src/Paket.Core/ProjectFile.fs
```diff
@@ -1337,6 +1338,18 @@ type ProjectFile with
...
+        |> List.map (fun a -> a.Value)
+        |> List.map (fun a -> Regex.Match(a, pattern, RegexOptions.ExplicitCapture))
```
smoothdeveloper: it feels like a helper local function to handle the regex logic (return `Some "the identified language"` or `None` with single function call) would avoid mixing list operation with multistep processing of regex (which is more imperative thing).

JohanLarsson:
@Forki updated with a mention in docs.

src/Paket.Core/ProjectFile.fs
JohanLarsson: Gah, I forgot about this refactoring.

src/Paket.Core/PackageMetaData.fs
JohanLarsson: I could probably remove the `FileInfo` as it is just an allocation as we have already checked `Exists`

src/Paket.Core/PackageMetaData.fs
JohanLarsson: Nice, I was stuck on that one for a while, used a fuzzying process.

forki:
mhmm looks like it's broken on mono. can you please take a look?

forki:
we also got a regression, so I reverted that change for now. see https://github.com/fsprojects/Paket/issues/1810

JohanLarsson:
I'll be mostly adj a couple of days. Checking as soon as I can.

PR #1814, opened 2016-07-16
Fix possible null ref when processing a vcxproj file

robertpi:


src/Paket.Core/ProjectFile.fs
```diff
@@ -1062,23 +1062,32 @@ module ProjectFile =
...
                     match getProperty "SolutionDir" project with
                     | Some slnDir -> normalizedPath.Replace("$(SolutionDir)",slnDir) 
                     | None -> normalizedPath.Replace("$(SolutionDir)", Directory.GetCurrentDirectory() + Path.DirectorySeparatorChar.ToString())
                 else normalizedPath
 
-            { Path =
-                if Path.IsPathRooted path then Path.GetFullPath path else 
-                let di = FileInfo(normalizePath project.FileName).Directory
-                Path.Combine(di.FullName,path) |> Path.GetFullPath
+            let optPath =
```
smoothdeveloper: maybe this can be replaced with

```fsharp
let path =
  node
  |> getAttribute "Include" 
  |> Option.map getNormalizedPath

match path with
| None -> ()
| Some path ->
  // make & yield the record
```

Also since this area of code seems to be tricky, I'd recommend pulling out the expression assigned to that `Path` record property to an intermediate variable.

smoothdeveloper:
@robertpi would you consider adding an integration test with few asserts on expected changes to vcxproj file:

https://github.com/fsprojects/Paket/tree/master/integrationtests/scenarios

adding unit test like you did is good but additional integration test will provide more security regarding overal outcome of running `paket restore` (or other command, up to you)

PR #1816, opened 2016-07-17
Pack localized assemblies.

### LINKED ISSUE #1810: Paket 3.7.0.0 broke paket pack
### Description

There is a regression in 3.7.0.0. Paket pack no longer works:

```
Paket version 3.7.0.0
Paket failed with:
        Did not find satellite assembly for (ontrols) try building and running pack again.
```

Sorry for not being able to provide more information at the moment, but I'll get back once I have an example that reproduces the problem

### Known workarounds

Use 3.6.1.0

JohanLarsson:
Second attempt at #1806. 

- Should fix #1810.
- Printing warning instead of throwing if satellite dll is not found. Added integration test for this.
- Better pattern for finding languages.
- Checking that matches for the regex is a culture name.

Thank you for all the help @smoothdeveloper!


integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -349,3 +349,19 @@ let ``#1596 pack works for reflected definition assemblies``() =
...
+let ``#1806 pack localized dll`` () =
+    let scenario = "i001806-pack-localized"
+    let rootPath = scenarioTempPath scenario
+    let outPath = Path.Combine(rootPath, "out")
+    let package = Path.Combine(outPath, "LocalizedLib.1.0.0.0.nupkg")
+    
+    paket ("pack -v output \"" + outPath) scenario |> ignore
+    ZipFile.ExtractToDirectory(package, outPath)
+
+    Path.Combine(outPath, "lib", "net45", "LocalizedLib.dll") |> checkFileExists
```
smoothdeveloper: Could you use </> operator defined in Paket.IO (it is AutoOpen so you don't even have to open the namespace):

outPath </> "lib" </> "net45" </> "LocalizedLib.dll" |> checkFileExists

JohanLarsson: While it looks sweet I have not seen it in the rest of the code so not sure. Trying to stay consistent with the codebase. Also as a noob `Path.Combine(...)` is nice as I know what it does.

smoothdeveloper: Yes, it's minor thing.

That same operator is defined in FAKE and pretty much any codebase dealing with paths so I thin it is not too far from reach for average F# developer.

You can use go to definition on operators also :smile:

forki:
Why do we show warnings and when? I mean having no satellite assemblies is the 95% case

JohanLarsson:
The warning is shown [here](https://github.com/fsprojects/Paket/pull/1816/files#diff-74c1bcac29cb19f058c1e6032cc1ce19R191), it happens when:
1) A language is found in the project file.
2) No matching `../culture/foo.resources.dll` is found. 
In the previous version I threw here but it was too harsh I think.
I can remove it of course but think it can be a good reminder to build before running `paket pack`.

src/Paket.Core/Cultures.fs
```diff
@@ -0,0 +1,19 @@
+﻿namespace Paket
+open System
+open System.Collections.Generic
+open System.Globalization
+
+[<CompilationRepresentation (CompilationRepresentationFlags.ModuleSuffix)>]
```
smoothdeveloper: Unless you have a type `Cultures` under `Paket` namespace, I don't think we need those attributes (which is used to disambiguate between module and type bearing same name).

JohanLarsson: I copy pasted them from the [Cache module](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Cache.fs#L48), no idea what they mean.

smoothdeveloper: This is needed for Cache module because there is also a Cache type under same namespace.

F# will coallesce both to expose them as Cache but the module need to be given that attribute which under the hood suffixes the static class with Module.

We don't need it for this Cultures since there is no Cultures type that would conflict.

JohanLarsson: Thanks for the explanation, removing it.

forki: there is always room for new PR ;-)

JohanLarsson: I was gonna try to fix the uri issue but was too slow :D

forki: That one is a bit tricky. not sure if my fix was good. have to think about it a bit more.

forki:
there is still one related test failing on mono: https://travis-ci.org/fsprojects/Paket/builds/145555619

JohanLarsson:
Checking it out, didn't see that one before, strange.

forki:
I assume there is something not commited

forki:
`#1816 pack localized happy path` fails on my machine with:


    System.Exception : Paket failed with:
	Could not find 'paket.dependencies'. To use Paket with this solution, please run 'paket init' first.
    If you have already run 'paket.init' then ensure that 'paket.dependencies' is located in the top level directory of your repository.
    Like this:
    MySourceDir
      .paket
      paket.dependencies

JohanLarsson:
Looks like the templates are missing, must have dumbed it during some rebase. Opening a new WIP PR for fixing it.

PR #1821, opened 2016-07-18
Fix localization tests

JohanLarsson:


forki:
still getting "Could not find 'paket.dependencies'. To use Paket with this solution, please run 'paket init' first."

JohanLarsson:
Ok, adding it, not sure it was ever there.

forki:
now it complains about missing lock files. keep pushing things ;-)

JohanLarsson:
No fun for the planet when I waste cycles like this. Adding lock files.

JohanLarsson:
Green now.

forki:
thanks!

Issue #1822, opened 2016-07-18
Environment variable password does not authenticate private feed during convert-from-nuget


john-patterson:
### Description

I am converting a project from NuGet to Paket using the `convert-from-nuget` command. As per [this page](https://fsprojects.github.io/Paket/nuget-dependencies.html) I should be able to authenticate a private feed through the `%PRIVATE_FEED_USER%` and `%PRIVATE_FEED_PASS%` variables, but I cannot. Supplying the credentials as plaintext in the file works and I have verified that the variables do exist in the executing environment as `echo %PRIVATE_FEED_USER%` and `echo %PRIVATE_FEED_PASS%` return the correct values. 

The `--creds-migrations` options are not satisfactory as I would not like to have plaintext passwords in source control, but for a multi-user environment, it does not make sense to encrypt one user's credentials and put it under source control. Users should be able to supply their own credentials for the private (TeamCity) NuGet feed.

### Repro steps

1. Have a non-Paket project that consumes a private NuGet feed (this one is TeamCity) for a dependency.
2. Create a `.paket` directory with `paket.exe`, `paket.bootstrap.exe`, and `paket.targets`
2. Create a paket.dependencies file to specify the source with ` source </path/to/feed/uri> username: "%PRIVATE_FEED_USER%" password: "%PRIVATE_FEED_PASS%"` 
3. Set the PRIVATE_FEED_USER and PRIVATE_FEED_PASS to valid credentials
4. Ensure that these are in effect in the current executing environment with echo
5. Run `.paket/paket.exe convert-from-nuget --force` or `.paket/paket.exe convert-from-nuget` in the main project directory.

### Expected behavior

Environmental variables should authenticate private feeds with `convert-from-nuget`.

### Actual behavior

Paket fails to retrieve package listing when using `convert-from-nuget` with environmental variables for a private NuGet feed, but succeeds when using plaintext credentials.

### Known workarounds

Put plaintext credentials in the `paket.dependencies` file.

### Related information 

* Operating system: Windows 7 Professional SP1
* Branch: [Paket master tagged 3.8.0](https://github.com/fsprojects/Paket/releases/tag/3.8.0)
* .NET Runtime, CoreCLR or Mono Version: .NET Framework 4.5.1
* Performance information, links to performance testing scripts

I can spend some more time on this later, but I want to make sure this is not "expected behavior" and I'm just missing something.

forki:
@theimowski any ideas?

forki:
[...]

PR #1825, opened 2016-07-20
Added convert-from-nuget environment vars note

john-patterson:
When converting a project from NuGet to Paket, it is possible to get into a state that a paket.dependencies file is generated, but the process is not finished. If you put plaintext credentials in this spot, it will continue as suspected with the --hard argument. If you put an environment variable it is not expanded.

As this situation might be confusing to anyone who gets into this situation and sees this document, a note linking to the proper credential migration procedure is a reasonable concession.

This references [issue 1822.](https://github.com/fsprojects/Paket/issues/1822)

### DIFF
--- a/docs/content/nuget-dependencies.md
+++ b/docs/content/nuget-dependencies.md
@@ -35,6 +35,9 @@ If you don't want to check your username and password into source control, you c
 
 The [`paket.lock` file](lock-file.html) will also reflect these settings.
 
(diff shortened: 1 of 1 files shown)


john-patterson:
Looks like the Travis server has a full disk.

forki:
thanks! /cc @theimowski

theimowski:
:+1:

PR #1828, opened 2016-07-20
WIP DNC support

forki:
This is first work for .NETCORE support
- [x] make the custom paket task obsolete - it's not going to work
- [x] generate project.deps.json file from paket.references and csproj and put it in obj folder


forki:
I have the feeling I should cherry pick the deletion of the custom task to master

PR #1831, opened 2016-07-25
Don't allow empty string as description - take II

inosik:
Seems like I forgot this in #1728.

However, I also changed the error message to show the merged metadata, not only the infos from `paket.template`. I think that was the intent anyways.

### DIFF
--- a/src/Paket.Core/PackageMetaData.fs
+++ b/src/Paket.Core/PackageMetaData.fs
@@ -61,7 +61,7 @@ let getTitle attributes =
 
 let getDescription attributes = 
     attributes 
-    |> Seq.tryPick (function Description d -> Some d | _ -> None) 
+    |> Seq.tryPick (function Description d when notNullOrEmpty d -> Some d | _ -> None)
 
 let readAssembly fileName =
     traceVerbose <| sprintf "Loading assembly metadata for %s" fileName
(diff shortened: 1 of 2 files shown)


richard-green:
FYI, this has broken our builds. Probably intentionally, but still!

forki:
I can revert if that makes sense

richard-green:
Hi - for now I've pinned our build to use 3.9.1, was more just an informational / warning note to others :-)

forki:
I reverted it. and releases 3.9.2 is in progress.

@inosik I think we should raise a warning instead.

inosik:
Sorry for the inconvenience, @richard-green, but I think this change is needed. Packages with any of `id`, `version`, `authors` or `description`, missing, are practically unusable outside of Paket. That's every tool which uses NuGet.Core, that being NuGet command line, NuGet server, Klondike, Squirrel, Octopus Deploy (AFAIK) and Visual Studio. Probably some more.

@biehlermi and I had a hard time figuring out why NuGet based test adapters didn't work in a project of ours. The reason was a package created with Paket, which broke VSs discovery process.

See also themotleyfool/Klondike#158.

inosik:
@forki, I would rather let Paket fail, because a warning message is easily overseen in the output of a build script. It also clearly indicates the cause of the failure, instead of an error later in the process, like in themotleyfool/Klondike#158. But I also agree that this is kind of breaking.

forki:
We could still give it a default description generated from the id and
raise a warning. This would be a soft fix and just silently repair stuff

inosik:
#1833

BADF00D:
As @inosik mentioned in https://github.com/themotleyfool/Klondike/issues/158, description is an obligatory field.  I'm not a friend of soft fixes. They are almost always a pitfall for anyone who does not know about this. 
By the way. We use a paket on a build server. As far as there is no error, nobody ever cares about log files from builds.

PR #1833, opened 2016-07-26
Generate default description when packing

inosik:
@forki you didn't revert the actual change yesterday, only the commit related to the error message. I brought the change back again.

I'm not entirely sure about the generated description, though.

/cc @richard-green

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,5 @@
-#### 3.9.3 - 25.07.2016
-* REVERT: Don't allow empty string as description in template file - https://github.com/fsprojects/Paket/pull/1831
+#### 3.9.2 - 25.07.2016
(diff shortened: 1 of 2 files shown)


inosik:
The warning gets printed even if a description exists, Have to change something.

forki:
:kiss: thanks for helping here and staying wit us ;-)

inosik:
I'm still not sure what @richard-green does with those invalid packages :smile:

forki:
probably using them internally with paket as client!?

inosik:
Should I undo the change to `RELEASE_NOTES.md`?

forki:
No all is well

richard-green:
Hi - yes most of our packages are for internal use, we have many of them too, so I'll get round to bulk-updating them with some generic description ;-)

PR #1838, opened 2016-07-28
Version in lock file

creib:
Enabled a way to specify a version in the local source override.

If a version for a dependeny is provided in the local-file, this version is used from the specified source location. If no version is provided, the original version is used. 

UseCase:
We have a buildprocess and releaseprocess which builds different versions as the local build. So for example the releasebuild creates a buildartifact with version 16.1.0. The local build creates an artifact with a version 0.0.0-local. By this extention, it is possible to reference the local-build-folder in the local-file **without changing the version in the lock-file**.


theimowski:
It seems I implemented the same functionality in #1843 - sorry! Wasn't aware of this PR.
Overally it looks good and I like you regex more than mine, however I'm not sure if the version should be specified for "Git overrides" as well.
Do you mind if we take your version of regex and close this PR in favor of #1843?

PR #1859, opened 2016-08-17
Bootstrapper: fixed partial download

Allon-Guralnek:
For some reason, when downloading using the three-parameter `DownloadFile` method (which was only used by the `GitHubDownloadStrategy`), it only downloaded a partial file. The simpler two-parameter overload (used by `NugetDownloadStrategyTests`) always downloaded a full file. It seems the `stream.CopyTo()` used in the three-parameter overload doesn't work well. I removed the problematic overload (which wasn't needed anyway, the simpler one is better). Now Paket always downloads files correctly for me.


forki:
cool thanks

PR #1866, opened 2016-08-21
add NUnit3TestAdapter so the unit tests can be run from visual studio

0x53A:
This enables the integrated Test Explorer:

image


forki:
I don't use that, but look sgood.

forki:
thanks

forki:
after merging this I get Compiler Warning 1685

    The predefined type 'System.Runtime.CompilerServices.ExtensionAttribute' is defined in multiple assemblies in the global alias; 

I had to revert for now. Any ideas?

0x53A:
Hm, I can't reproduce the warning with a2596d04dd8b328d226d8a7740392c4d15f25f27.

Also, I can't find the warning in the CI build of this pr: https://ci.appveyor.com/project/SteffenForkmann/paket/build/0.0.1.6673

Did reverting the PR fix it for you?

forki:
yes it did. It only fails to build in VS. It worked in normal build

0x53A:
I did a git clean and then a rebuild in VS but still can't reproduce it:
image

0x53A:
Ok so I found the cause - nunit defines the attribute ...
image

I'm not sure why it doesn't fail in VS 2015, this looks like a difference between the native compiler and roslyn.

DigitalFlow:
@forki:
Is there somethign special about the NUnit3TestAdapter package?
When adding the package using NuGet Visual Studio detects all tests as it should.
Once I remove it and add it using paket, Visual Studio is not detecting the tests anymore.

If I only exclude the packages.config from the project and change the NUnit3TestAdapter reference that is added as project import to the files added by paket, it is still working.
Once I completely remove the packages.config, everything breaks again. Do you know of any voodoo that is going on in there?

Thanks for your help.

Kind Regards,
Florian

forki:
I assume it's one of the packages that need the following setting: https://fsprojects.github.io/Paket/nuget-dependencies.html#Putting-the-version-number-in-the-path

DigitalFlow:
@forki That seems to be the solution, thanks a lot :)

PR #1868, opened 2016-08-22
Add Unit test adapter try 2

0x53A:
ref https://github.com/fsprojects/Paket/pull/1866#issuecomment-241552025

@forki, this solved the issue for me, can you please try if this also works on your machine?

This pr is not really important, since it changes nothing in paket itself, but it helps me (and maybe a few others) to run the unit tests because i have neither resharper nor ncrunch.

I only excluded it in the one C# project, since there seem to be no warnings in the F# test projects, even though they reference the same set of assemblies...


forki:
seems to work. let's see how it goes. thanks

0x53A:
Thanks

Issue #1871, opened 2016-08-23
paket.exe install fails to install package on OSX/mono.


MorganPersson:
### Description
`mono .paket/paket.exe install` does not install any packages.
(and adding argument -v doesn't add any extra output)

### Repro steps
Files
paket.dependencies:
source http://www.nuget.org/api/v2
framework >= net45

group Main
source http://www.nuget.org/api/v2

nuget Suave >= 1.0.0

paket.references:
Suave

Please provide the steps required to reproduce the problem
`$ mono .paket/paket.exe -v install`
output:
```
Paket version 3.15.2.0
Resolving packages for group Main:
 - Suave 1.1.3
    Package not available.
 - Suave 1.1.2
    Package not available.
 - Suave 1.1.1
    Package not available.
 - Suave 1.1.0
    Package not available.
 - Suave 1.0.0
    Package not available.
Paket failed with:
	There was a version conflict during package resolution.
  Could not resolve package Suave:
   - Dependencies file requested package Suave: >= 1.0
   - Available versions:
     - (2.0.0-alpha4, [http://www.nuget.org/api/v2])
     - (2.0.0-alpha3, [http://www.nuget.org/api/v2])
     - (1.1.3, [http://www.nuget.org/api/v2])
     - (1.1.2, [http://www.nuget.org/api/v2])
     - (1.1.1, [http://www.nuget.org/api/v2])
     - (1.1.0, [http://www.nuget.org/api/v2])
     - (1.0.0, [http://www.nuget.org/api/v2])
     - (1.0.0-beta2, [http://www.nuget.org/api/v2])
.... (more versions listed)
     - (0.0.1, [http://www.nuget.org/api/v2])

  Please try to relax some conditions.
```

### Expected behavior
Suave should be installed.
Or, if I supply '-v' I get details about why the different versions couldn't be installed.

### Related information 
* OS X El Capitan 10.11.6
* Mono 4.2.3.4
* F# 4.0
* Paket 3.15.2.0
* Working internet connection :-)
* NuGet working, I can download Suave 1.1.3 manually

forki:
I don't think this is mono related, 
and it works for me. 
I will add a integration test to test on mono. wait a sec

forki:
test at https://github.com/fsprojects/Paket/pull/1872

forki:
can you try to run "paket update -f" ?

MorganPersson:
No difference at all.
[...]

PR #1872, opened 2016-08-23
Try to reproduce #1871

forki:


forki:
ok breaks on mono, but for different reason

PR #1874, opened 2016-08-24
[utils] trying to fix #1873

### LINKED ISSUE #1873: Hitting github limits
### Description

With a few file dependencies to github and a morning of updating of dependencies, I'm hitting the limits with

```
Paket failed with:
       	Could not find hash for https://api.github.com/repos/haf/YoLo/commits/master
```

or

```
Paket version 3.15.4.0
Paket failed with:
       	Could not find hash for https://api.github.com/repos/xyncro/chiron/commits/e0f4c122b21cb02d47ae3f4f83924b2dcb8dea9e
```

```
$ curl https://api.github.com/rate_limits
{
  "message": "API rate limit exceeded for 89.255.229.33. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```

### Repro steps

1. Step A – spend a morning upgrading our solution

2. Step B – reach error message

### Expected behavior

Working calls to github.

haf:
implements #1873

### DIFF
--- a/src/Paket.Core/Utils.fs
+++ b/src/Paket.Core/Utils.fs
@@ -332,9 +332,14 @@ let getDefaultProxyFor =
 
 let inline createWebClient (url,auth:Auth option) =
     let client = new WebClient()
+    client.Headers.Add("user-agent", "Paket")
+    client.Proxy <- getDefaultProxyFor url
+
+    let githubToken =
+        Environment.GetEnvironmentVariable "PAKET_GITHUB_API_TOKEN"
+
     match auth with
-    | None -> client.UseDefaultCredentials <- true
-    | Some(Credentials(username, password)) -> 
+    | Some (Credentials(username, password)) ->
         // htttp://stackoverflow.com/questions/16044313/webclient-httpwebrequest-with-basic-authentication-returns-404-not-found-for-v/26016919#26016919
         //this works ONLY if the server returns 401 first
         //client DOES NOT send credentials on first request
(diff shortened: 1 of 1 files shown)


haf:
It doesn't cover the Token in config, because that's how I interpreted the `token` DU despite not finding any docs for it

forki:
Mhm. maybe we don't need the token in the config.

But we definetly need some docs for this.

forki:
Ok I will write something...

forki:
thanks for adding this!

Thorium:
I always have problems with this one. And the reason is that I have multiple files from same repository like this:
```
  github borisyankov/DefinitelyTyped lodash/lodash.d.ts
  github borisyankov/DefinitelyTyped jquery/jquery.d.ts 
  github borisyankov/DefinitelyTyped jqueryui/jqueryui.d.ts
  github borisyankov/DefinitelyTyped signalr/signalr.d.ts
  github borisyankov/DefinitelyTyped google.analytics/ga.d.ts 
  github borisyankov/DefinitelyTyped mustache/mustache.d.ts
  github borisyankov/DefinitelyTyped rickshaw/rickshaw.d.ts 
  github borisyankov/DefinitelyTyped d3/d3.d.ts
  github borisyankov/DefinitelyTyped yamljs/yamljs.d.ts
  github borisyankov/DefinitelyTyped rx/rx.lite.d.ts
  github borisyankov/DefinitelyTyped rx/rx.async-lite.d.ts
//...and 40 more...
```
Now what happens it that when I call paket update, it asks for the latest comment id for all in parallel multiple requests. (How could it know that it is the same for all!) And when the first response comes, it caches it, but before the response it has already made the other requests. GitHub thinks that this is flood (like it is) and blocks me. So there is no any kind of singleton in this: it makes the requests in parallel and no locks. Responses are saved to a ConcurrentDictionary which is not used as locking (remember maybe the concurrency-memoize-getOrAdd what I have been talking about). And the reason that it can't lock is that it's async operation to fetch.

I don't want to reference the whole huge repository of DefinitelyTyped.

forki:
@Thorium does the new feature solve it for you?

Thorium:
I think the API keys are not the actual problem.
It's like "does ibuprofen help"?

forki:
ok that it could be smatter when asking for the same repo is another issue. But API key still helps right? 
Care to send a PR for that?

Thorium:
#1880

Issue #1875, opened 2016-08-24
Missing argument `--version`


haf:
Support showing the version from Paket.

```
$ mono bin/paket.exe --version
ERROR: unrecognized argument: '--version'.
// snip
```

forki:
done with #1878

PR #1878, opened 2016-08-24
Allow printing of version number through command-line option

ErikSchierboom:
This PR implements the feature requested in #1875, where you can specify the `--version` argument to print the version to the console. However, while implementing this,  I found I had to remove the `RequireSubcommand` attribute, as `--version` is (rightfully) not considered a sub-command and having the `RequireSubcommand` attribute would cause it to fail. The result of this change is of course that sub-commands are now not necessary, which I think is a bit odd. What do you think about this? 

The main problem is that #1875 suggests that the `--version` flags is used for displaying the version, but perhaps it should just be a regular subcommand?

### DIFF
--- a/src/Paket/Commands.fs
+++ b/src/Paket/Commands.fs
@@ -311,12 +311,12 @@ with
         | Framework _ -> "Framework identifier to generate scripts for, such as net4 or netcore."
         | ScriptType _ -> "Language to generate scripts for, must be one of 'fsx' or 'csx'."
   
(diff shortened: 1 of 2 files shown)


eiriktsarpalis:
@forki LGTM

ErikSchierboom:
I'll remove the [WIP] part from the title then.

forki:
thanks!

ErikSchierboom:
:tada:

PR #1881, opened 2016-08-24
Use DU-based parameters for auto restore flags

eiriktsarpalis:
I incidentally noticed that the auto-restore command uses an odd way of specifying on/off params. This PR showcases Argu's ability to specify flags using DUs/enums. Here's what the syntax looks like now:
```
$ ./bin/paket.exe auto-restore --help
Paket version 3.16.0.0
Help was requested:
USAGE: paket auto-restore [--help] <on|off>

FLAGS:

    <on|off>              enables or disables auto restore for the repo.

OPTIONS:

    --verbose, -v         Enable verbose console output for the paket process.
    --log-file <path>     Specify a log file for the paket process.
    --silent, -s          Suppress console output for the paket process.
    --help                display this list of options.
```

### DIFF
--- a/src/Paket/Commands.fs
+++ b/src/Paket/Commands.fs
@@ -82,15 +82,15 @@ with
     interface IArgParserTemplate with
         member __.Usage = ""
 
+type AutoRestoreFlags = On | Off
+
 type AutoRestoreArgs =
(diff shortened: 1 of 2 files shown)


forki:
cool. thanks

PR #1886, opened 2016-08-25
Verbose should be set when --verbose is passed in

MorganPersson:
Seems weird to set verbose only when silent is set.


forki:
thanks. is your other issue fixed as well?

MorganPersson:
Well, with verbose fixed you get the exception when you run with --verbose, but I will see if I can get a good error without verbose. It's just that the function getPackageDetails in NuGetV2 wasn't so easy to grasp.

PR #1892, opened 2016-08-26
Better Nuget V3 API and async caching

Thorium:
Better Nuget V3 API and async caching.
Public API and Nuget V2 API calls to `Async.RunSynchronously` should be still refactored away to be real async, but those would cause function signature changes and for that reason bigger refactoring.


forki:
so what is doing? or what was the problem?

Thorium:
getNuGetV3Resource function was doing network fetching but:
- getNuGetV3Resource errors were swallowed also in verbose mode. Not after this commit.
- cache was not thread-safe. After this commit it is.
- network operation was sync, not async. Because "Async.RunSynchronously" is actually sync. It's still sync.

Thorium:
The build is already failing before this...

forki:
:kiss:

Issue #1905, opened 2016-09-01
bug: odata URL wrong capitalization


llehn:
### Description

This commit added toLower() to the URL: 
https://github.com/fsprojects/Paket/commit/fd303582827f069cbdb9818701db7a41ee735145

OData spec says "toLower" should be "tolower" (in fact all filter function are lower-case only)

### Repro steps

go to

https://www.nuget.org/api/v2/Packages?$filter=toLower(Id)%20eq%20%27system.linq%27

it does not work

replace toLower with tolower

https://www.nuget.org/api/v2/Packages?$filter=tolower(Id)%20eq%20%27system.linq%27

it does work.

I use ProGet and it also doesn't understand toLower but tolower

llehn:
merged and fixed

PR #1906, opened 2016-09-01
Fixed toLower > tolower for odata url parameter

llehn:
Fix for https://github.com/fsprojects/Paket/issues/1905


forki:
cool thanks.

PR #1923, opened 2016-09-18
Remmi Demmi! - Framework restriction could not be determined during restore

dnauck:
.. so just ignore those error messages


forki:
thx

PR #1932, opened 2016-09-30
Generate binding redirect that covers entire range of possible assembly versions

### LINKED ISSUE #1930: In 'bindingRedirect' element, 'oldVersion' range of '0.0.0.0-999.999.999.999' does not cover all possible versions
### Description

There are some libraries that have major version greater than 999 (e.g., https://www.nuget.org/packages/Stimulsoft.Reports.Web.Trial/ has version 2016.3.14). Because of this, the binding redirect generated by Paket does not apply to assemblies from those packages.

I propose to use range like '0.0.0.0-65535.65535.65535.65535' (which is a maximum possible version for .net assembly) to ensure that all possible versions are covered.

### Repro steps

1. Unpack the attachment
[paket-redirect-999-repro.zip](https://github.com/fsprojects/Paket/files/497859/paket-redirect-999-repro.zip)

2. Run `paket install`

### Actual behavior

App.config file contains this fragment:

```
  <dependentAssembly>
    <Paket>True</Paket>
    <assemblyIdentity name="Stimulsoft.Base" publicKeyToken="ebe6666cba19647a" culture="neutral" />
    <bindingRedirect oldVersion="0.0.0.0-999.999.999.999" newVersion="2016.1.3.0" />
  </dependentAssembly>
```

(note that 2016.1.3.0 is outside of range 0.0.0.0-999.999.999.999; so this redirect will not apply to earlier versions of this assembly)

### Expected behavior

`oldVersion` attribute should have range that at least covers the version 2016.1.3.0.

dmitryvk:
Fixes #1930

### DIFF
--- a/integrationtests/scenarios/i001248-redirects-off/before/MyClassLibrary/MyClassLibrary/app.config
+++ b/integrationtests/scenarios/i001248-redirects-off/before/MyClassLibrary/MyClassLibrary/app.config
@@ -6,7 +6,7 @@
   <dependentAssembly>
     <Paket>True</Paket>
     <assemblyIdentity name="FSharp.Core" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
-    <bindingRedirect oldVersion="0.0.0.0-999.999.999.999" newVersion="4.4.0.0" />
+    <bindingRedirect oldVersion="0.0.0.0-65535.65535.65535.65535" newVersion="4.4.0.0" />
   </dependentAssembly>
 </assemblyBinding></runtime>
 </configuration>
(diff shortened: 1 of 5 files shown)


forki:
there are couple of broken integration tests. Could you please try to fix these as well?

dmitryvk:
Yes, I will (actually, I just tried to fix them once and a little surprised that they failed).

dmitryvk:
There is now one failing test remains (and an timeout-caused error) but I don't think it's related:

```
Paket.IntegrationTests.InstallSpecs.#1270 install net461
  Expected string length 4623 but was 5057. Strings differ at index 3952.
  Expected: "...ifier) == '.NETFramework' And ($(TargetFrameworkVersion) =..."
  But was:  "...ifier) == '.NETFramework' And $(TargetFrameworkVersion) ==..."
  --------------------------------------------^
```

PR #1945, opened 2016-10-06
generate-include-scripts: don't check dll order when it can be skipped 

smoothdeveloper:
reduce time to generate loading script by not checking dll metadata when it is useless (took 26s before and now 13s on a big paket.dependencies file)


forki:
thanks!

PR #1946, opened 2016-10-07
generate-include-script should not #r FSharp.Core.dll

### LINKED ISSUE #1942: generate-include-script should not #r FSharp.Core.dll
### Description

If you depend on a package that depends on `FSharp.Core` package, the `generate-include-script` command should not generate `#r "...FSharp.Core.dll"

### Repro steps

Run `paket install` and `paket generate-include-scripts framework net45` for a `paket.depenencies` file containing:

```
nuget Suave
```

Then you should be able to run:

```
#load @"paket-files\include-scripts\net45\include.main.group.fsx"
```

### Expected behavior

Evaluating the `#load` command in FSI works.

### Actual behavior

You get an error:

> error FS1222: When mscorlib.dll or FSharp.Core.dll is explicitly referenced the --noframework option must also be passed

### Known workarounds

You can start your `fsi.exe with `--noframework` option, but I think we don't want to make people do this.

smoothdeveloper:
Add test case and fix implementation

Fixes #1942

### DIFF
--- a/integrationtests/Paket.IntegrationTests/LoadingScriptGenerationTests.fs
+++ b/integrationtests/Paket.IntegrationTests/LoadingScriptGenerationTests.fs
@@ -171,3 +171,23 @@ let ``mscorlib excluded from f# script`` () =
         )
 
     Assert.False hasFilesWithMsCorlib
+
+
+[<Test; Category("scriptgen")>]
+let ``fsharp.core excluded from f# script`` () =
+    let scenario = "fsharpcore"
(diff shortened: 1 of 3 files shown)


smoothdeveloper:
@forki this PR failed to pass integration due to timeout on appveyor, do you have a way to run it again?

forki:
I think easiest way is to push a new commit (which clean up or zero diff)

forki:
thanks!

Issue #1959, opened 2016-10-12
Create "why" command


forki:
https://twitter.com/codepo8/status/785968938307903488

this command should analyze the paket.lock and show the reason in nice syntax.

theimowski:
In addition to the resolution from paket.lock I'd also love to see whether any project actually references the package (can take that from paket.references).
This could be further extended to show whether the reference is reasonable from code POV (whether it's used at all)

Those however would imply that "why" command could end with "there's no reason for depending on X package" reason

forki:
@theimowski did you just volunteered?

theimowski:
I think so! Hope I can find some spare time to hack on this

Stift:
Does not "find-refs" displaying what project references a dependency?

PR #1960, opened 2016-10-12
[WIP] Why oh why?

theimowski:
#1959

TODO:
- [ ] recognize github dependencies
- [X] show dependencies up to the top level
- [ ] option for displaying version and depending packages version constraints
- [x] show the shortest path by default, give option to show all paths
- [x] mark all direct dependencies in path
- [ ] reword "dependency path" -> "dependency chain", make the output more user-friendly
- [ ] what should be displayed when asking for a direct dependency, and what if it is not a top-level dependency?
- [ ] handle circular dependencies ?

Feedback welcome!

image

### DIFF
--- a/src/Paket.Core/Paket.Core.fsproj
+++ b/src/Paket.Core/Paket.Core.fsproj
@@ -141,6 +141,7 @@
     <Compile Include="NugetConvert.fs" />
     <Compile Include="FindOutdated.fs" />
     <Compile Include="FindReferences.fs" />
+    <Compile Include="Why.fs" />
     <Compile Include="PublicAPI.fs" />
     <Compile Include="ScriptGeneration.fs" />
     <None Include="paket.template" />
--- a/src/Paket.Core/Why.fs
+++ b/src/Paket.Core/Why.fs
@@ -29,24 +29,31 @@ let depGraph (res : PackageResolver.PackageResolution) : AdjGraph<Domain.Package
 type WhyOptions = 
     { AllPaths : bool }
 
-let prettyFormatPath path = 
+type WhyPath = list<PackageName * bool>
+
+let prettyFormatPath (path: WhyPath) = 
     path 
-    |> List.mapi 
-        (fun i v -> 
-            sprintf "%s-> %s%s"
+    |> List.mapi
+        (fun i (name, isDirect) -> 
+            sprintf "%s-> %O%s"
                     (String.replicate i "  ")
-                    v
-                    (if i = 0 then sprintf " (%s)" Constants.DependenciesFileName else ""))
+                    name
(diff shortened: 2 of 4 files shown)


cdrnet:
This will become very useful, thanks! Could it also provide some info on why a particular version is used, if applicable?

Stift:
When a package dependency is shown, do we show all dependencies up the dependency tree up to the top-level, only top level or only to the next level?
Could also be a command switch like `--full`

forki:
@theimowski please keep posting screenshots if you change something. This is awesome

theimowski:
will do, I'll also let know whenever I get to the point where it could be merged - possibly with not yet all features implemented

isaacabraham:
Would it be worth making package id optional - so that it shows why all packages are in the lock file for a given group (although this might be a bit verbose)?

theimowski:
@Stift I think it makes sense to show all deps up to the top level by default.
What I'm wondering is whether it makes sense to print all paths from the top level package, or just choose a shortest one? (see below e.g. Microsoft.Bcl.Build) 

image

forki:
Maybe show the shortest and say "and x more paths"?

tommes:
Maybe (as a 2nd step) we couln't just ask for package id's, but also for artefacts - i.e. why is the dll/js/whatever in my output. In case I've overlooked that just ignore the comment.

forki:
That would be brilliant

Am 14.10.2016 09:27 schrieb "Thomas Koch" <notifications@github.com>:

forki:
@tommes but it's probably hard to decide without reimplementing MSBuild logic

Stift:
I think by package Id would be sufficient for the most use case. 
If there are more than one path then, I'd like to have an option like -full or -all, if possible.
If a tree notation would be too complex I would recommend sth like ´Package1 >> Package2 >> Package3´
@theimowski Would it make sense to merge the trees, so that the output is more compact?

forki:
If we merge the trees then we will get the lock file back

Am 14.10.2016 17:05 schrieb "Christian Fiebrig" <notifications@github.com>:

theimowski:
Can NuGets have cyclic dependencies? And if so does it resolve to paket.lock?

theimowski:
How about this? I've grouped the paths by the top level dependency and if from a specific top level dep there are >1 paths, we show shortest by default.
This way we won't ignore multiple top-level deps

image

forki:
I released it in alpha. please create new PRs to keep it going ;-)

theimowski:
Alright! Can one continue work on  the same branch and reopen PR in github or do i have to create new PRs?

forki:
You can try to reopen. Not sure if this works.

Am 22.10.2016 17:51 schrieb "Tomasz Heimowski" <notifications@github.com>:

PR #1963, opened 2016-10-13
Added option to have paket restore fail on check failure

sarudak:
I removed the exception swallowing. If there's a specific exception it was meant to suppress that should be added back in. Also are there tests around this functionality? I couldn't find any related ones.

PS: Total F# noob here


forki:
cool.

Do we need the same param for install and update?

sarudak:
I don't think it's necessary. The particular check that I care about doesn't happen on install or update because they always update the lock file. If there are other warning but not failing events that people would prefer to fail in certain situations then maybe but I don't know about them.

Any idea why I've got integration tests failing in travis-ci but not locally?

kostrse:
This is a great flag.

Maybe give it a shorter name?
`--fail-on-checks` for example

forki:
Yes that name is better. Please change it.

Am 19.10.2016 02:09 schrieb "Sergey Kostrukov" <notifications@github.com>:

forki:
thx. this is now in latest alpha

PR #1969, opened 2016-10-18
Add Install-Paket.ps1 to prevent embedding .exe in git.

jschneidereit:
This checks the github api for the latest release and appropriately
downloads the paket bootstrapper. An optional parameter `-AddToPath` can
be used to add the paket directory to `$env:Path` for CI boxes.

The intent is to allow those who have *.exe in their .gitignore to more
easily use paket on Windows.

### DIFF
--- a/Install-Paket.ps1
+++ b/Install-Paket.ps1
@@ -1,0 +1,36 @@
+[CmdletBinding()]
+Param(
+    [Parameter(Mandatory=$true)][System.IO.FileInfo]$Path,
+    [Switch]$AddToPath
+)
+
+$old_location = Get-Location
+$paket_directory = Join-Path $Path ".paket"
+If (-not (Test-Path $paket_directory)) {
+    New-Item -Path $paket_directory -Type Directory
+}
+
+$bootstrapper_name = 'paket.bootstrapper.exe'
+$latest_request = "https://api.github.com/repos/fsprojects/Paket/releases/latest"
+$executable =  Join-Path $paket_directory $bootstrapper_name
+
+$latest = (Invoke-WebRequest -Uri $latest_request).Content | ConvertFrom-Json
+$asset_request = "https://api.github.com/repos/fsprojects/Paket/releases/$($latest.id)/assets"
+
+$bootstrapper = (Invoke-WebRequest -Uri $asset_request).Content | ConvertFrom-Json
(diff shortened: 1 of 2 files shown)


forki:
I recommend to use the bootstrapper. 

*.exe in git can easily worked around for that file. The bottstrapper as different fall back mechanisms like downloading from nuget when github is down. This will keep your builds much more green,

matthid:
Tbh nobody likes PowerShell because it is not really xplat (no matter what they say about it being open source now, because it is just not installed on unix boxes).
Bash is more xplat because it is part of git which itself is part of visual studio (which most have installed)

jschneidereit:
@forki The intention was not to replace bootstrapper, just to provide an additional option for those who don't wish to embed binaries in their source control.

forki:
Still can't understand that strange requirement. We created the small
bootstrapper in order to keep size small and it doesn't change that often.
It's just not a real issue.

BTW did you see https://github.com/fsprojects/Paket/pull/1961 -  it's
taking the whole thing to the next level.

Am 19.10.2016 20:03 schrieb "Jim Schneidereit" <notifications@github.com>:

jschneidereit:
I think it's common enough of a scenario not to include binaries in source that I felt it was useful to give this option on Windows via PowerShell.  I'm happy to close the PR if you think the bootstrapping model is the way to go.

forki:
Yes I think for a general recommendation the bootstrapper is the way to go. It's a much safer choice. 

While I admire your effort to create a powershell bootstrapper and I am happy that you are willing to share it, I'm thinking this is not the way to go. Tbh that don't commit exe rule should not be considered as a 100% dogma. That just makes no sense and your life will be harder than it should be. Try to understand where the rule is coming from and why it is there. Try to understand what exceptions to that rule are useful. 
Hint: often it's just that someone heard rumors that git is bad with binary files. Which is basically wrong.

matthid:
I agree with @forki. On the other side if there would be a `good` way to do the same without a binary and X-Plat I would prefer it... Suggestions are welcome. So thanks for sharing your work. In fact there might even come the time when Powershell is installed by default on the large distributions... (I still hope god saves us from this ;) )

jschneidereit:
Sounds good, thanks for taking a look! (I hope god brings us PowerShell on all large distros :laughing:)

PR #1977, opened 2016-10-24
WIP Make Paket work with dotnet cli preview3

forki:
- [x] - Make "paket install" hook into MSBuild 15 / dotnet cli preview3 restore
- [x] - Make Paket restore it's own PaketRestoreTask.dll
- [x] - Do not emit dll references for MSBuild 15 projects
- [x] - Use magic mode
- [x] - Make everything work in ionide
- [ ] - Make Paket create it's own Nuget.Config or hook into that in MSBuild
- [ ] - Convert from Nuget should read / remvoe PackageReferences from *proj - (@theimowski ?)


src/PaketRestoreTask/Program.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.IO;
+using System.Text;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Paket.Build.Tasks
+{
+    public class PaketRestoreTask : Task
```
matthid: Just some crasy thinking: Would it make sense to think about extension points in paket?

Considering all those different use cases we might have soon (dotnetcore, toml, legacy, fable?) shouldn't we try to get things like these out of paket core into modules or something like that?

forki: sure. bringe them on

forki:
paketcore

theimowski:
Re convert - there's no packages.config anymore,  but rather every dep is moved to *proj? Is it a 1-1 mapping?

forki:
It's looking like this now:

    <Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
      <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" />
      <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFramework>netcoreapp1.0</TargetFramework>
      </PropertyGroup>
      <ItemGroup>
        <Compile Include="**\*.cs" />
        <EmbeddedResource Include="**\*.resx" />
      </ItemGroup>
      <ItemGroup>
        <PackageReference Include="Microsoft.NETCore.App">
          <Version>1.0.1</Version>
        </PackageReference>
        <PackageReference Include="Microsoft.NET.Sdk">
          <Version>1.0.0-alpha-20161019-1</Version>
          <PrivateAssets>All</PrivateAssets>
        </PackageReference>
      </ItemGroup>
      <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
    </Project>

theimowski:
I can give it a try, but can take a week or two  - little spare time right now

forki:
yeah don't worry. no need to hurry

theimowski:
which dotnet core build you're using? Is it 1.0.0-preview3-003884?
How to download that build @enricosada ?
I tried https://dotnetcli.blob.core.windows.net/dotnet/Sdk/rel-1.0.0/dotnet-dev-win-x64.1.0.0-preview3-003884.zip but that was a miss

With the latest dotnet (https://dotnetcli.blob.core.windows.net/dotnet/Sdk/rel-1.0.0/dotnet-dev-win-x64.latest.zip), when I try to repro your showcase, I get 

image

forki:
@theimowski you must run paket install once  before you can restore.

theimowski:
Thanks, that was the case. Also turns out, in latest dotnet core there's no more "3" suffix for commands https://github.com/dotnet/cli/issues/4205

theimowski:
I'm hacking around convert - pushing changes directly to this branch - is it fine with you?
What I'm struggling with is:
```Paket failed with:
        Could not find versions for package Microsoft.NET.Sdk on any of ["https://www.nuget.org/api/v2"; "http://api.nuget.org/v3/index.json"].```
which should probably work, as `dotnet restore` outputs: ``` Feeds used:
      https://api.nuget.org/v3/index.json ```

**Comment by forki** [comment]
Yes that's fine. Maybe we need another feed. @enricosada?

**Comment by forki** [comment]
maybe https://github.com/dotnet/netcorecli-fsc/blob/master/examples/preview3/console/NuGet.Config#L4 ?

**Comment by enricosada** [comment]
https://github.com/dotnet/netcorecli-fsc/blob/master/examples/preview3/lib/NuGet.Config


Sent from mobile

**Comment by enricosada** [comment]
Also the dotnet sdk feeds

https://github.com/dotnet/cli/blob/rel/1.0.0/NuGet.Config

Sent from mobile

**Comment by theimowski** [comment]
Nope, I tried that as well - maybe they changed the API for nuget v3 and Paket can't consume it yet?

**Comment by forki** [comment]
@theimowski we can consume v3 API. just make sure all the nuget feeds are added to paket.deps and I think it's without the last part of the url

**Comment by theimowski** [comment]
yeah sorry my fault - didn't spot last @enricosada comment. It seems it's resolving deps now with those 4 additional feeds, but taking ages...

**Comment by forki** [comment]
@theimowski welcome to the brave new world of dotnet shipping via nuget

**Comment by theimowski** [comment]
Huh...

```
c:\sandbox\core3paket (master)
λ paket convert-from-nuget --force
Paket version 4.0.0-alpha021
Adding Microsoft.NET.Sdk 1.0.0-alpha-20161026-2 to c:\sandbox\core3paket\paket.dependencies into group Main
Adding Microsoft.NETCore.App 1.0.1 to c:\sandbox\core3paket\paket.dependencies into group Main
Dependencies files saved to c:\sandbox\core3paket\paket.dependencies
 - Microsoft.NETCore.App is pinned to 1.0.1
 - Microsoft.NET.Sdk is pinned to 1.0.0-alpha-20161026-2
 - newtonsoft.json 9.0.1
 - ...
Locked version resolution written to c:\sandbox\core3paket\paket.lock
Downloading ...
Garbage collecting ...
34 minutes, 5 seconds - ready.
```

and that's for just 3 deps:

```
source https://www.myget.org/F/xunit/api/v3/index.json
source https://www.myget.org/F/netcorecli-fsc-preview3/api/v3/index.json
source https://dotnet.myget.org/F/dotnet-core/api/v3/index.json
source https://dotnet.myget.org/F/cli-deps/api/v3/index.json
nuget Microsoft.NET.Sdk 1.0.0-alpha-20161026-2 alpha
nuget Microsoft.NETCore.App 1.0.1
nuget newtonsoft.json
```

**Comment by forki** [comment]
how fast is second update?

**Comment by theimowski** [comment]
you mean running now `paket install --force` ?

**Comment by enricosada** [comment]
Cold cache.
The sdk and packages ship with .net core sdk package so usually is not a problem.

And other packages are from nuget.org<http://nuget.org> so faster.
It's possible to tweak the feeds, and use less big feeds.
Or create a new feed for paket-dev with   mirror of useful packages.
 Pretty much all are in nuget.org<http://nuget.org>v (minus msbuild/xunit dev/fsharp.sdk)

Sent from mobile

**Comment by forki** [comment]
@theimowski no --force would take same time (since it skips cache). just run paket update

**Comment by theimowski** [comment]
``` 56 seconds - ready. ```
much better ;)

How does it work now? Paket downloads the nupkgs to ./packages dir and then copies them to %HOME%\.nuget\.packages ?
Or do we somehow tell dotnet to take deps from "./packages"?

**Comment by forki** [comment]
Currently we download to both locations.

Am 03.11.2016 14:27 schrieb "Tomasz Heimowski" <notifications@github.com>:

**Comment by enricosada** [comment]
ok i started the work in https://github.com/fsprojects/Paket/pull/2004
blocked because the .net core version (`src/Paket.Core/Paket.Core/project.json`) doesnt compile anymore.
ideas?

**Comment by enricosada** [comment]
Ok also `Paket` is a `netcoreapp1.0` console app. 
using an msbuild fsproj for .net core sdk `preview3`

**Comment by theimowski** [comment]
Can you have side-by-side `packages.config` and `PackageReference` entries in *proj in the new world?
This matters in context of `convert-from-nuget` - with regards to the multitargeting I suppose one can have both ways of referencing nugets?

**Comment by forki** [comment]
I don't know

**Comment by theimowski** [comment]
Alright, I've come to the point where you can invoke `convert-from-nuget` just after `dotnet new`, but I hardcoded the dotnet-core and cli-deps [feeds](https://github.com/dotnet/cli/blob/rel/1.0.0/NuGet.Config) - @enricosada any ideas where I could read those from?

**Comment by enricosada** [comment]
@theimowski there is not place you can read these, are hardcoded. 
but these should not needed **if* the package version inside dotnet-new template is the same as the one in the downloaded sdk (there is a cache so unitial big download on first restore is not needed). 
Usually is like that, but sometimes version Numbers are not updated/aligned. 

At RTM these are going to be on nuget.org, so are justa temporary thing. 

I think i can just put fsharp.net.sdk package in nuget.org so one feed less. 

@theimowski I think you miss the dev feed of fsharp.net.sdk package

**Comment by theimowski** [comment]
```
λ dotnet restore
  Paket version 4.0.0-alpha023
   - c:\sandbox\core3paket\obj\core3paket.csproj.references already up-to-date
  0 seconds - ready.
  Nothing to do. None of the projects specified contain packages to restore.
```

I might be wrong, but I think that if there's no more PackageReference in csproj (after running paket conver), then the restore step has `Nothing to do`, which makes it impossible for Paket to hook to the restore process?

**Comment by forki** [comment]
Urgs. That would be bad. @davkean

Am 11.11.2016 12:38 nachm. schrieb "Tomasz Heimowski" <
notifications@github.com>:

**Comment by enricosada** [comment]
@theimowski @forki can you just whitelist the `Microsoft.NET.Sdk` and `FSharp.NET.Sdk` PackageReference for now and dont remove these (maybe just microsoft.net.sdk is enough).
A more general solution is to not remote PackageReferences with `<PrivateAssets>All` because mean these may contains msbuild targets.
These contains all the msbuild targets (nuget too). the csproj/fsproj is not special and doesnt contains anything. So maybe just work because of first restore (and autoimport inside bin) before convert. after a clean `dotnet restore` should not work anymore (no target, no `/t:Restore`) if there are zero packages.

**Comment by enricosada** [comment]
@forki @theimowski remove special feed, i am going to publish on nuget.org
ref https://github.com/dotnet/netcorecli-fsc/issues/19#issuecomment-261218159 for sdk version

**Comment by theimowski** [comment]
@enricosada at the moment I'm trying to make the convert thing work for vanilla C# project - haven't tried yet with F# project, so didn't need the FSharp.Net.SDK package.

**Comment by enricosada** [comment]
As a note @theimowski @forki , the next `preview4` will support the `Version` as attribute too (instead of Element). maybe it's better not be too picky with xml selector 😄  

ref https://github.com/dotnet/cli/pull/4821

so like 

```xml
<PackageReference Include="Microsoft.NETCore.App" Version="1.0.1" />
```
in addition of

```xml
<PackageReference Include="Microsoft.NETCore.App">
    <Version>1.0.1</Version>
</PackageReference>
```

enricosada:
Q: what is missing?

theimowski:
I have tested convert stuff against C# console app - but according to @enricosada suggestion and ignoring refs with PrivateAssets, it just converts the SDK package. Havent tested for F# proj though yet

enricosada:
@theimowski f# is the same as c#, just reference an additional sdk package `FSharp.NET.Sdk` with `PrivateAssets` (so should not be removed). It's a normal .net proj, should work ok.
Other blocking stuff?

forki:
@enricosada could you please take a look at dotnet pack? Currently that's disabled

enricosada:
Sure @forki didn't see was not enabled   
I fear `dotnet pack` becuase there is https://github.com/NuGet/Home/issues/1365 
So cannot pack a non prerelease version but let's see

forki:
@enricosada in 91c7485 and e8ba92b I tried to use the latest dotnet core, but it failed. ideas?

PR #1979, opened 2016-10-24
More boostrapper config

vbfox:
Add a "Prerelease" setting in `.config` file so that even magic mode bootstrapper can use pre-releases as discussed in #1961.

To use it :

```xml
<configuration>
  <appSettings>
    <add key="Prerelease" value="True"/>
  </appSettings>
</configuration>
```

Also do a small cleanup in ArgumentParser of the bootstrapper.

### DIFF
--- a/docs/content/bootstrapper.md
+++ b/docs/content/bootstrapper.md
@@ -79,7 +79,7 @@ Would do the same thing as :
 paket.bootstrapper.exe -s --max-file-age=720 --run add nuget FAKE
 ```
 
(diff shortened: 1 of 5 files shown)


forki:
cool. magic mode is released in 3.24

vbfox:
@forki nice!

Btw isn't the master branch in a strange state with the bootstrapper renamed in .paket but the build.cmd and build.sh scripts still calling paket.bootstrapper.exe ?

forki:
yeah. fixing this...

PR #1983, opened 2016-10-26
Bootstrapper magic mode enhancements

vbfox:
Quite a few changes to the boostrapper, mainly related to magic mode :

* Add tests around the fact that `MaxFileAgeInMinutes` shouldn't be auto set in magic mode when a version or prerelease is specified.
* Clean the argument parser code and clearly show in what order we apply our different sources of options
* Add a new source of bootstrapper options (Magic mode or not, but especially useful in magic mode) : The `paket.dependencies` file. The first comment starting with regex `(#|//)(\s*)bootstrapper(\s*):` ( is taken into account for that :
    ```
# bootstrapper: 3.24.1 --prefer-nuget

nuget FAKE
nuget FSharp.Core ~> 4
nuget FSharpLint.MSBuild
    ```
* The new order of option sources is : AppSettings, paket.dependencies, Environment variables, Command line (Processing order in code so, from less to more important. If something is on the command line it'll override any other source for example)
* There is a new variant of "silent" mode that displays only errors. It can't be triggered from command line but the magic bootstrapper uses it (Otherwise it display nothing in case of error and that's confusing)
* Known nuget download errors are now displayed with a better explanation
* MSBuild targets file can now work with a `paket.exe` (Most often the bootstrapper in magic mode) at the root instead of in .paket
* The directory / file proxy have been merged into a filesystem proxy and this proxy gained a few methods for better magic mode tests.


forki:
I really like the idea of specifying a needed paket version in the deps file. That is a very often requested feature (IIRC we already have an issue for that but I can't find it). But I don't think we should make that a "comment". It's too important. How about this:

    version 3.24.1 --prefer-nuget
    
    nuget FAKE
    nuget FSharp.Core ~> 4
    nuget FSharpLint.MSBuild

or something like that. It would also be interesting to specify version ranges for paket. Someting like `version >= 3.23`

agross:
It'd be great if the version statement would be optional to always get latest (within `MaxFileAgeInMinutes `).

forki:
yes it needs to be optional - also for backwards compat

vbfox:
Using a normal line instead of a comment is conceptually easy and will look cleaner but I need to test compatibility of paket parser itself. I'll add that if it works well.

Ranges are possible but we'll start to duplicate some of the version check logic of paket itself in the bootstrapper. Maybe for a later PR ? 🦊

forki:
Absolutely. We don't need that now. Pinning a paket version is already awesome

vbfox:
New version with the syntax suggested by @forki :

```
version 3.24.1 --prefer-nuget

nuget FAKE
nuget FSharp.Core ~> 4
nuget FSharpLint.MSBuild
```

vbfox:
@agross The feature is actually completely independent of the "magic" mode where paket.exe is actually the bootstrapper. It's very useful in this mode but could have existed independently of it and would work without it.

What happens is :

* `paket.dependencies` contains `version 1.2.3.4 --prefer-nuget`
* User run bootstrapper
* The bootstrapper start and read it's options from all it's sources, including the `version` line in `paket.dependencies`, merging them into one set of options. If neither environment variables or the command line specifies a version **1.2.3.4** is used. Prefer nuget is always used as there is no way to disable a flag.
* It then download paket or not as usual (If already present on that version or not)
* User start paket
* Paket read the same `paket.dependencies` file completely ignoring the `version` marker

Where it really shine is when you combine this, the `--run` parameter and the magic mode (That is nothing else than an implicit `--run` parameter depending on the name of the executable) :

* `paket.dependencies` contains `version 1.2.3.4`
* User run `paket.exe restore`
* The bootstrapper start and read it's options from all it's sources, including the `version` line in `paket.dependencies`.  If environment variables doesn't specify a version **1.2.3.4** is used.
* It then download paket or not as usual and place it in the temp folder (that's activated by "magic" mode)
* Paket is started with the "restore" parameter.
* Paket read the same `paket.dependencies` file completely ignoring the `version` marker

agross:
Hm, if `--prefer-nuget` is always enabled, why do I need to specify it? It looks like a command line parameter to me.

Say for example paket.bootstrapper.exe v2017 supports `--fallback <url>`. Would it be possible to have something like this in paket.dependencies?

```
version 1.2.3 --prefer-nuget --fallback http://example.com/paket.exe
````

vbfox:
> Hm, if --prefer-nuget is always enabled

It's not always enabled, what's after `version` is the same as the command line of the bootstrapper and just treated as another option source.
I was noting that it was always enabled there because the bootstrapper doesn't support negative options (Like an hypothetical `--no-prefer-nuget`) so once one option source specify that a Boolean is different from it's default there is no way to go back so if paket.dependencies specifies `--prefer-nuget` there is no way to use github from the command line. (But it's not specific to this change, it already happened before between appsettings and the command line)

> Say for example paket.bootstrapper.exe v2017 supports `--fallback <url>`. Would it be possible to have something like this in paket.dependencies?

Yes, see the code of `ArgumentParser.cs` in this PR, the line in paket.dependencies is parsed using the new `FillNonRunOptionsFromArguments` and it's the same function that parses the normal command line. The only difference is that `--run` isn't possible for the rest of the arguments it's the same code.

The only real complexity is in correctly splitting the arguments into a `string[]` but that's the role of `WindowsProcessArguments.Parse` that implements how windows processes normally do it (The OS doesn't provide it and anyway as we want paket.dependencies to be shareable between OSes we must fix ourselves on one algorithm)

agross:
@vbfox Alright, now I got it. Thanks!

forki:
http://fsprojects.github.io/Paket/dependencies-file.html#Required-Paket-version this OK?

Will announce it soon

vbfox:
Yes 👍 

Maybe the text could be a little clearer in specifying that the version can be the special "prerelease" string but maybe we don't want to incite users to always get prereleases 😁

forki:
please send PRs to make the docs graet again

2016-10-28 15:07 GMT+02:00 Julien Roncaglia <notifications@github.com>:

vbfox:
Trump

Issue #1984, opened 2016-10-27
Authentication configurations expose passwords


haraldsteinlechner:
### Description

Exceptions and their stdout might print passwords in plaintext

### Repro steps

I have no repro steps, the print occured on one of our managers computer and leaked his password to me.

### Expected behavior

do not print plaintext password in exception texts

### Known workarounds

I suspect the complete plaintext authentication sheme is not really secure. Still, having plaintext passwords printed to screen renders paket really untrustworthy ;) 

I'd suggest simply to put StructuredFormatDisplay to authenication configs.

haraldsteinlechner:
i think this might be the right location https://github.com/fsprojects/Paket/pull/1985

cdrnet:
We seem to run into this from time to time, see also #1224, #1357. Maybe it's time to consider switching to `SecureString` to prevent this from happening in the first place?

haraldsteinlechner:
+1

cdrnet:
(kind of fitting that this issue has id 1984...)

PR #1985, opened 2016-10-27
custom print for NugetSourceAuthentication types

haraldsteinlechner:
not to leak passwords, see https://github.com/fsprojects/Paket/issues/1984


forki:
thx!!

Issue #1986, opened 2016-10-27
Paket reports DependencyParseError when paket actually fails to parse package sources


haraldsteinlechner:
### Description

Whenever, for whatever reason a package source cannot be parsed paket reports parse error in paket.dependencies which is misleading for users.
### Repro steps

Destroy a package source authentication info xml and observe the parse error.
### Expected behavior

Provide more diagnostic message.
### Known workarounds

My pull request tracks internal parse errors and reports them via DomainMessage.

haraldsteinlechner:
see https://github.com/fsprojects/Paket/pull/1987

haraldsteinlechner:
@forki i think this can be closed?

PR #1987, opened 2016-10-27
DependenciesFileParser now tracks inner exceptions for package sources

haraldsteinlechner:
see https://github.com/fsprojects/Paket/issues/1986


forki:
awesome. thanks

PR #1988, opened 2016-10-28
Fix build error in Paket.Tests

inosik:
I added explicit type information to the `NuspecWriterSpecs.fs` file, and this fixes the build for me at least.

I think the issue is that the Travis build uses a F# 4.0 compiler (https://travis-ci.org/fsprojects/Paket/builds/171333514#L1582), and I have a F# 4.1 compiler on my Mac, as it is bundled with Mono, installed via Homebrew. Seems there are some changes in the type inference stuff.

@piedar could you please try to build this branch and tell us if it succeeds for you now?

### DIFF
--- a/tests/Paket.Tests/NuspecWriterSpecs.fs
+++ b/tests/Paket.Tests/NuspecWriterSpecs.fs
@@ -18,8 +18,8 @@ let ``should serialize core info``() =
     <description>A description</description>
   </metadata>
 </package>"""
-    
-    let core = 
+
+    let core : CompleteCoreInfo =
         { Id = "Paket.Tests"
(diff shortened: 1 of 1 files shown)


forki:
urgs: @dsyme @vasily-kirichenko @kevinransom is this expected? did we chage type inference in 4.1?

piedar:
:+1: Works for me.

inosik:
@piedar Do you have F# 4.1 on your machine, too? The first line of `fsharpc --help` should be something like this:
```
F# Compiler for F# 4.1
```

piedar:
@inosik Yes, I have F# 4.1 built from the [4.0.1.15 release](https://github.com/fsharp/fsharp/releases/tag/4.0.1.15).

PR #1997, opened 2016-11-02
Update the FAQ

monkieboy:
Indicate that the paket-files directory and packages directory should not be committed to source control.


forki:
thanks

PR #1999, opened 2016-11-02
Make download loop to terminate in max N=5 iterations

### LINKED ISSUE #1998: paket -v install enters an infinite loop "license already downloaded"
### Description

`paket -v install` enters an infinite loop on Paket version 3.26.0.0

### Repro steps

Reproducing is hard. This requires a source with basic HTTP auth that somehow malfunctions. It is easy to see where the loop is happening in source code though.

### Expected behavior

Terminate with an error

### Actual behavior

Infinite loop.

### Known workarounds

NA.

t0yv0:
Fixes #1998


forki:
thanks!

forki:
will be released later today.

PR #2007, opened 2016-11-06
[WIP] Verbose bootstrapper

vbfox:
This PR change the way `-s` work in the bootstrapper. There is now a verbosity level that can be decreased via `-s` and increased via `-v`.

The levels are :

* Silent -> Never display anything.
* ErrorsOnly -> Only errors are shown on the console (Default in magic mode)
* Normal -> Informative logs (Default in non-magic mode)
* Trace -> More logs, especially timing for most operation is displayed.

**This PR change the default behavior, before a single `-s` resulted in the boostrapper being silent but 2 are now needed for the same result**


forki:
can you please make it compile with older C#? I have trouble in VS2013

PR #2026, opened 2016-11-18
Report which nuspec file is invalid when the nuspec cannot be loaded

busterwood:
I had a nuspec file in the `packages` folder that was empty (zero bytes) , paket failed with an XML error, but did not report *which* file had an error.  This change will report which nuspec file had the problem, and what the problem was.


forki:
Cool. thx

PR #2027, opened 2016-11-18
Line matching pattern refactoring

tomasdeml:
As the matching pattern reached maximum number of possible partitions and also got a bit unwieldy, I took the liberty of converting it to the partial pattern. It will be easier to introduce new line patterns in future.


forki:
cool stuff. thanks

Issue #931, opened 2015-07-23
When to use paket.dependencies and paket.references?


exercitusvir:
I've read the documentation on paket.dependencies and paket.references, but it's still not clear to me which one I need to edit. It says that paket.dependencies specifies the rules for the application's dependencis. paket.references adds references to an MSBuild project and apparently it's optional.

But what happens when there are no paket.references? Will Visual Studio still recognize the dependencies? Should I place paket.dependencies in each project folder?

After automatically converting my package.config files with paket, I now have a single paket.dependencies file in the root of my solution and paket.references files in each project folder of my solution. If I want to add dependencies, do I only need to add them to the paket.references files? Will paket.dependencies in the solution folder be then updated automatically?

Basically, I don't understand the possible interaction between paket.dependencies and paket.references.

forki:
paket.dependencies is a global view of your dependencies. there is no such thing in nuget.

paket.references contains all direct dependencies of a csproj or fsproj. It's not optional. You need to specify that in order to get the references in your project.

so in short: you need to edit both, but commands like `paket add` help you there.

isaacabraham:
paket.dependencies = solution wide set of dependencies, not related to any project.
paket.references = dependencies required by a specific fsproj / csproj project etc.

exercitusvir:
Thanks for the quick answers! They really helped clarify their purpose and relationship. You may want to add that to the index page.

I guess I was confused because it explicitly says that it is optional in [The paket.references files](http://fsprojects.github.io/Paket/references-files.html):
> It acts a lot like NuGet's packages.config files but there are some key differences:
...
> - **It's optional**

Is that a mistake or is that meant differently?

forki:
@exercitusvir please send a pull request with whatever you think would clarify the docs. We need the help of newcomers here.

isaacabraham:
It *is* optional insofar as you don't need it just to manage dependencies. I often use just a dependencies file when working with F# scripts when I don't have a project.

You *do* need it if you want to marry dependencies to VS projects (AFAIK).

exercitusvir:
OK. Let me get this working first and then I'll try to improve the documentation for beginners.

One more question. Is there a command to check all paket.references files and update the paket.dependencies file correspondingly? That way, I would only have to edit the paket.references files and run paket with that command just once. 

Another suggestion would be to add an overview of the commands with a short description. I only found the list on the right of the index page where it seems that I have to open each page of a command to find out what it's for.

[...]

PR #2028, opened 2016-11-18
Add clarifications regarding paket.dependencies purpose

eugene-g:
Newcomer here. As #931 shows, I'm not alone in confusing paket.dependencies and paket.references, seeing them for the first time. I hope that change would help others to quicker get comfortable with Paket.


forki:
Thx this is basically correct, but can you please try to say that it only needs to contain the top level dependencies? I mean we don't have to specify "all" dependencies.

eugene-g:
@forki, indeed, "all dependencies" is misleading. Is it better now?

forki:
Thx!

PR #2031, opened 2016-11-20
Fix package name displayed when package is found in different group

TeaDrivenDev:
When a package is not found in the same group in `paket.lock` and a project's `paket.references` during `paket install`, the package name is displayed as `<StartupCode$Paket-Core>.$LockFile+CheckIfPackageExistsInAnyGroup@521-5` in the error message. This PR fixes that.


forki:
thanks!

PR #2035, opened 2016-11-22
Allow override of NuGetCacheFolder location through environment variable

chappoo:
The NuGet CLI (2.x) provides an environment variable override for specifying the location of the NuGet cache.  This is key when running multiple processes that pull NuGet packages to avoid file locks in the cache (i.e. in a multi- build agent per server environment).  This change enables support for the override in Paket, via an environment variable (e.g. Windows PS):

`$env:NuGetCachePath = "C:\Custom\Cache\Location"`


forki:
can you please add this to the docs as well?

chappoo:
Done - also added a bit of extra blurb about the machine cache, as it seems it's used regardless of any additional caches.  If you'd prefer it elsewhere, happy to do so?

chappoo:
Also, looks like there's problem with the CI checks.. unrelated?

forki:
awesome. thx

PR #2036, opened 2016-11-22
Fix netcore build with workaround

enricosada:
apply workaround: Move fsproj for preview3 in another directory.


enricosada:
annoying. locally build ok.

I'll leave there, because i cannot fix it tonight, maybe you can do it @forki

I tested locally the error.
on my local was working if i have `preview3` in `PATH`
and fail with `preview2` in `PATH`, same error.
I think that's because you are just running `dotnetcore/dotnet.exe` in FAKE script instead of adding `dotnetcore` dir to `PATH` env var **before** calling `dotnet build`.
Like that, in the target file, when i run `dotnet compile-fsc` is going to use the `dotnet` in `PATH` (pretty sure is preview2, so doesnt find `compile-fsc`) instead of the one inside `dotnetcore` (the `preview3`).

.net core sdk require `dotnet` sdk directory in `PATH`, that's how things works atm to resolve correct `dotnet.exe`. directly invoking `dotnet.exe` is not enough (there is also a discussion about that in FAKE)

enricosada:
tried the proposed fix, let's see

forki:
cool thanks

Issue #2037, opened 2016-11-22
Push returns "(500) Internal Server Error" when pushing to VSTS on mono/linux


varon:
### Description
#2034 fixes #2033, allowing me to now push to VSTS from windows.

However, my CI server running on Linux/Mono receive a 500 server error from VSTS when pushing up a package.

I'm not quite sure it's an upstream bug, given the response we receive.

Versions:
Mono:  `Mono JIT compiler version 4.4.2 (Stable 4.4.2.11/f72fe45 Fri Jul 29 09:58:49 UTC 2016`
Linux: `Debian Jessie 8.6`
Paket: `Paket version 3.29.4`

### 1. Add credentials for the NuGet Feed
```
> ./paket.exe config add-credentials https://placeholder.pkgs.visualstudio.com/_packaging/Test2/nuget/v2
Paket version 3.29.1
Username: placeholder
Password: ************
```
Here the password is my personal access token created above.

I can confirm that after step 1, I have a valid paket.config file in the expected location.

### 2.  Push a package to VSTS
```
./paket.exe --verbose push apikey VSTS file placeholder.nupkg url https://placeholder.pkgs.visualstudio.com/_packaging/Test2/nuget/v2
```
[According to VSTS documentation, the api key can be any non-empty string.](https://www.visualstudio.com/en-us/docs/package/nuget/publish)

### Expected behavior

This works fine on Windows/.Net, so I'd expect it to be the same from Mono.
Perhaps we need to do something differently on the WebClient?

### Actual behavior

A (500)  Internal Server Error is received.

`Could not push placeholder.nupkg: The remote server returned an error: (500) Internal Server Error.`

The response body was:
```
{
    "$id":"1",
    "innerException":null,
    "message":"Unexpected end of MIME multipart stream. MIME multipart message is not complete.",
    "typeName":"System.IO.IOException, mscorlib",
    "typeKey":"IOException",
    "errorCode":0,
    "eventId":0
}
```

### Known workarounds

None

varon:
Solved by #2038

PR #2038, opened 2016-11-22
Workaround for Paket.Push 500 error on VSTS + Linux

varon:
Solves #2037.

This is a simple fix that shouldn't cause any issues (seeing as we're already using CLRF from windows previously).


forki:
thx

Issue #2030, opened 2016-11-20
paket4 install crashes with stack-overflow exception


cloudRoutine:
### Description

I've been using Paket4 in my migration of the netcore port of FAKE, and the parts that work are great, [once you get past some of the initial pitfalls you might stumble into](https://github.com/fsprojects/Paket/issues/2029)

### Repro steps

[paket-install-stackoverflow.zip](https://github.com/fsprojects/Paket/files/601894/paket-install-stackoverflow.zip)

1.  unzip and cd into repo
1. `.\.paket\paket.bootstrapper.exe 4.0.0-alpha024`
1. `.\.paket\paket.exe install`
1. Let it run until 💥 

[ Log of packages installed before stack overflow exception ](https://gist.github.com/cloudRoutine/c15d682028ccfbfbcc7b203ed48b5d57#file-log-ini) 

### Expected behavior

paket installs packages without crashing

### Actual behavior

paket crashes during install

### Known workarounds

haven't found one yet, will update if I do

forki:
weird. can't reproduce.

cloudRoutine:
@forki any ideas about how I could get around this? I ran it 10 times, every time it ended in

![](http://i.imgur.com/3jDmBNt.png)

forki:
can you try to debug it?

cloudRoutine:
@forki  version `4.0.0-alpha030` is still blowing the stack on me

this time on the **paket.dependencies** file 

```
source https://nuget.org/api/v2

github logary/logary src/Logary.Facade/Facade.fs 4b21d888e83d72d1238ab0f36110e2a3946240e7

nuget Argu
nuget FsCheck
nuget FSharp.Core ~> 3
nuget FSharpx.Core
nuget BenchmarkDotNet
nuget NuGet.CommandLine
nuget Microsoft.NET.Sdk alpha 
nuget FSharp.NET.Sdk 1.0.0-alpha-000007 alpha framework: netstandard1.6
nuget NETStandard.Library 1.6.1 framework: netstandard1.6
nuget System.Diagnostics.TraceSource

group netcore 
    source https://www.nuget.org/api/v2

    nuget FSharp.Core 4.0.1.7-alpha alpha framework: netstandard1.6

group build
    source https://www.nuget.org/api/v2

    nuget FAKE
```

![](http://i.imgur.com/ISgQyab.png)

but at least it happens really fast now 😆 
it's gone from 10m down to ~10s

forki:
[...]

Issue #144, opened 2014-09-21
Could not resolve dependencies despite non-empty intersection


cdrnet:
Probably a misunderstanding on my side, so this may be something for the FAQ:

Situation:

In my solution I have some projects using NUnit only, and some using FsUnit which in turn depends on NUnit. So NUnit is both a direct and an indirect dependency.

When I add the following to paket.dependencies:

```
nuget NUnit ~> 2.6
nuget FsUnit ~> 1.3
```

paket update returns:

```
Paket failed with:
   Could not resolve dependencies.
Dependencies file depends on
  NUnit (>= 2.6, < 3.0)
FsUnit 1.3.0.1 depends on
  NUnit (>= 2.6.3)
```

I would have expected it to resolve NUnit to the largest version matching (>= 2.6.3, < 3.0), i.e. currently 2.6.3. Am I misunderstanding the version specifiers in paket.dependencies?

PR #2039, opened 2016-11-23
PackageResolver Refactor

cloudRoutine:
Started working on this in an attempt to address https://github.com/fsprojects/Paket/issues/2030

I wasn't able to get the step function to the point where its tail recursive, but its much closer than it was before and it no longer relies on mutation across iterations.

I think it should be possible by adding a third case to 
```fsharp
type private Stage =
    | Outer of conflictState : ConflictState
    | Inner of conflictState : ConflictState
```
and using it as an argument for `step` to direct it down one of three pathways for the start of a step, an outer loop, or an inner loop, although when I tried coding that up I couldn't get it to work.

I think this is the block of code that was causing the issues when I took that approach
```fsharp
                    match conflictState.Status with
                    | Resolution.Conflict(_,conflicts,_,_)
                        when
                            (Set.isEmpty conflicts |> not)
                            && nextStep.CurrentResolution.Count > 1
                            && not (conflicts |> Set.exists (fun r ->
                                r = currentRequirement
                                || r.Graph |> List.contains currentRequirement)) ->
                        let flags = StepFlags(flags.Ready,flags.UseUnlisted,flags.HasUnlisted,true,flags.FirstTrial)
                        stepLoop flags (Inner (conflictState))
                    | _ ->
                        stepLoop flags (Inner (conflictState))
```

but I'm not totally sure.

### DIFF
--- a/src/Paket.Core/PackageResolver.fs
+++ b/src/Paket.Core/PackageResolver.fs
@@ -125,7 +125,8 @@ module Resolution =
         match conflicts with
         | [] -> errorReport
         | req::conflicts ->
-            errorReport.AddLine (sprintf "  Could not resolve package %O:" req.Name)
+            
+            errorReport.AddLine (sprintf "  Could not resolve package %O %O:" req.Name req.VersionRequirement.Range)
             let hasPrereleases =
                 conflicts |> List.exists (fun r -> r.VersionRequirement.PreReleases <> PreReleaseStatus.No)
 
@@ -611,11 +612,11 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
             ,   firstTrial  = true
             )
 
-    let stopLooping (flags:StepFlags) (conflictState:ConflictState) =
-        if flags.ForceBreak then true else
-        if conflictState.Status.IsDone || Seq.isEmpty conflictState.VersionsToExplore then true else
-        if (flags.FirstTrial || Set.isEmpty conflictState.Conflicts) then false
-        else true
+    let keepLooping (flags:StepFlags) (conflictState:ConflictState) =
+        if flags.ForceBreak then false else
+        if conflictState.Status.IsDone || Seq.isEmpty conflictState.VersionsToExplore then false else
+        flags.FirstTrial || Set.isEmpty conflictState.Conflicts
+        
 
 
 
@@ -688,7 +689,7 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
                 stepLoop flags (Inner(conflictState))
 
             | Inner (conflictState) ->
-                if stopLooping flags conflictState then
+                if not (keepLooping flags conflictState) then
                     let flags =
                         if not flags.UseUnlisted && flags.HasUnlisted && not conflictState.Status.IsDone then
                             StepFlags(flags.Ready,true,flags.HasUnlisted,flags.ForceBreak,flags.FirstTrial)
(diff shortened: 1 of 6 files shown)


src/Paket.Core/PackageResolver.fs
```diff
@@ -94,90 +99,111 @@ type ResolverStep = {
...
+    let getConflicts (res:Resolution) =
+        match res with
+        | Resolution.Ok _ -> []
+        | Resolution.Conflict (currentStep,_,lastPackageRequirement,_) ->
             currentStep.ClosedRequirements
             |> Set.union currentStep.OpenRequirements
             |> Set.add lastPackageRequirement
-            |> Seq.filter (fun x -> x.Name = lastPackageRequirement.Name)
-            |> Seq.sortBy (fun x -> x.Parent)
+            |> Seq.filter ^ fun x -> x.Name = lastPackageRequirement.Name
```
forki: please don't

forki:
this is really really hard to review- can you please split it in two PRs. one with the resolver change and one with all the cosmetical changes? thanks

cloudRoutine:
@forki this fine?

forki:
it seems the new resolver is not resolving correctly tests like ##1254 are red. can you confirm?

cloudRoutine:
a bug got reintroduced when I redid the commits

forki:
#1883 install FSharp.Core from Chessie fails!?

cloudRoutine:
I've been having issues running the integration tests on my machine

forki:
yeah these new packages are fucking up everything. MS is clearly out of control here.

cloudRoutine:
even when I try to run the [integration tests on the core3 branch they don't complete](https://gist.github.com/cloudRoutine/347eb5b83b6078158f825183ad3b22e9)

forki:
do you have VS open at that time? or something else that keeps an handle on the nupkg?

cloudRoutine:
I had all of that stuff closed at the time. The first time I got those errors I thought that might be the source but when I tried again with everything off it still happened.

I get these errors when I run the tests just in VS too

forki:
And from fake build?

Am 23.11.2016 16:50 schrieb "Jared Hester" <notifications@github.com>:

cloudRoutine:
the log in the gist is from the FAKE build

forki:
Ok can you please try to find out which tool is locking it? Sorry for all the work, but this whole .net standard conversion is a real mess.

forki:
https://github.com/fsprojects/Paket/pull/1977 seems to work on appveyor. (travis is expected to fail)

cloudRoutine:
The readme might need a note to install F# 3.0, without it `#144` will fail with



But even after I installed F# 3.0

cloudRoutine:
now I'm really confused about where this error is coming from
when I run the contents of test 1883 in FSI

cloudRoutine:
A more detailed output

src/Paket.Core/PackageResolver.fs
```diff
@@ -257,313 +283,489 @@ type UpdateMode =
...
+
+                match getExploredPackage packageDetails getPackageDetailsF exploredPackages with
+                | None ->
+                    stepLoop flags (Inner(conflictState))
+                | Some exploredPackage ->
+                    let hasUnlisted = exploredPackage.Unlisted || flags.HasUnlisted
+
+                    let flags = StepFlags(flags.Ready,flags.UseUnlisted,hasUnlisted,flags.ForceBreak,flags.FirstTrial)
+                    if exploredPackage.Unlisted && not flags.UseUnlisted then
+                        tracefn "     unlisted"
```
cloudRoutine: now i'm pretty sure the bug is here, where I last made a change :facepalm:
which I'd had as 
```fsharp
                    if exploredPackage.Unlisted && not flags.UseUnlisted then
                        tracefn "     unlisted"
                        stepLoop flags (Inner(conflictState))
                    else
```
before, but that caused an infinite loop printing `unlisted`

@forki it'd be really useful for whoever has to work on this next, if you or someone else with a deeper understanding of what's going on in this module could add some comments explaining the processes. I'd do it myself, but I don't think I have a firm enough grasp and I wouldn't want to put down something that's wrong

cloudRoutine:
@forki finally passed on appveyor 🎉

cloudRoutine:
anything else I need to do for this?

forki:
thanks so much. let's what the real world thinks about it ;-)

PR #2061, opened 2016-11-30
Fix of Git dependency change detection

tomasdeml:
Change detection of Git dependencies does not work correctly for two or more dependencies as their `RemoteFileChange`s have `Name` set to "". This causes mismatch of dependencies during change detection. This PR fixes the bug by including remote's `Origin` in comparison.


forki:
thanks!

Issue #2059, opened 2016-11-30
References are not added to projects with "paket add nuget X"


Gonnagle:
### Description

I've assumed that running `paket add nuget X` would add the package for all projects in the given solution. It seems that the package is only added to the group main, but not at any of the projects. Is this broken or have I just misunderstood how the command is intended to work?

### Repro steps

I've used the following simple repo: https://github.com/Gonnagle/paket-tests/releases/tag/vanilla
 
1. `paket add nuget xunit`

### Expected behavior

In addition to modification on `paket.dependencies` and `paket.lock`, I'd expect that the nuget is added to `ConsoleApplication/paket.references` and it is added (with it's transitive dependencies) as reference to the project (`ConsoleApplication.csproj`).

### Actual behavior

Only `paket.dependencies` and `paket.lock` are updated. Nothing else is changed.

### Known workarounds

With interactive mode and when targeting the project the references are added as expected

* `paket add nuget xunit --interactive`
* `paket add nuget xunit project ConsoleApplication`

forki:
No add command does not add to a project if you don't ask it to do that

Gonnagle:
So is there any way to add package to all projects (besides the interactive mode)?

The docs say that

> It's also possible to add a package to a specified project **_only_**:

I guess that the word **_"only"_** led me to interpret that without defining a specific project it would be added to all of the projects.

forki:
Open references file in editor. Imho it's really not a common use case to
install to all projects.

Am 30.11.2016 2:08 nachm. schrieb "Mikke Tanska" <notifications@github.com>:

So is there any way to add package to all projects (besides the interactive
mode)?

The docs say that

It's also possible to add a package to a specified project *only*:

I guess that the word *"only"* led me to interpret that without defining a
specific project it would be added to all of the projects.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/fsprojects/Paket/issues/2059#issuecomment-263869128>,
or mute the thread

Gonnagle:
Yeah, that's true. I'll drop a pull request for updated docs tomorrow (at least the removal of word "only" that I find misleading).

forki:
You could even try to come up with an additional parameter which would do
that. Don't think it's too hard to do

Am 30.11.2016 2:28 nachm. schrieb "Mikke Tanska" <notifications@github.com>:

Gonnagle:
[...]

PR #2062, opened 2016-12-01
Docs: Clarify that package is not added to projects by default

Gonnagle:
Attempt to make it clearer how the add function works (related to #2059)


forki:
Thanks

PR #2067, opened 2016-12-04
Package names are compared using non-linguistic Ordinal comparison

tomasdeml:
Test `ParserSpecs.should parse lock file with depdencies` fails under OS with Czech regional settings because of linguistic comparison of `Chessie` package name. In Czech, [`ch` is a diagraph that is inserted between _H_ and _I_ letters in collation order](https://en.wikipedia.org/wiki/Ch_(digraph)#Czech).  I fixed the issue by using non-linguistic ordinal comparison.

### DIFF
--- a/src/Paket.Core/Domain.fs
+++ b/src/Paket.Core/Domain.fs
@@ -1,5 +1,6 @@
 ﻿module Paket.Domain
 
+open System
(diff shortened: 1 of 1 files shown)


forki:
cool. thanks a lot!

Issue #2018, opened 2016-11-14
Semantic version resolving is broken


tsibelman:
### Description

In 3.27.0 of paket mechanism of version resolving got broken, I have an internal
nugget, what has versions 2.0.0.9 and 2.0.10 I defined dependency on it as ~> 2.0, previously paket resolved it to 2.0.10 as it should now it resolves it to 2.0.0.9.

Sorry I can't give you repro because it internal nugget but I know that it got broken in 3.27.0

forki:
probably related to https://github.com/fsprojects/Paket/pull/2008
@cdrnet I will revert that for now. please open new PR that just adds that for your server

tsibelman:
Thank you @forki

forki:
@tsibelman what kind of server do you have?

tsibelman:
This one https://www.nuget.org/packages/NuGet.Server/

tsibelman:
Took the latest all works now, thank you.

cdrnet:
And as a result this is now broken for all Artifactory users again. Welcome to the brave world of OData where it seems impossible to work around one broken implementation without breaking the work around of another broken implementation...

Do we know what exactly was broken with NuGet.Server? Being opensource, maybe this could actually be fixed properly over there. I expect the NuGet Package Explorer doesn't work with this server then?

@forki will do

enricosada:
> Took the latest all works now, thank you.

Closing

PR #2073, opened 2016-12-09
NuGetV2-OData: retrieve versions in descending order for artifactory

cdrnet:
Related to #2008 which was reverted in #2018 because it affected non-artifactory servers negatively (in an unknown way).

This change adds an alternative variant of tryGetAllVersions to fetch the versions in descending order (by publish date) instead of modifying the original one.

GetVersions chooses to use the alternative variant in case the source URI contains the string "artifactory". This is a very unreliable way of discovering a server to be an artifactory one, but is in line with the existing rules and should prevent affecting non-artifactory servers.

Ideally the decision should be more controllable than URI matching, e.g. by introducing the concept of source options (explicitly in the dependencies and lock files).


forki:
thanks so much!

PR #2080, opened 2016-12-15
One of the file writes was missing a Directory.Create()

### LINKED ISSUE #1927: generate-include-scripts fails
### Description

paket generate-include-scripts fails when FAKE is the only dependency in a group

### Repro steps

```
framework: net45
nuget Newtonsoft.Json

group Build
framework: net45
nuget FAKE
```

`paket generate-include-scripts framework net45 type fsx`

Produces this error:

`Could not find a part of the path "paket-files/include-scripts/net45/build/include.build.group.fsx".`

jamesjrg:
1. generate-include-scripts used ToLowerInvariant() on the full path name which caused issues in Linux that wouldn't appear in Windows. Though apparently wasn't an issue on Macs? not sure why that would be, I don't have a mac.

2. At first I thought the issue was the lack of Directory.Create() before File.WriteAllText for some files, but turned out this wasn't the issue due to file creation order meaning the directory was actually already there.

Regardless I changed it so in the function to write fsharp scripts it does a Directory.Create immediately before writing, which makes the functions to create fsharp and csharp scripts more consistent, and I think it's less confusing

3. The Linux CI build is broken because of an issue with the .NET Core build, I have the same  problem locally

### DIFF
--- a/src/Paket.Core/ScriptGeneration.fs
+++ b/src/Paket.Core/ScriptGeneration.fs
@@ -360,7 +360,8 @@ module ScriptGeneration =
 
       let getGroupFile group = 
         let folder = getScriptFolder includeScriptsRootFolder framework group
-        FileInfo(Path.Combine(folder.FullName, sprintf "include.%s.group.%s" (group.GetCompareString()) extension).ToLowerInvariant())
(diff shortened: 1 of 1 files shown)


forki:
thanks!

PR #2088, opened 2016-12-18
Fix displayed package name for packages found in another group

TeaDrivenDev:
This is the same problem and fix as #2031, just in a different place. I did a regex search over the solution for more occurrences but couldn't find any others for the `PackageName` DU.


forki:
thx

PR #2089, opened 2016-12-18
Don't remove group with only remote files

danielrbradley:
Created the test first and confirmed it failed. I've not yet had time to see if there are any similar bugs in other areas of the dependencies code - only taken a look at the removal case so far.

Edit: I've also added an extra fix for building on a path with spaces, though this still fails after due to a similar bug in the dotnet core targets file which I've reported here: https://github.com/dotnet/netcorecli-fsc/issues/59


forki:
cool thx!

Issue #2085, opened 2016-12-18
Paket "auto-restore on" corrupts VS2017 csproj


cjbhaines:
### Description

Paket version: 4.0.0-alpha033

When enabling auto-restore for a .NET Core project, Paket adds a namespace to the import which is invalid in the schema:

```
<Import Project="..\.paket\Paket.Restore.targets" />
<Import Project="..\.paket\paket.targets" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" />
```

error  : The attribute "xmlns" in element <Import> is unrecognized.  C:\Temp\PaketTest\packages_unlisted\ConvertTest\ConvertTest.csproj

Is auto-restore meant to work with .NET Core projects? 

### Repro steps

https://dl.dropboxusercontent.com/u/49591225/Code/PaketTest_UnknownNamespace.zip

Unzip solution, then run:

```
dotnet restore
.paket\paket auto-restore on
```
### Expected Behaviour

Paket does not add an unexpected namespace.

forki:
Good point. We don't need to do that in core projects.

cjbhaines:
Just updated to Paket version: 4.0.0-alpha034 and tried a convert form nuget on one of my real solutions. It added the following targets:

```
<Import Project="..\..\..\.paket\paket.targets" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" />
```

Instead of the new .NET Core ones:

```
<Import Project="..\..\..\.paket\Paket.Restore.targets" />
```

bhugot:
Hello with the paket version:  4.0.0-alpha035 when we do install in a project the following import is added to csproj

  <Import Project=".paket\Paket.Restore.targets" xmlns="http://schemas.microsoft.com/developer/msbuild/2003" /> but xmlns is not recognize and if I remove this attribute so I have  only <Import Project=".paket\Paket.Restore.targets" /> then it's working without any problem.

forki:
do you have the full fsproj for me?

forki:
@bhugot could you please try latest?

bhugot:
@forki still the same, the proj is a simple dotnet create -t console on which I did convert from nuget with the last prerelease. I have some issue accessing to github from my computer at work why do not use
`let createNode name (project:ProjectFile) = 
        project.Document.CreateElement (name, project.Document.DocumentElement.NamespaceURI)`

forki:
That's what we do in latest paket 4

[...]

PR #2093, opened 2016-12-20
null is awaited so no need to test for null #2085

bhugot:


src/Paket.Core/ProjectFile.fs
```diff
@@ -200,17 +200,8 @@ module ProjectFile =
...
-        let namespaceURI = 
-            try
-                project.Document.DocumentElement.NamespaceURI
-            with
-            | _ -> null
-        if String.IsNullOrWhiteSpace namespaceURI then
-            project.Document.CreateElement (name, Constants.ProjectDefaultNameSpace)
-        else
-            project.Document.CreateElement (name, namespaceURI)
+        project.Document.CreateElement (name, project.Document.DocumentElement.NamespaceURI)
```
forki: are you sure this doesn't crash on weird input?

bhugot: It s working on dotnet core csproj like a charm and tests are passing but 2 who where already down

forki:
Can you please still put it under try catch? I have a weird feeling here.

Am 20.12.2016 9:52 nachm. schrieb "Benjamin Hugot" <notifications@github.com

PR #2097, opened 2016-12-23
paket outdated: group -parameter added.

Thorium:
Support for group -parameter for "paket outdated", to check just a single group of outdated packages.


Thorium:
I can't indentify Travis failure "dotnet failed" related to my source code modification. Can you request it again?

PR #2100, opened 2016-12-27
Correctly escape no_proxy domains

juergenhoetzel:
Bootstrapping fails on my setup, because I use wildcard in `NO_PROXY`:

```
→ echo $NO_PROXY 
*.fritz.box

→ ./obj/Debug/paket.bootstrapper.exe
Checking Paket version (downloading latest stable)...
The type initializer for 'Paket.Bootstrapper.EnvProxy' threw an exception. (Github - cached)
```

This is actually the same issue as https://github.com/fsprojects/Paket/pull/1939 which also need to be fixed in the bootstrapping code.

### DIFF
--- a/src/Paket.Bootstrapper/EnvProxy.cs
+++ b/src/Paket.Bootstrapper/EnvProxy.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Net;
+using System.Text.RegularExpressions;
+using System.Linq;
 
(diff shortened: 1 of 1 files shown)


forki:
thanks!

forki:
This broke one of the nunit bootstrapper tests. Please take a look at the appveyor build output (Travis is red for unrelated reason). Can you please send me a PR with a fix? Thanks

PR #2101, opened 2016-12-30
Improve subcommand document generation

eiriktsarpalis:
Did a bit of tweaking in document generation code to improve subcommand documentation. What currently looks like

has been changed to this


forki:
awesome

PR #2102, opened 2016-12-30
Fix "directory doesn't exist" in NugetV2

MrLuje:
When doing a paket restore using a local nuget repository that is not reachable, the process crashes without telling which package or feed cause the issue. There's already a check if the package is not present in the repository but not if the repository itself is unreachable.

Had the issue with a repo in the paket.lock that can't be reached from another server.


forki:
thanks!

PR #2103, opened 2017-01-02
Fix bootstrapper test broken by #2100

juergenhoetzel:
The BypassList property contains an array of regular expressions.


forki:
Thx.

PR #2105, opened 2017-01-06
Ensure stable ordering of references in the same ItemGroup

flin-zap:


forki:
It looks like you need to fix the tests

forki:
what was the error that lead to this?

flin-zap:
There were no errors, but we have a large c# solution, and every time we add a paket dependency to one of the projects and run "paket install", many of the other csproj files that have no dependency changes would have their references reordered

forki:
Interesting. I wonder why that would happen.

flin-zap:
Sorry it doesn't happen every time we add a paket dependency. it's probably quite niche, as I've been struggling to repro the issue on demand. 

It could be a combination of changing project references + paket update that somehow changes the order of the references?

I just looked through the history of this file, a sortBy PATH was added on July 10 (fd3c3ac2), and removed on August 12 (e580b8f9). Do you know if that was intentional?

flin-zap:
Here's a sample of a diff from before
image

PR #2106, opened 2017-01-10
Ensure case when getting packages from nuget feed

MrLuje:
Fix an issue when comparing a nuget name against a lowered nuget id in odata query that may have different case


forki:
thanks!

PR #2110, opened 2017-01-17
Make config writer use XmlWriter for disk write

### LINKED ISSUE #2003: Initial creation of paket.config produces malformed file
### Description

Running `paket config add-credentials` with no existing paket.config file produces a malformed paket.config file (encoding in XML header does not match actual file encoding) which subsequently cannot be read in by paket. Seems to be related to #1954

### Repro steps

1. Delete (or otherwise remove) your paket.config file
2. Use `paket config add-credentials <somedomain>` to add to the credential store
3. Run some paket command that hits the <somedomain> NuGet server (e.g. `paket update`)

Environment: Windows 10 Pro, 64-bit, paket 3.26.3 (also seen in 3.23.2).

### Expected behavior

Command run in step 2 creates valid paket.config file; command run in step 3 uses the credentials entered in step 2.

### Actual behavior

Command run in step 2 writes out a UTF-8 encoded paket.config file, but the `<?xml ... ?>` declaration at the top specifies it to be UTF-16. Command in step 3 is unable to read the file and exits with an error.

### Known workarounds

Edit the paket.config file after creation so that the `<?xml ... ?>` declaration specifies UTF-8, or have an existing valid paket.config file.

jdpage:
... instead of doing XmlWriter -> string -> disk.
- This guarantees that the file encoding and the XML declaration are
  properly in sync.
- Adds a `saveNormalizedXml` function to Utils to facilitate this.
- Adds a test to UtilsSpecs for the new function.
- Updates ConfigFile.saveConfigNode to use the new function.
- Should fix #2003 more permanently.
- Removes the previous fix in commit d1d06b7

### DIFF
--- a/src/Paket.Core/ConfigFile.fs
+++ b/src/Paket.Core/ConfigFile.fs
@@ -17,13 +17,6 @@ let private rootElement = "configuration"
 let private getConfigNode (nodeName : string) =
     let rootNode = 
         let doc = XmlDocument ()
-        
-        let xmldecl = doc.CreateXmlDeclaration("1.0",null,null)
-        xmldecl.Encoding <- "UTF-8"
-        xmldecl.Standalone <- "yes"
-
-        doc.InsertBefore(xmldecl, doc.DocumentElement) |> ignore
-
         if File.Exists Constants.PaketConfigFile then 
             try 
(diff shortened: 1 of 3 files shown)


forki:
cool. thanks!

PR #2111, opened 2017-01-18
first stab at restructuring Paket.Core project

smoothdeveloper:
to leverage *Folder Structure*

* Created a Files and Parsers folders, moved few modules there
* had to move NugetPackage record (which is very simple) out of PackagesConfigFile


smoothdeveloper:
Failure related to downloading dotnet-core preview 4, any idea @forki?

forki:
yeah I think @enricosada wanted to look at that.

PR #2112, opened 2017-01-18
A fix proposal for - Convert from nuget: Local package source give false error #1794

johannesvorster:
I was able to track down the error I was getting on my computer for issue #1794. 

The problem was that the path string that was pass to the Directory.Exists function had double quotes at the start and beginning.

I tested it on a Windows 10 computer, but was able to test it on a linux computer because of issue #1809 : Install problem recurred in Ubuntu.

I don't have access to a ios computer.

I hope it help.


theimowski:
Looks OK, wondering whether we could add this to [normalizePath](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Utils.fs#L672) utility? Also can we get rid of changes in solution and project file? Those seem irrelevant

johannesvorster:
Thanks for the feed back. Will move it to normalizePath utility and remove the changes in the solution and prject files. I was wondering about the changes that was introduced for testing on my computer.

forki:
thx!

johannesvorster:
Good day Steffen


Thanks for accepting my proposal.


If you have more bugs that need fixing, please feel free to point me to them.


I like working with your code and I start to get the hang of it.


Regards


Johannes Vorster

T: (+27) 079 490 8706
F: (+27) 088 012 661 6672
E: johannes.vorster@outlook.com<mailto:elconi.johannes@gmail.com>
<http://za.linkedin.com/in/johannesv>


forki:
Cool and welcome. 

Basically all open issues are available for all contributors. Some are really challenging and others are much easier. It's hard to tell sometimes without trying ;-)

PR #2113, opened 2017-01-19
Generate load scripts on install abidding to new paket.dependencies option

smoothdeveloper:
new option `generate_load_scripts: on|off|true|false` in `paket.dependencies`

this option, if on/true, calls the loading script generation

Note that this does it in very eager way (if defined for any group, it will generate all), it also doesn't allow to specify framework or language like from command line, keeping it simple for now.

The option is off by default, because there is performance penality and it generates lots of small script files, there is also no proper logic for "garbage collection" process.

Implementation notes:
* A new convenience method is also exposed on PublicAPI: Dependencies.GenerateIncludeScripts()
* I had to migrate a good deal of stuff in a module I named Queries, mostly stuff which were raw in Program.fs
* added a unit test for parser, testing all cases of the new option
* added an integration checking calling install with option on results in script being generated without further invocation
* Since both .dependencies and .lock files are used often in conjunction, I made a PaketFiles DU (in Queries module)
* Since both group name and package name are used often in conjunction, I made a QualifiedPackageName single case DU, which helped me remove a bit of cruft related to default group


smoothdeveloper:
@forki (or other maintainers) I think few people were looking to have this available, could you have a quick look, I think the tests are conclusive enough for this to get merged unless you have code review comments?

src/Paket.Core/PublicAPI.fs
```diff
@@ -192,6 +192,9 @@ type Dependencies(dependenciesFileName: string) =
             semVerUpdateMode = defaultArg semVerUpdateMode SemVerUpdateMode.NoRestriction,
             touchAffectedRefs = defaultArg touchAffectedRefs false)
 
+    static member GenerateIncludeScripts() =
```
smoothdeveloper: Going to rename that one, harmonizing on "load scripts" which better fits F# terminology

smoothdeveloper:
@forki this is finally passing CI

forki:
very cool. /cc @dsyme 

@smoothdeveloper can we have some docs as well?

smoothdeveloper:
@forki good point, I'll make a doc PR soon.

PR #2115, opened 2017-01-19
Fix preview4 install

enricosada:
use 1.0.0-preview4-004233 that's the latest released preview4
see [.net core download archive](https://github.com/dotnet/core/blob/master/release-notes/download-archive.md) for all released versions

also changed to use the azureedge mirror because is faster.

that's the url used by official install script https://github.com/dotnet/cli/blob/rel/1.0.0-preview4/scripts/obtain/dotnet-install.ps1 (or .sh)

/cc @forki


forki:
Thx. Kiss

PR #2120, opened 2017-01-24
add osx-specific changes to allow for easier building

baronfel:
Simple changes to the build script to grab the osx dotnet sdk and extract it.


baronfel:
Build failure here is due to a path too long exception, not the build script.

forki:
Thx

PR #2121, opened 2017-01-25
Fix CI test breakage

baronfel:
This renames an integration test and the associated scenario folder to come in under the path length limit that we were hitting on Appveyor.


forki:
Thx

PR #2122, opened 2017-01-25
Added docs for `generate_load_scripts` option in `paket.dependecies` file

battermann:


smoothdeveloper:
@forki looks good, @battermann thanks a lot!

forki:
https://fsprojects.github.io/Paket/paket-generate-include-scripts.html <- can we update this as well?

PR #2125, opened 2017-01-26
[WIP] finalize paket integration for new .net sdk (msbuild 15)

enricosada:
WIP, but

basically developing the final example in `dev`, so:

- `dev/.paket` contains the boostrapper
- `dev/console` contains a sample console app

NOTES

The `Paket.Sdk` is just for my easy of development, final is inside `Paket` nupkg

TODO

- [X] boostrapping with zero binaries
- [X] target to cleanup bootstrapped paket.exe
- [X] boostrap on first `dotnet restore`
- [X] `paket restore` called on `dotnet restore` after boostrapping
- [X] xplat (paket is an FDD = `dotnet publish`)
- [ ] configurable paket version to restore on boostrap, with sane default
- [ ] .cmd/.sh wrapper for boostrapping?
- [X] should work from `dotnet msbuild` so run .net core (`dotnet build` = `dotnet msbuild /t:Build`)
- [ ] should work from `msbuild` so .net full (so call `paket.exe`)
- [ ] use `Paket` nupkg instead of `Packet.Sdk`
- [ ] integrate in current build script

@forki not lot of work todo left 😄


enricosada:
I am trying different ways to integrate at restore step, each has a prop/cons ( I written something in dev/docs ).

I want to maintain the current nuget targets logic as much as possible, because make it easier for other commands (build/publish, the pack is not much useful for paket I think)

But override is hard because nuget target is imported AFTER our target file, becuase is inside the Sdk attribute. 
And nuget target extensibility is limited (but we can completely override the nuget tasks)

Monday i’l try using Sdk attribute for paket too, so we can override single targets.

forki:
Hi what's the status here. Next week I'd like to go rc or even release paket 4.

enricosada:
status is:

- bad for deeper sdk integration, i cannot find yet the correct flow/targets
- okish for bootstrapper

maybe i should split in two..

enricosada:
closing because obsolete

PR #2127, opened 2017-01-28
Support GitHub dependencies with spaces

inosik:
This should enable support for GitHub dependencies which contain spaces in their file name.


forki:
this doesn't compile!?

inosik:
I always forget that `String.Contains` doesn't accept `char`s 😅

inosik:
Okay, I will look at the failing tests.

PR #2128, opened 2017-01-29
Ensure that boostrapper --help always work

### LINKED ISSUE #2099: Bootstrapper doesn't print its help page
### Description

`paket.bootstrapper.exe --help` doesn't print anything.

### Repro steps

```
$ .paket/paket.bootstrapper.exe --help
```

### Expected behavior

The bootstrapper prints the help page.

### Actual behavior

The bootstrapper doesn't print anything, because the condition in [ConsoleImpl.cs#L54](https://github.com/fsprojects/Paket/blob/89b11627d73d322dcad9d042c3b1314652caef6f/src/Paket.Bootstrapper/ConsoleImpl.cs#L54) evaluates to `true`.

### Known workarounds

None. `paket.bootstrapper.exe -v --help` doesn't work either, because the verbosity gets set later.

vbfox:
It should display something even when the boostrapper is in silent or very
silent mode.

Fixes #2099

### DIFF
--- a/src/Paket.Bootstrapper/ConsoleImpl.cs
+++ b/src/Paket.Bootstrapper/ConsoleImpl.cs
@@ -49,6 +49,16 @@ namespace Paket.Bootstrapper
             WriteConsole(message, ConsoleColor.DarkGray, Verbosity.Trace);
         }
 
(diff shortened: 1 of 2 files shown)


forki:
thanks!

Issue #2109, opened 2017-01-17
BenchmarkDotNet doesn't seem to install all necessary dependencies


smoothdeveloper:
### Description

paket.dependencies:

```
source https://www.nuget.org/api/v2
strategy: max
redirects: on
framework: >= net40

nuget BenchmarkDotNet = 0.10.1
```

in the nuspec of BenchmarkDotNet.Core (which is transitive dependency and installed), it seems to mention `System.Threading.Tasks.Extensions`
```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
    <metadata>
        <id>BenchmarkDotNet.Core</id>
        <version>0.10.1</version>
        <title>BenchmarkDotNet.Core</title>
        <authors>.NET Foundation and contributors</authors>
        <owners>.NET Foundation and contributors</owners>
        <licenseUrl>https://github.com/dotnet/BenchmarkDotNet/blob/master/LICENSE.md</licenseUrl>
        <projectUrl>https://github.com/dotnet/BenchmarkDotNet</projectUrl>
        <iconUrl>https://raw.githubusercontent.com/dotnet/BenchmarkDotNet/master/docs/guide/logo.png</iconUrl>
        <requireLicenseAcceptance>false</requireLicenseAcceptance>
        <description>Powerful .NET library for benchmarking</description>
        <copyright>.NET Foundation and contributors</copyright>
        <language>en-US</language>
        <tags>benchmark benchmarking performance</tags>
        <dependencies>
            <group targetFramework=".NETCoreApp1.1">
                <dependency id="Microsoft.NETCore.App" version="1.1.0" />
                <dependency id="System.Runtime.Serialization.Primitives" version="4.1.1" />
                <dependency id="Microsoft.DotNet.InternalAbstractions" version="1.0.0" />
                <dependency id="System.Xml.XPath.XmlDocument" version="4.0.1" />
            </group>
            <group targetFramework=".NETFramework4.5">
                <dependency id="System.Threading.Tasks.Extensions" version="4.0.0" />
            </group>
        </dependencies>
        <frameworkAssemblies>
            <frameworkAssembly assemblyName="System.Management" targetFramework=".NETFramework4.5" />
            <frameworkAssembly assemblyName="System.Xml" targetFramework=".NETFramework4.5" />
        </frameworkAssemblies>
    </metadata>
</package>
```

[...]

PR #2129, opened 2017-01-29
Add "netcoreapp1.1"

vbfox:
Add .Net Core 1.1 in the list of frameworks, some packages like [BenchmarkDotNet](https://www.nuget.org/packages/BenchmarkDotNet/) have conditional dependencies on it and can't be correctly used without paket knowing about it.

Related to #2109


forki:
thanks!

PR #2131, opened 2017-01-30
Report broken project dependencies properly

vivainio:
Currently, if there are broken project dependencies, Paket just says "Unable to parse: something.csproj". It doesn't specify what project has the broken reference, or that there are broken project dependencies in the first place.

This new warning specifies what csproj you need to fix, and that we are indeed dealing with broken project dependencien.


forki:
thanks!

Issue #1943, opened 2016-10-06
Suggestions for load script generation


dsyme:
@sylvanc and I have been using the amazing package-load-script-generation feature.  It's fantastic

Here are some suggestions for how it can be adjusted 
- Always generate the package load scripts by default.  They are just too useful to not do it
- Always generate by assuming the latest .NET Framework by default (net462) unless told otherwise.
- Always generate "paket-refs.fsx" (or "paket-refs.csx") next to `paket.dependencies` that references all the package loads.  
- If there is a non-main group of references then generate `paket-build-refs.fsx`
- Never generate a `#r` on FSharp.Core.  I think @tpetricek is adding a separate issue on this.

This would make the feature really, really usable.  Right now putting things deep in `paket-files` makes things hard.

thanks
don

cc @dungpa

matthid:
I agree with most, actually I proposed https://github.com/fsprojects/Paket/issues/1681 and https://github.com/fsprojects/Paket/issues/1680.

Defaulting to latest framework sounds like a sane default, however we might not have all dependencies available when framework restrictions are in place. Therefore I'd rather default to the framework restriction we have in place (`paket.dependencies`) and fall back to latest full framework if there is none.

dsyme:
> Therefore I'd rather default to the framework restriction we have in place (paket.dependencies) and fall back to latest full framework if there is none.

Yes, that

dsyme:
Yes, #1681 and #1680 cover most of this.  +2 for those :)

smoothdeveloper:
@dsyme excluding `FSharp.Core` should be easy, could you share the contents of your `paket.dependencies` where you hit that issue and I'll give it a look (I never noticed it occurring but maybe I used nuget packages which don't rely on that package).

https://github.com/fsprojects/Paket/blob/f5f509ace1cc58ce1f7e487c2a9f13226e5cd395/src/Paket.Core/ScriptGeneration.fs#L98

Regarding #1681, I'm not sure it can be done by default, the implementation is kind of costly, on huge `paket.dependencies` file I have here, it takes almost 30 sec to generate all the current scripts, I'm sure it could be made more efficient with few tricks but I don't see the cost coming close to 0, we have to analyze assembly metadata to figure out correct order of stuff and I don't think there is a shortcut for that. 

https://github.com/fsprojects/Paket/blob/f5f509ace1cc58ce1f7e487c2a9f13226e5cd395/src/Paket.Core/ScriptGeneration.fs#L46

[...]

PR #2137, opened 2017-02-08
For review: breaking changes to generate-load-scripts

smoothdeveloper:
* handle generate-load-script command
* change the output folder and script names for generate-load-scripts

addresses #1943

### DIFF
--- a/docs/content/commands/generate-load-scripts.md
+++ b/docs/content/commands/generate-load-scripts.md
@@ -1,14 +1,14 @@
-## Generating include scripts for all NuGet packages
+## Generating load scripts for all NuGet packages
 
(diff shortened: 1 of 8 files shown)


forki:
can we keep the old command as alias?

smoothdeveloper:
@forki yes I've done that actually

https://github.com/smoothdeveloper/Paket/blob/06f454440f9f00f8942fdbae2ed219d7f124e92e/src/Paket/Program.fs#L365

it still execute the command but prints `please use generate-load-scripts`.

Although it won't create the scripts in the previous location, I don't think it warrants generating those at the old location but open for suggestion.

PR #2142, opened 2017-02-16
Fix cmd argument quoting in Paket.Restore.targets

jdpage:
The line that runs `paket restore` in Paket.Restore.targets needs to wrap the path to the .csproj file in quotes if its path has spaces.


forki:
mhm looks this breaks the build

forki:
Can you please take a look again?

forki:
this was merged. thx

PR #2150, opened 2017-02-23
Lower case group folder name

JonCanning:
generate-include-scripts fails on Linux systems when the group name is upper case since the folder is created lower case

This causes the dir.Exists check here:
https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NuGetV2.fs#L609
to fail

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -384,7 +384,7 @@ type Dependencies(dependenciesFileName: string) =
             | None -> failwithf "Package %O is not installed in group %O." packageName groupName
             | Some resolvedPackage ->
                 let packageName = resolvedPackage.Name
-                let groupFolder = if groupName = Constants.MainDependencyGroup then "" else "/" + groupName.ToString()
+                let groupFolder = if groupName = Constants.MainDependencyGroup then "" else "/" + groupName.ToString().ToLower()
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
@JonCanning thanks for the report, could you give a paket.dependencies that you are looking at and a summary of folder names where they mismatch (wether in file structure or in the generated script clauses)?

Thanks!

smoothdeveloper:
Mhh this is a PR already :smile: yes lowercasing the group name seems correct, we probably need same fix in core3 branch: https://github.com/fsprojects/Paket/blob/39468c9404932e20c3b9a8189d3b8b7f03e5b6f5/src/Paket.Core/PublicAPI.fs#L425

If you can consider adding a test that would be nice.

src/Paket.Core/PublicAPI.fs
```diff
@@ -384,7 +384,7 @@ type Dependencies(dependenciesFileName: string) =
             | None -> failwithf "Package %O is not installed in group %O." packageName groupName
             | Some resolvedPackage ->
                 let packageName = resolvedPackage.Name
-                let groupFolder = if groupName = Constants.MainDependencyGroup then "" else "/" + groupName.ToString()
+                let groupFolder = if groupName = Constants.MainDependencyGroup then "" else "/" + groupName.ToString().ToLower()
```
forki: the groupd should have a compare name - I will use that one.
ANd thanks for fixing!

PR #2153, opened 2017-02-25
remove print statements in generated load scripts

smoothdeveloper:
Letting CI check we don't assert anything on those print statements.


forki:
thanks

PR #2155, opened 2017-02-25
Link to documentation on how to resolve a conflict

0x53A:
When I got this error (due to a broken package), it took me way too long to find the correct documentation.

I'm a big fan of error messages telling you what to do.

(I didn't mention the == operator directly in the error message on purpose, because fixing the version can cause a lot of fun ( MethodNotFoundException <3 ), and there should be at least a chance to warn the user.

What would you think about adding a big bold warning to http://fsprojects.github.io/Paket/nuget-dependencies.html#Use-exactly-this-version-constraint how it can cause all kinds of trouble? (Here be dragons ...)


forki:
💋

PR #2171, opened 2017-03-04
Polish and Cleanup

cloudRoutine:
- upgraded packages to stable versions
- fixed msbuild15 search error for F# targets
- cleaned up project files & paket.references
- removed unused references


build.fsx
```diff
@@ -61,7 +62,7 @@ let gitName = "Paket"
 // The url for the raw files hosted
 let gitRaw = environVarOrDefault "gitRaw" "https://raw.github.com/fsprojects"
 
-let dotnetcliVersion = "1.0.0-preview4-004233"
+let dotnetcliVersion = "2.0.0-alpha-005165"
```
forki: Huh? @enricosada is this good?

paket.dependencies
```diff
@@ -2,20 +2,20 @@ source https://nuget.org/api/v2
 
 nuget Newtonsoft.Json redirects: force
 nuget Argu prerelease
-nuget FSharp.Core prerelease redirects: force
+nuget FSharp.Core == 4.0.1.7-alpha redirects: force
```
forki: Why this?

build.fsx
cloudRoutine: it passed all the tests didn't it? 😉

paket.dependencies
cloudRoutine: This is the same as for the test library, which ends up referencing the projects using FSharp.Core from outside of the test group so it has to be the same version in both of them, it just says it up top now instead of down below.

`4.0.0.1` doesn't support netcore so it's either this or we add a new group for netcore switch this to `4.0.0.1` and use `4.1.0` with netcore. 

It's just a superficial change.

build.fsx
cloudRoutine: plus I used the latest so I could be sure the slimmed down fsproj and csproj would be compatible

PR #2184, opened 2017-03-15
proposal for dotnet paket

ctaggart:
This is not to be merged. I'm proposing that a `donet-paket` be created to make distributing paket extremely easy. Distributing build tools like paket and FAKE via DotNetCliToolReference's is a great way to do things for MSBuild 15. I shipped SourceLink v2 last week as one. Pretend in the code example that it was `paket.ps1` instead. Perhaps the directory needs to be passed in too. Those details can be worked out.

### DIFF
--- a/buildtools/build.proj
+++ b/buildtools/build.proj
@@ -1,0 +1,5 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <ItemGroup>
+    <DotNetCliToolReference Include="dotnet-sourcelink" Version="2.0.2" />
+  </ItemGroup>
+</Project>
--- a/sourcelink.ps1
+++ b/sourcelink.ps1
@@ -1,0 +1,7 @@
+Push-Location buildtools
+if (-not (Test-Path 'obj/project.assets.json')) 
+{
+    dotnet restore
+}
+dotnet sourcelink @args
+Pop-Location


forki:
I think @enricosada is already working on this. Take a look into `./tools`

enricosada:
@ctaggart i love dotnetcli tool, i have some atm [dotnet-mergenupkg](https://github.com/enricosada/dotnet-mergenupkg), [dotnet-proj-info](https://github.com/enricosada/dotnet-proj-info).

but there are annoying drawback like:

- cannot have multiple *proj in same dir https://github.com/dotnet/cli/issues/4808 
- require `dotnet-` prefix https://github.com/dotnet/cli/issues/4722 so need to rename fake.exe
- working directory of `dotnet paket` should be the same directory where is defined the `.*proj` who contains the `DotNetCliTool`.

but are good. Is possibile to use a `tools` directory with that (or .paket).
i use that atm for repo specific tools, until a real per repo install exists (sdk 2.0?)

For paket i was doing some work/exploration in https://github.com/fsprojects/Paket/pull/2125
 about boostrapper and deeper sdk integration.

For paket boostrapper i think a cli tool is not the best solution.

One constraint i know exists is paket should run on net40 too, without netcore sdk installed. just from vs (who embeed the sdk, but not the .net core tools).
Another is we need it at restore step, so need anyway to restore it somehow.

Best solution ihmo atm is to package fake as nupkg with multiple tools for framework (net45, netcoreapp1.* fdd).
At restore step, that package is downloaded with restore (so normal `<PackageReference Include="Paket" Version="4.*" />`, so easy to version) into `.paket/out` directory.
The paket target file will execute `dotnet .paket/out/build/netstandard1.0/paket.dll` or `./paket/out/build/net45/paket.exe` (with mono prefix on mono), based on msbuild host runtime.
That can be specified in a .props file who set `$(PaketCommand)` property, and is imported later for next runs.

I have some questions left, like:

- how to run it from command line (what command wrap in the `.\paket\paket` script? `dotnet paket` or `paket.exe`/`mono paket.exe`?), maybe save the choose at restore step. if was restored in vs -> net45, otherwise netcore

enricosada:
feedback welcome! not about my PR because is a bit stale, but about how to boostrap

forki:
@enricosada is working on this in new PR.

PR #2186, opened 2017-03-15
check if file obj/{proj}.references is empty, not if exists

enricosada:
if file doesnt exists, `ReadLinesFromFile` doesnt fail, just return empty items list as Lines


forki:
Will this cover the case where the file doesn't exist at all?

enricosada:
@forki yes, as i written if file doesnt exists, ReadLinesFromFile doesnt fail, just return empty items list as Lines.

the change skip adding projectreferences if list is empty, so is ok.

tested locally chaning the path in target file to read a non existing file

Issue #1994, opened 2016-10-31
Bootstrapper doesen't handle corrupt paket.exe 


MiguelAlho:
### Description

If for some reason paket.exe is not completely downloaded during the bootstrap process (for instance , power outage), the file can be corrupted. We observed on a build server that all subsequent calls to the bootstrapper will pull an invalid paket.exe from the cache breaking builds (as in avoiding them from continuing). Cleaning paket's cache is required to continue.

### Repro steps

Not sure, but creating and invalid paket.exe and placing it in cache might be enough to exemplify this. Might be able to automate it in a test.

### Expected behavior

Bootstrapper should verify that the downloaded paket .exe (wether from the web or cache) is valid, possibly by comparing ti with some checksum. If the paket.exe that is present is determined to be invalid, redownload it and overwrite existing versions.

### Actual behavior

A broken paket.exe is put in the .paket directory.

### Known workarounds

Manually cleaning the cache will allow for a new download. This step may be limiting due to user access rights to build servers.

TeaDrivenDev:
Fixed with #2192?

JohnyWS:
@forki and @enricosada I believe this has been fixed. The only place it is missing is with the NuGet download strategy, and I'm assuming that has built-in hashsum checks natively?

PR #2192, opened 2017-03-20
Handling corrupt paket.exe downloads

simonhdickson:
Potential fix for #1994, don't think its ready for pulling yet but thought I'd put it up for feedback.


forki:
That looks interesting!

simonhdickson:
Is this error related to anything I've done?

simonhdickson:
Also does it make sense to validate the NuGet package version? Seems like it should be unnecessary.

PR #2209, opened 2017-03-30
Docs: Change command palette name for VS Code editor

RamanBut-Husaim:
Change command palette name for VS Code editor as confuse readers a little.


RamanBut-Husaim:
@forki hello.

Could you please take a look?

forki:
Thx

PR #2211, opened 2017-03-30
Create `.nuspec` files from project file

cloudRoutine:
create a nuspec from a project file, usable from commandline, can be integrated into paket restore


cloudRoutine:
@forki could use some review

forki:
What does it change in the workflow?

cloudRoutine:
right now nuget is creating the nuspec that's getting trimmed down before paket packs the output into a nupkg, so this will let us remove nuget from another point in the build pipeline.

PR #2214, opened 2017-03-31
Fix #2213 - Workaround for TFS dependency resolution

### LINKED ISSUE #2213: Dependency resolution does not work for on-premise TFS private feed
I encountered with the same issue as described in #1798 but for TFS. TFS has the same bug with package dependencies as VSTS (I create issue for this [here](https://developercommunity.visualstudio.com/content/problem/38370/package-management-server-does-not-return-package.html)).

I found this code to workaround for VSTS:

https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NuGetV2.fs#L682

and this:

https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/NuGetV2.fs#L296

There are IF conditions for visualstudio.com.

But I have URLs something like this:

"http://vc-tfs-server:8080/tfs/DefaultCollection/_packaging/default/nuget/v3/index.json" - in first case (L682) and:

"http://vc-tfs-server:8080/tfs/DefaultCollection/_packaging/default/nuget/v2" - in second case (L296)

Can you fix it for me? Or I can fix this for both cases and send you pull request. What do you think about it?

drwatson1:
I tested this solution for TFS, VSTS and NuGet.org - all work fine now

### DIFF
--- a/src/Paket.Core/NuGetV2.fs
+++ b/src/Paket.Core/NuGetV2.fs
@@ -265,6 +265,9 @@ let getDetailsFromNuGetViaODataFast auth nugetURL (packageName:PackageName) (ver
             return parseODataDetails(url,nugetURL,packageName,version,raw)
     }
 
+let urlSimilarToTfsOrVsts url = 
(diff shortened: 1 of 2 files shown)


forki:
cool!

Issue #1645, opened 2016-04-26
Referencing portable library


JonCanning:
### Description

Using Auth0 libraries which only has portable targets

[nuspec](https://github.com/auth0/auth0.net/blob/master/nuget/Auth0.AuthenticationApi.nuspec)

Reference is not added to project
### Repro steps

paket add nuget Auth0.AuthenticationApi project MyProject
### Expected behavior

Reference is added
### Actual behavior

Warning shown in console

`Package Auth0.AuthenticationApi contains libraries, but not for the selected TargetFramework net452 in project`

Reference is not added
### Known workarounds

Rename folders in lib to `net45`

inosik:
The `portable45-net45+win8+wpa81` directory doesn't look right to me. As of [NuGet docs](https://docs.nuget.org/create/targetframeworks), there is no `portable45`, only `portable`.

Their project file has the `Profile111` `TargetFrameworkProfile`, which should translate to `portable-net45+win8+wpa81`.

NuGet can't install this package either, BTW.

forki:
Yeah I guess that's the issue.

JonCanning:
I'll raise it with the Auth0 devs

inosik:
Paket still translates the incorrect moniker into this:
```xml
<Choose>
  <When Condition="($(TargetFrameworkIdentifier) == 'WindowsPhoneApp') Or ($(TargetFrameworkIdentifier) == '.NETCore') Or ($(TargetFrameworkProfile) == 'Profile32')">
    <ItemGroup>
      <Reference Include="Auth0.AuthenticationApi">
        <HintPath>..\packages\Auth0.AuthenticationApi\lib\portable45-net45+win8+wpa81\Auth0.AuthenticationApi.dll</HintPath>
        <Private>True</Private>
        <Paket>True</Paket>
      </Reference>
    </ItemGroup>
  </When>
</Choose>
```
Which is wrong, obviously. Should we do something about that?

forki:
I guess it will just go away when the package is fixed ;-)

enricosada:
Closing, fixed in upstream package

PR #2215, opened 2017-03-31
Fix to handle weird malformed portable-only libraries.

OniBait:
Fix for issue #1645 

The nuget parser handles strings in the form of:
{Identifier}{Version}-{Profile}

(see `NuGet.VersionUtility.ParseFrameworkName`)

So the string: `portable-net45+win8+wp8` gets parsed to:
`.NETPortable,Version=v0.0,Profile=net45+win8+wp8`

The string: `portable45-net45+win8+wp8` gets parsed to:
`.NETPortable,Version=v4.5,Profile=net45+win8+wp8`

Both end up being "valid", but for portable frameworks, the version doesn't matter (the profile does).

The fix is to use a regex to strip out "portable-" and "portable{version}-" when doing platform matching.

(note: `NuGet pack` seems to be using the `portable45` for whatever reason now)

### DIFF
--- a/src/Paket.Core/PlatformMatching.fs
+++ b/src/Paket.Core/PlatformMatching.fs
@@ -7,7 +7,7 @@ let MaxPenalty = 1000000
 
 let inline split (path : string) = 
     path.Split('+')
-    |> Array.map (fun s -> s.Replace("portable-", ""))
+    |> Array.map (fun s -> System.Text.RegularExpressions.Regex.Replace(s, "portable\\d*-",""))
     
 let extractPlatforms = memoize (fun path  -> split path |> Array.choose FrameworkDetection.Extract |> Array.toList)
 


inosik:
> (note: NuGet pack seems to be using the portable45 for whatever reason now)

Yes, I have seen this on a couple of other packages now. BTW, this behavior is still not documented.

OniBait:
Yeah. Closest I've found to any sort of documentation on it is in the old v2 code here:
https://github.com/NuGet/NuGet2/blob/2.13/src/Core/Utility/VersionUtility.cs#L241

Although, I guess one could argue that _technically_ it is documented here:
https://docs.microsoft.com/en-us/nuget/create-packages/supporting-multiple-target-frameworks#framework-version-folder-structure

with the caveat that for portable frameworks, it should have the profile after:
`{framework name}[{version}]-{profile}`

In any case, the following code in the NuGet library works:
`NuGetFramework.Parse("portable45-net45+win8+wp8")`
returns:
`[.NETPortable,Version=v4.5,Profile=Profile78]`

OniBait:
Alternatively, does it make sense to reference the NuGet DLLs and use their methods for parsing frameworks? I'm not well versed enough in the project to know if there was a conscious decision to not have NuGet as a dependency...

forki:
That's a big nono. It's failing for million other reasons.

Am 03.04.2017 19:03 schrieb "onibait" <notifications@github.com>:

Alternatively, does it make sense to reference the NuGet DLLs and use their
methods for parsing frameworks? I'm not well versed enough in the project
to know if there was a conscious decision to not have NuGet as a
dependency...

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/fsprojects/Paket/pull/2215#issuecomment-291206731>, or mute
the thread

Issue #2119, opened 2017-01-21
Make upstream the default remote for Git dependencies


rintje:
### Description

I'm using git dependencies to checkout several related repositories in addition to the main repo. They all come from a central bitbucket server at example url https://mybitbucket.nl/bitbucket/scm/sp/myproject.git  I would like to be able to push changes back to that server.

When paket clones a repo, it sets the URL for the origin remote to the central user paket folder, something like 'C:\Users\myusername\\.paket\git\db\myrepo' instead of the upstream URL. This makes it harder than necessary to push changes to upstream.  I need to select a remote at every push, or need to git config remote each git dependency manually.  

This seems unnecessary complex to me. How about making upstream the default origin for git dependencies?
 
### Repro steps
1. Step A
Add a git dependency, then paket update/install to clone and checkout the repo.
2. Step B
Make changes to the repo working dir
3. Step C
Commit and push the changes. Git attempts to push the changes to the repo in the central user paket folder, instead of the repo at the server URL specified in paket.dependencies. 

### Expected behavior
1. Step A
Add a git dependency, then paket update/install to clone and checkout the repo.
2. Step B
Make changes to the repo working dir
3. Step C
Commit and push the changes. Changes are pushed to the server URL (upstream) specified in paket.dependencies.

### Known workarounds
Select the proper remote on each push, or change git config manually after paket restore/update.

_Edit: typos, textual_

rintje:
Hmm, just changing upstream does not cut it either. because the tracking branch is not set to the branch specified in paket.dependencies.

I guess the actual question should be: is pushing changes to git dependencies a use case Paket supports (or should support)? For now the answer seems no.

PR #2226, opened 2017-04-11
[WIP] Make pushing changes from Git dependency repositories easier

tomasdeml:
This PR mostly addresses issues raised in #2119 and changes the way Git dependencies are initialized to make pushing changes from the cloned repository easier. The following workflow will be now possible/easier with Paket:
1. Add a git dependency to `paket.dependencies`, then run `paket update/install` to clone and checkout the repo.
1. Checkout `master` or create a feature branch in the cloned (_downstream_) repo, i.e. the repo under `paket-files` directory.
1. Make changes in the cloned repo, run tests and commit the changes.
1. Push the changes from the cloned repo to the `origin` (_upstream_) repo, i.e. the repo specified in `paket.dependencies`.
1. Run `paket update` to update `paket.lock` with SHA1 of the changes pushed to the upstream repo.

I understand this PR may be controversial as it develops Paket in a way that may be currently served by Git submodules. However, Paket has some advantages over submodules:
- Paket supports semver tags when resolving Git dependencies
- Paket caches the upstream repository locally which helps build servers and saves space/time when working with multiple projects depending on same Git dependency

The changes helped our company to move away from (very painful) Nuget source packages. Feel free to suggest changes or reject the PR entirely :-). However, if you choose to accept the proposed changes, **a migration strategy has to be devised before merging the PR. Otherwise users would be required to clear Paket local temp directory and respective `paket-files` directories as this is a breaking change.**
  
### The following changes were made to the implementation of Git dependencies:

- Cache repository (i.e. the repo under `%USERPROFILE%\.paket\git\db`) is cloned as a bare mirror

  Bare clone saves disk space as it does not need a working directory. Mirror clone tracks the upstream remote more closely as it always recreates refs.

- Downstream repository is configured with the `origin` remote pointing to the upstream repository (instead of the cache repository)

  Having the upstream repo configured as the `origin` makes pushing changes to the upstream more natural. We also want to prevent user from pushing the changes to the cache directly - the cache should be managed by Paket.

- Tracking branch for each downstream repo (`b<REPO_PATH_HASH>`) is no longer created in the cache repo

  The branch IMO serves no practical purpose now, see next point.

- `paket/lock` tag is created and checked out in the downstream repository when `update`ing/`restore`ing the dependency

  This allows user to see what the original checked out commit was when making changes. By checking out the tag - instead of resetting the downstream repo `--hard` to the tracking branch - the repo starts in the detached HEAD state which:
    * allows easy experimentation - user can discard her commits simply by checking out the `paket/lock` tag again
    * forces user to checkout a branch when doing meaningful changes.

- Downstream repository is checked for uncommitted changes before `update`ing/`restore`ing the dependency

  This prevents hard-to-track bugs as well as non-repeatable builds that might occur by merging user's uncommitted changes into the new working directory.

- Downstream repository is checked for commits made in the detached HEAD state before `update`ing/`restore`ing the dependency

  This makes working in the downstream repository safer as it saves user from having to inspect reflog after "losing" her commits. It also forces the user to move the changes to a permanent branch when making meaningful changes.

No tests were impacted by the changes. However, there are currently few integration tests covering Git dependencies.

### DIFF
--- a/src/Paket.Core/GitHandling.fs
+++ b/src/Paket.Core/GitHandling.fs
@@ -171,16 +171,15 @@ let checkoutToPaketFolder repoFolder cloneUrl cacheCloneUrl commit =
     try
         // checkout to local folder
         if Directory.Exists repoFolder then
-            CommandHelper.runSimpleGitCommand repoFolder ("remote set-url origin " + quote cacheCloneUrl) |> ignore
-            verbosefn "Fetching %s to %s" cacheCloneUrl repoFolder 
-            CommandHelper.runSimpleGitCommand repoFolder "fetch origin -f" |> ignore
+            verbosefn "Fetching %s to %s" cacheCloneUrl repoFolder
+            CommandHelper.runSimpleGitCommand repoFolder (sprintf "fetch --tags --prune %s +refs/heads/*:refs/remotes/origin/*" <| quote cacheCloneUrl) |> ignore
         else
             let destination = DirectoryInfo(repoFolder).Parent.FullName
             if not <| Directory.Exists destination then
                 Directory.CreateDirectory destination |> ignore
             verbosefn "Cloning %s to %s" cacheCloneUrl repoFolder
-            CommandHelper.runSimpleGitCommand destination ("clone " + quote cacheCloneUrl) |> ignore
-            CommandHelper.runSimpleGitCommand repoFolder ("remote add upstream " + quote cloneUrl) |> ignore
+            CommandHelper.runSimpleGitCommand destination (sprintf "clone %s %s" (quote cacheCloneUrl) (quote repoFolder)) |> ignore
+            CommandHelper.runSimpleGitCommand repoFolder (sprintf "remote set-url origin %s" <| quote cloneUrl) |> ignore
 
         tracefn "Setting %s to %s" repoFolder commit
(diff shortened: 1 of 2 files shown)


forki:
is this ready?

tomasdeml:
Oh, I did not expect the PR to be merged so easily 😄. Just to be sure - so the missing migration strategy is not an issue?

forki:
Can you please send a pull request to the release notes that makes this manual step clear?

tomasdeml:
Sure, hang on...

tomasdeml:
Just wanted to share that we have stumbled upon an issue when using Git LFS. Git LFS [cannot fetch files using the file protocol](https://github.com/git-lfs/git-lfs/issues/724). As this protocol is used when fetching from local Paket cache repository, a [workaround](https://github.com/git-lfs/git-lfs/issues/724#issuecomment-146906712) must be applied. Before invoking `paket install/restore/update`, you have to:
1. Disable Git LFS smudge filter via an env var
2. Redirect Git LFS endpoint in local config of the downstream repository to the remote endpoint of the original upstream repository.

Once the `install/restore/update` command completes, you have to explicitly pull large files from the new LFS endpoint (`git lfs pull`).

PR #2229, opened 2017-04-11
[WIP] bootstrapper: Support NugetSource app-setting key

gimpf:
Options like PreferNuget, ForceNuget and NugetSource often go together (for instance to force a specific internal NuGet server for CI). Added one that I wanted.

Tests are still missing, but is this otherwise OK?


forki:
is this ready?

forki:
thanks!

PR #2230, opened 2017-04-11
[WIP] Fix rotating app.config entries when generating redirects

gimpf:
An older commit termed "cleanup" seems to have inadvertently broken app.config binding redirects; they work, but on every paket install the order of the redirects is rotated, that is, the first one is moved to the last one (or the other way round).

This makes diffs more verbose than necessary, so I patched that out.

Tests still missing, but is it otherwise OK?


forki:
can you please add a comment that making it strict is important!?

gimpf:
of course ;-)

PR #2232, opened 2017-04-11
paket.targets: add DownloadPaket target

gimpf:
Those targets were here once, but were removed in a commit mentioning mono support, although without further explanation.

While not strictly _necessary_ per se, the targets file is still on the release pages, and the bootstrapping was a nice example.

In case I'm not missing anything important, can we add that back in?


vbfox:
That PR would completely break magic mode :(

forki:
@gimpf did you read about magic mode? https://fsprojects.github.io/Paket/bootstrapper.html#Magic-mode 

I think that's the reason why we removed all that.

gimpf:
@forki No, didn't know that. Makes this PR totally obsolete.

PR #2236, opened 2017-04-11
Only include dlls in analyzers

### LINKED ISSUE #2219: Running paket install installs .pdb files
### Description

When running `paket install` with .pdb files in `packages\analyzers\SomeAnalyzer\analyzers\dotnet\cs\SomeAnalyzer.pdb` the pdb file is added as references to the projects.

Also if I remember correctly analyzers are uninstalled from the project if the folder is empty. Maybe this is the intended behavior.

### Repro steps

1. Create a C# project and add [an analyzer nuget](https://www.nuget.org/packages/Gu.Analyzers/)
2. Add a pdb file to `packages\analyzers\SomeAnalyzer\analyzers\dotnet\cs\SomeAnalyzer.pdb`
3. Run paket install

[ConsoleApp1.zip](https://github.com/fsprojects/Paket/files/891139/ConsoleApp1.zip)

### Expected behavior

I don't expect `paket install` to install anything that is not in the .nupkg

vbfox:
Fixes #2219

### DIFF
--- a/src/Paket.Core/NuGetV2.fs
+++ b/src/Paket.Core/NuGetV2.fs
@@ -605,26 +605,28 @@ let DownloadLicense(root,force,packageName:PackageName,version:SemVerInfo,licens
                     traceWarnfn "Could not download license for %O %O from %s.%s    %s" packageName version licenseUrl Environment.NewLine exn.Message
     }
 
-
-let private getFiles targetFolder subFolderName filesDescriptionForVerbose =
+let private getFilesMatching targetFolder searchPattern subFolderName filesDescriptionForVerbose =
     let files =
         let dir = DirectoryInfo(targetFolder)
         let path = Path.Combine(dir.FullName.ToLower(), subFolderName)
         if dir.Exists then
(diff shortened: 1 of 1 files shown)


forki:
THANKS!

JohnyWS:
Hmm, it seems this has been reintroduced again. :-/

I've tracked it down to 809513c2f5d0bcddc8062e387a9f54720a132187 which was a pretty big refactoring. It now respects completely what is in the nuspec file, however, that now conflicts with analyzers again.

Current issue for me, is when attempting to use https://www.nuget.org/packages/FakeItEasy.Analyzer.CSharp/ - version 5.2.0 has a /analyzers/dotnet/cs/FakeItEasy.Analyzer.CSharp.pdb file included, which is then added to the csproj files, which introduces compiler errors.

I'm not sure as to what is the proper way to fix it though, since the context of the original fix is gone, and I'm not sure having a global exclude pattern is the correct way to go? (e.g. exclude anything BUT *.dll when the path begins with "analyzers")

PR #2258, opened 2017-04-18
Update dependencies-file.md

0x53A:
Clarify references: strict

ref https://github.com/fsprojects/Paket/issues/2257


forki:
Thanks!

PR #2273, opened 2017-04-24
[wip/discussion] add a mechanism to auto-update the baseline files

0x53A:
How do you update the baseline files if a test fails? manually? or is there already another tool i missed?

If there isn't already a better way, and you don't mind, I would go through the rest of this file and adapt all remaining tests.

### DIFF
--- a/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/InstallSpecs.fs
@@ -22,6 +22,10 @@ open System.Diagnostics
 open Paket
 open Paket.Domain
 
+// set this to true and the tests will overwrite the baseline files with the new result.
+// then you can easily diff them in git and decide if the changes are OK or represent a bug.
+let updateBaselines = false
+
 [<Test>]
 let ``#1135 should keep lockfile stable when using framework restrictions``() = 
     let newLockFile = install "i001135-stable-install-on-framework-restrictions"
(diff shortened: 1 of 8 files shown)


0x53A:
looks like master is broken - I updated the baselines in the second commit, but you need to decide if that is actually ok or if it represents a bug.

cc @forki 

-----------------

No idea why - the last build from master was green ...

image

----------------

forki:
I like that. can you please write a bit about this in the readme?

0x53A:
Sure, will do that (prob. in the evening, after work) - do you have any idea why master seems to be broken?

matthid:
While I see the pain point there (and I was hit time and time again). This is at the same time dangerous as well. Just my regular disclaimer. we'd need to make sure the flag is never set in master and that we carefully review those changes in the tests now (I'm not sure this is done atm). maybe a big disclaimer is enough. THIS FLAG IS NOT MEANT FOR YOU TO INTRODUCE LARGE CHANGES EASILY.

0x53A:
For me, it was to just be able to compare it - the test outputs two gigantic strings (actual vs expected), good luck seeing the difference. With this you can override the baselines then use git to compare it.

forki:
Yes git compare is the trick that really works.

Am 24.04.2017 10:41 schrieb "Lukas Rieger" <notifications@github.com>:

matthid:
My proposed solution would be to improve the test output to show the diff notation as well... But yeah I'm not saying I'm against it :). Like I said I know the pain :)

matthid:
I just discussed this yesterday with @cloudRoutine as well on Slack

PR #2284, opened 2017-04-25
Refactoring

eugene-g:
getPackageDetails argument was ignored, so I deleted it and removed needless 2nd call to getPackageDetails.

PS Trailing whitespaces removed by [extension](https://github.com/Predelnik/RemoveTrailingWhitespaces). Should I avoid this in the future?


forki:
I think the correct fix would be to introduce the force parameter in tryV2 and tryV3 and then passing it down

eugene-g:
Got it. Well be done.

forki:
Are you sending a new pr? I think you found a potential bug.

PR #2293, opened 2017-04-27
Fix bug with ignored argument of getPackageDetails

eugene-g:
The argument "force" of getPackageDetails was ignored. PR fixes that issue by propagating argument to tryV2 and tryV3 functions.

Initially discussed [here](https://github.com/fsprojects/Paket/pull/2284).


forki:
Something is red here

eugene-g:
@forki I was planning to install VS2015 later today to dig into red tests (currently, project doesn't build on VS2017).

Is it fine now or I should proceed?

forki:
I think it was unrelated

Issue #2221, opened 2017-04-06
Failed to install http dependency with url parameter


vilinski:
### Description

Can't install http dependency. Seems like url parameters disturbs the file name resolution.

Background: Gogs (Go Git Server) allows accessing the repositonry files using token, which is provided as an url parameter. Here is the working link in the test repository:
https://try.gogs.io/mrbest/converter/raw/master/HttpDependency.fs?token=2d7f47cff7aaa721336e84a8f605888d525030c6

### Repro steps

- Extract [test.zip](https://github.com/fsprojects/Paket/files/902174/test.zip) or do `paket init` in any empty folder
- add following line to paket.dependencies:
`http https://try.gogs.io/mrbest/converter/raw/master/HttpDependency.fs?token=2d7f47cff7aaa721336e84a8f605888d525030c6`
-  `.paket/paket.exe install -v`

### Expected behavior

Download the file which is otherwise accessible with given url parameter

### Actual behavior

```
Paket version 4.1.6
found: C:\project\path\paket.dependencies
Parsing C:\project\path\paket.dependencies
Paket failed with:
        Illegales Zeichen im Pfad.
StackTrace:
     bei System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   bei System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPat
h)
   bei System.IO.FileInfo.Init(String fileName, Boolean checkHost)
   bei Paket.RemoteDownload.downloadDependenciesFile$cont@100[a](String rootPath, ResolvedSourceFile remoteFile, FSharpFunc`2 parserF, GroupName groupName, Boo
lean force, Unit unitVar)
   bei Paket.RemoteDownload.downloadDependenciesFile@97.Invoke(Unit unitVar)
   bei Microsoft.FSharp.Control.AsyncBuilderImpl.callA@841.Invoke(AsyncParams`1 args)
--- Ende der Stapelüberwachung vom vorhergehenden Ort, an dem die Ausnahme ausgelöst wurde ---
   bei Microsoft.FSharp.Control.AsyncBuilderImpl.commit[a](AsyncImplResult`1 res)
   bei Microsoft.FSharp.Control.CancellationTokenOps.RunSynchronouslyInCurrentThread[a](CancellationToken token, FSharpAsync`1 computation)
[...]

PR #2295, opened 2017-04-27
http dependency - strip query string to detect a file name

vilinski:
Fix and test for #2221


forki:
It's released in paket 5 alpha

Issue #2289, opened 2017-04-26
Paket 4.x install command takes hours to complete (while 3.x takes minutes)


pmbanka:
### Description

Paket 4.x install command takes a very long time to process my [`paket.dependencies`](https://github.com/pmbanka/paket-perf-repro/blob/master/paket.dependencies). I let it run for ~hour and then just killed it. Paket 3.x handled it without problems.

I'm on windows.

### Repro steps

1. Clone https://github.com/pmbanka/paket-perf-repro
2. Run `paket_old.exe install`
3. `git add .; git reset --hard`
4. Try running `paket.exe install`

### Expected behavior

Both versions should end running in short time.

### Actual behavior

`paket.exe install` takes > 1h

### Known workarounds

Use Paket 3.x.

0x53A:
I was a bit bored, so I let the profiler run for a while parallel to my normal work.

![image](https://cloud.githubusercontent.com/assets/4236651/25442766/34565f98-2aa6-11e7-93b4-895298c8619e.png)

The 30% inclusive in String.Concat is slightly concerning.

![image](https://cloud.githubusercontent.com/assets/4236651/25443011/e8ed2f04-2aa6-11e7-9356-b5151e7a9599.png)
![image](https://cloud.githubusercontent.com/assets/4236651/25443018/ed49965a-2aa6-11e7-98e9-a8fcc8b12327.png)

I think there are a few big quick-wins, will try to send a pr in a few minutes.

forki:
haha. we should remove verbosefn completely!

0x53A:
After eliminating that:

![image](https://cloud.githubusercontent.com/assets/4236651/25443311/c0580414-2aa7-11e7-9997-834c759c0fdc.png)

matthid:
yep very nice! The bug report really indicated some algorithmic problem (like the graph list you found, instead of the string concat). thanks for looking into this. I'm really waiting for these improvements :) (you saved me from some work I wanted to do eventually, thanks!)

forki:
> r.Graph is currently a List, I will try to convert that to a set.

I'm not sure if order is important. And I think the issue is equality checks on Requirements. they are very slow

0x53A:
@forki, yup, 43% of total time is Requirements.Compare. Luckily that compare function should have a bit of potential for improvements.

Changing list to set didn't improve it noticeably.

cloudRoutine:
[...]

Issue #2285, opened 2017-04-25
can't load PDBs for master anymore


0x53A:
When I press F11 to step into paket.exe:

![image](https://cloud.githubusercontent.com/assets/4236651/25411255/07d85be0-2a1c-11e7-806a-cac68db61fa8.png)

The following diff re-enabled debugging for me: https://github.com/0x53A/Paket/tree/debuggable

master: fdbb707a
VS 2017:
![image](https://cloud.githubusercontent.com/assets/4236651/25411308/5400bce2-2a1c-11e7-9fc3-e680a1dfba47.png)

0x53A:
cc @ctaggart

ctaggart:
Can you please file an appropriate bug about the portable pdb format not working yet in visualfsharp repro?

0x53A:
@ctaggart Can you reproduce it? It may also be an issue with my machine, or my settings, or whatever ..

ctaggart:
I'll check tomorrow.

ctaggart:
@0x53A, I'm not sure this is related to #2200. I just tried to reproduce the problem and couldn't using the latest from master.

I have Visual Studio 15.1 (26403.7) installed. Looks like you were on (26403.0).  I didn't see anything that stood out in https://www.visualstudio.com/en-us/news/releasenotes/vs2017-relnotes, but may be try updating. It doesn't look like I have the F# daily installed on this machine.

![image](https://cloud.githubusercontent.com/assets/80104/25613909/34b3eaf8-2f30-11e7-8654-9f038c82297a.png)

0x53A:
For me it is related - if i revert back to full pdbs then it works again. But I asked a few others and they don't have the same issue, so it seems to be related to my machine...

Thanks

ctaggart:
May be try updating Visual Studio. I'm not sure how to downgrade it, else I would try.

0x53A:
Already did, that issue was a few days old:
![image](https://cloud.githubusercontent.com/assets/4236651/25615710/04ea7fae-2f39-11e7-9bc1-1603cab3f714.png)

The only thing I can think of would be debugger settings, e.g. managed compatibility mode and so on, I will try to play a bit with them, maybe it will start working.

PR #2200, opened 2017-03-25
switch to SourceLink v2

ctaggart:
This is work-in-progress. It hasn't been as easy as I was hoping. The current issue is that if I try to do a `msbuild /t:build Paket.sln` or `msbuild /t:build Paket.fsproj` after a `msbuild /t:restore` I run into this error:

```
---------------------------------------------------------------------
  1) Building C:\projects\paket\src\Paket\Paket.fsproj failed with exitcode 1.
  2) : C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\Microsoft\NuGet\15.0\Microsoft.NuGet.targets(197,5): The project.json is referencing the project 'C:\projects\paket\src\Paket\Paket.Core\Paket.Core.fsproj', but an output path was not specified on an item in the ProjectReferencesCreatingPackages property.
```

As far as I can tell, it has something to do with https://github.com/NuGet/Home/issues/4532

Essentially, all I want to do with MSBuild 15 is:
```
msbuild /t:restore Paket.sln
msbuild /t:build Paket.sln
```

The good news is if I limit it to just Paket.Core, it does work:
``` powershell
C:\Users\CameronTaggart\fs\Paket\src\Paket.Core [sourcelink2 ≡]> dotnet sourcelink print-json "C:\Users\CameronTaggart\Downloads\bin\Paket.Core.pdb"
{"documents":{"C:\\projects\\paket\\*":"https://raw.githubusercontent.com/ctaggart/Paket/0910a5678cb1542217efc61dfddfcf2c2d32cc6b/*"}}

C:\Users\CameronTaggart\fs\Paket\src\Paket.Core [sourcelink2 ≡]> dotnet sourcelink print-urls "C:\Users\CameronTaggart\Downloads\bin\Paket.Core.pdb"
ed4678f7c9fe6396b80fbcd4033ea784 md5 fsharp C:\projects\paket\src\Paket.Core\SemVer.fs
https://raw.githubusercontent.com/ctaggart/Paket/0910a5678cb1542217efc61dfddfcf2c2d32cc6b/src/Paket.Core/SemVer.fs
2c2b3a0f5f57dae031bd6e2775d7a8ea md5 fsharp C:\projects\paket\src\Paket.Core\VersionRange.fs
https://raw.githubusercontent.com/ctaggart/Paket/0910a5678cb1542217efc61dfddfcf2c2d32cc6b/src/Paket.Core/VersionRange.fs
[...]

PR #2299, opened 2017-04-28
[Ready for Review] Slightly improve performance

0x53A:
Let's see what the CI says.

ref https://github.com/fsprojects/Paket/issues/2289

### DIFF
--- a/src/Paket.Core/Dependencies/PackageResolver.fs
+++ b/src/Paket.Core/Dependencies/PackageResolver.fs
@@ -654,16 +654,22 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
         let inline fuseConflicts currentConflict priorConflictSteps conflicts =
             let findMatchingStep priorConflictSteps =
                 priorConflictSteps // row
-                |> Seq.tryFind (fun (_,_,lastRequirement:PackageRequirement,_,_) ->
+                |> Seq.tryFindIndex (fun (_,_,lastRequirement:PackageRequirement,_,_) ->
                     let currentNames =
                         conflicts |> Seq.collect (fun c ->
                             let graphNameList =
                                 c.Graph |> List.map (fun (pr:PackageRequirement) -> pr.Name) 
                             c.Name :: graphNameList)
                         |> Seq.toArray
-                    currentNames |> Array.contains lastRequirement.Name                        
-                )
-                |> Option.map (fun r -> r, priorConflictSteps |> List.filter(fun r2 -> not (obj.ReferenceEquals(r, r2))))
+                    currentNames |> Array.contains lastRequirement.Name)
+                |> Option.map (fun i ->
+                    let row = priorConflictSteps |> List.item i
+                    let priorExceptRow =
+                        priorConflictSteps
+                        |> List.mapi (fun i2 v -> i2,v)
+                        |> List.filter (fun (i2, _) -> i2 <> i)
+                        |> List.map snd
+                    row, priorExceptRow)
 
             match priorConflictSteps, findMatchingStep priorConflictSteps with
             | [], _  -> currentConflict
--- a/src/Paket.Core/Dependencies/PackageResolver.fs
+++ b/src/Paket.Core/Dependencies/PackageResolver.fs
@@ -653,8 +653,8 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
 
         let inline fuseConflicts currentConflict priorConflictSteps conflicts =
             let findMatchingStep priorConflictSteps =
-                priorConflictSteps // row
-                |> Seq.tryFindIndex (fun (_,_,lastRequirement:PackageRequirement,_,_) ->
+                priorConflictSteps
+                |> List.tryExtractOne (fun (_,_,lastRequirement:PackageRequirement,_,_) ->
                     let currentNames =
                         conflicts |> Seq.collect (fun c ->
                             let graphNameList =
@@ -662,24 +662,16 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
                             c.Name :: graphNameList)
                         |> Seq.toArray
                     currentNames |> Array.contains lastRequirement.Name)
-                |> Option.map (fun i ->
-                    let row = priorConflictSteps |> List.item i
-                    let priorExceptRow =
-                        priorConflictSteps
-                        |> List.mapi (fun i2 v -> i2,v)
-                        |> List.filter (fun (i2, _) -> i2 <> i)
-                        |> List.map snd
-                    row, priorExceptRow)
 
-            match priorConflictSteps, findMatchingStep priorConflictSteps with
-            | [], _  -> currentConflict
-            | _, Some (head, priorConflictSteps) ->
+            match findMatchingStep priorConflictSteps with
+            | None, []  -> currentConflict
+            | (Some head), priorConflictSteps ->
                 let (lastConflict, lastStep, lastRequirement, lastCompatibleVersions, lastFlags) = head
                 let continueConflict = 
                     { currentConflict with VersionsToExplore = lastConflict.VersionsToExplore }        
                 step (Inner((continueConflict,lastStep,lastRequirement), priorConflictSteps))  stackpack lastCompatibleVersions lastFlags
             // could not find a specific package - go back one step
-            | head :: priorConflictSteps, None ->
+            | None, head :: priorConflictSteps ->
                 let (lastConflict, lastStep, lastRequirement, lastCompatibleVersions, lastFlags) = head
                 let continueConflict = 
                     { currentConflict with VersionsToExplore = lastConflict.VersionsToExplore }        
--- a/src/Paket.Core/Dependencies/PackageResolver.fs
+++ b/src/Paket.Core/Dependencies/PackageResolver.fs
@@ -272,7 +272,7 @@ let calcOpenRequirements (exploredPackage:ResolvedPackage,globalFrameworkRestric
             Settings = { dependency.Settings with FrameworkRestrictions = newRestrictions } })
     |> Set.filter (fun d ->
         resolverStep.ClosedRequirements
-        |> Seq.exists (fun x ->
+        |> Set.exists (fun x ->
             x.Name = d.Name &&
                x.Settings.FrameworkRestrictions = d.Settings.FrameworkRestrictions &&
                 (x = d ||
@@ -281,7 +281,7 @@ let calcOpenRequirements (exploredPackage:ResolvedPackage,globalFrameworkRestric
         |> not)
     |> Set.filter (fun d ->
         resolverStep.OpenRequirements
-        |> Seq.exists (fun x -> x.Name = d.Name && (x = d || x.VersionRequirement.Range.IsGlobalOverride) && x.Settings.FrameworkRestrictions = d.Settings.FrameworkRestrictions)
+        |> Set.exists (fun x -> x.Name = d.Name && (x = d || x.VersionRequirement.Range.IsGlobalOverride) && x.Settings.FrameworkRestrictions = d.Settings.FrameworkRestrictions)
         |> not)
     |> Set.union rest
 
@@ -390,7 +390,7 @@ let private explorePackageConfig getPackageDetailsF  (pkgConfig:PackageConfig) =
 
 type StackPack = {
     ExploredPackages     : Dictionary<PackageName*SemVerInfo,ResolvedPackage>
-    KnownConflicts       : HashSet<Set<PackageRequirement> * ((SemVerInfo * PackageSource list) list * bool) option>
+    KnownConflicts       : HashSet<HashSet<PackageRequirement> * ((SemVerInfo * PackageSource list) list * bool) option>
     ConflictHistory      : Dictionary<PackageName, int>
 }
 
@@ -494,23 +494,26 @@ let private getCompatibleVersions
         compatibleVersions, false, tryRelaxed
 
 
-let private getConflicts (currentStep:ResolverStep) (currentRequirement:PackageRequirement) (knownConflicts:HashSet<_>) =
+let private getConflicts (currentStep:ResolverStep) (currentRequirement:PackageRequirement) (knownConflicts:HashSet<HashSet<PackageRequirement> * ((SemVerInfo * PackageSource list) list * bool) option>) =
+    
     let allRequirements =
-        currentStep.OpenRequirements
-        |> Set.filter (fun r -> r.Graph |> List.contains currentRequirement |> not)
-        |> Set.union currentStep.ClosedRequirements
+        Set.toSeq currentStep.OpenRequirements
+        |> Seq.filter (fun r -> r.Graph |> List.contains currentRequirement |> not)
+        |> Seq.append currentStep.ClosedRequirements
+        |> HashSet
 
     knownConflicts
     |> Seq.map (fun (conflicts,selectedVersion) ->
         match selectedVersion with
-        | None when Set.isSubset conflicts allRequirements -> conflicts
+        | None when conflicts.IsSubsetOf allRequirements -> conflicts
         | Some(selectedVersion,_) ->
             let n = (Seq.head conflicts).Name
             match currentStep.FilteredVersions |> Map.tryFind n with
-            | Some(v,_) when v = selectedVersion && Set.isSubset conflicts allRequirements -> conflicts
-            | _ -> Set.empty
-        | _ -> Set.empty)
-    |> Set.unionMany
+            | Some(v,_) when v = selectedVersion && conflicts.IsSubsetOf allRequirements -> conflicts
+            | _ -> HashSet()
+        | _ -> HashSet())
+    |> Seq.collect id
+    |> HashSet
 
 
 let private getCurrentRequirement packageFilter (openRequirements:Set<PackageRequirement>) (conflictHistory:Dictionary<_,_>) =
@@ -575,7 +578,7 @@ let private boostConflicts
         match conflicts with
         | c::_  ->
             let selectedVersion = Map.tryFind c.Name filteredVersions
-            let key = conflicts |> Set.ofList,selectedVersion
+            let key = conflicts |> HashSet,selectedVersion
             stackpack.KnownConflicts.Add key |> ignore
 
             let reportThatResolverIsTakingLongerThanExpected =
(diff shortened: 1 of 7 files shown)


src/Directory.Build.props
```diff
@@ -1,8 +1,8 @@
 <Project>
   <PropertyGroup>
-    <DebugType>portable</DebugType>
+    <!--<DebugType>portable</DebugType>-->
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.0" PrivateAssets="all" />
+    <!--<PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.0" PrivateAssets="all" />-->
```
cloudRoutine: Why are these commented now?

src/Directory.Build.targets
```diff
@@ -10,5 +10,5 @@
     <CreateProperty Value="@(EmbeddedFiles)">  
       <Output TaskParameter="Value" PropertyName="embed" />  
     </CreateProperty>
-  </Target>
+  </Target>-->
```
cloudRoutine: Why is this commented now?

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -494,23 +494,26 @@ let private getCompatibleVersions
...
+        Set.toSeq currentStep.OpenRequirements
+        |> Seq.filter (fun r -> r.Graph |> List.contains currentRequirement |> not)
+        |> Seq.append currentStep.ClosedRequirements
+        |> HashSet
 
     knownConflicts
     |> Seq.map (fun (conflicts,selectedVersion) ->
         match selectedVersion with
-        | None when Set.isSubset conflicts allRequirements -> conflicts
+        | None when conflicts.IsSubsetOf( allRequirements ) -> conflicts
```
cloudRoutine: doesn't need parens

src/Directory.Build.props
0x53A: https://github.com/fsprojects/Paket/issues/2285

src/Directory.Build.targets
0x53A: https://github.com/fsprojects/Paket/issues/2285

src/Directory.Build.props
cloudRoutine: cool 👍

0x53A:
@cloudRoutine Thanks for taking a look, but that is a really early version. =)

Stylistically, I need to clean it up and revert a bunch.

The relevant change to review is ``fuseConflicts`` and especially the condition leading to
https://github.com/fsprojects/Paket/pull/2299/files#diff-67971315cd3928195536b7ac5f0428fbR666

src/Directory.Build.props
0x53A: I will revert it after finishing this pr and before it get's merged, that was just so I could debug it locally.
Someone else should take care of this, either fix it or just roll https://github.com/fsprojects/Paket/pull/2200/files back.

0x53A:
How confident are ya all that the unit tests will find potential resolve issues? :D

src/Directory.Build.props
0x53A: @cloudRoutine does debugging work for you?

cloudRoutine:
very confident, working on this one is annoying isn't it? 😆 
```
1) Error : Paket.DependencyGraphSpecs.should solve strange graph
System.Exception : There was a version conflict during package resolution.
  Resolved packages:
   - P3 1.1.3
   - P7 11.10.10.3
  Could not resolve package P8 <= 0.2.8:
   - No versions available.
  Please try to relax some conditions or resolve the conflict manually 
...
2) Error : Paket.UpdateProcess.Test.SelectiveUpdate considers package name case difference
System.Exception : There was a version conflict during package resolution.
  Resolved packages:
   - Castle.Core-log4net 4.0.0
  Could not resolve package castle.core 3.2:
   - Castle.Core-log4net 4.0.0 requested package castle.core: >= 4.0
  Please try to relax some conditions or resolve the conflict manually 
...
3) Error : Paket.UpdateProcess.Test.SelectiveUpdate does not update when a dependency constrain is not met
System.Exception : There was a version conflict during package resolution.
  Resolved packages:
   - Castle.Core-log4net 4.0.0
  Could not resolve package Castle.Core 3.2:
   - Castle.Core-log4net 4.0.0 requested package Castle.Core: >= 4.0
  Please try to relax some conditions or resolve the conflict manually 
 ...
```
it looks like it's cutting out too quickly now

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -681,7 +717,7 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
                         let flags = 
                             StepFlags(flags.Ready,flags.UseUnlisted,flags.HasUnlisted,true,flags.FirstTrial,flags.UnlistedSearch)
```
cloudRoutine: Now that we're on F# 4.1 StepFlags should be changed to a struct record

cloudRoutine:
@0x53A you're checking the IL to make sure it stays tailrec right? (just making sure)

src/Directory.Build.props
cloudRoutine: Debugging does work for me, although I never found debugging particularly useful for working on this function. Way too many iterations to go through, but maybe I should have tried conditional breakpoints.

0x53A:
@cloudRoutine If it weren't, it would stackoverflow =)

I have "solved" the failing UTs by falling back to the old behaviour (go back one) if it can't find a matching step to fall back to, but that probably just masks an issue within ``findMatchingStep ``.

cloudRoutine:
Were you encountering stack overflows on the unit tests when it wasn't tailrec? I'm not sure that any of them pull in enough dependencies to cause it, and I can't remember if I added one to make sure 😄

0x53A:
Not on the unit tests, but on the dependencies file from the original issue.

You can see my ``let x = ...; x`` dance inhttps://github.com/fsprojects/Paket/pull/2299/commits/2fa007c2e5a0670541bda824431eb55ca2bd3459, that was to disable tailrec, so that I could properly debug it.

src/Paket/Paket.fsproj
```diff
@@ -35,9 +36,13 @@
     <StartArguments>install</StartArguments>
     <StartWorkingDirectory>C:\dev\src\Paket\integrationtests\scenarios\loading-scripts\dependencies-file-flag\temp</StartWorkingDirectory>
     <StartArguments>install</StartArguments>
+    <StartWorkingDirectory>C:\Users\lr\Source\Repos\castos</StartWorkingDirectory>
+    <StartWorkingDirectory>C:\Users\lr\Source\Repos\paket-perf-repro</StartWorkingDirectory>
```
cloudRoutine: don't forget to cut these out

cloudRoutine:
Can you add that dependencies file as a test case?

0x53A:
Soo ci is green again. Will do the rest on the evening.

cloudRoutine:
Great work on this 💪

src/Paket.Core/Versioning/Requirements.fs
```diff
@@ -791,25 +791,30 @@ type PackageRequirement =
     member this.Depth = this.Graph.Length
 
     static member Compare(x,y,startWithPackage:PackageFilter option,boostX,boostY) =
-        if x = y then 0 else
-        seq {
-            yield compare
-                (not x.VersionRequirement.Range.IsGlobalOverride,x.Depth)
-                (not y.VersionRequirement.Range.IsGlobalOverride,y.Depth)
-            yield match startWithPackage with
+        if obj.ReferenceEquals(x, y) then 0 else
```
cloudRoutine: why was `obj.ReferenceEquals` necessary instead of `=`? just curious

0x53A: Not necessary, but performance improvement.
It first compared the whole objects, then compared each member again.

Profiler showed a small improvement.

cloudRoutine: I thought `=` made use of custom equality, or is the `obj.ReferenceEquals` faster than that?

0x53A: yes and yes.

ref-equals is just a single pointer comparison, the custom equality is one (potentially de-virtualized) virtual call, one type-test (https://github.com/0x53A/Paket/blob/932400899c3927c823cc9b869de545de71c650f2/src/Paket.Core/Versioning/Requirements.fs#L771) and up to 6 member comparisons. And if any of these members are non-equal, then they are just compared again in Compare itself.

Afaik profiling showed a single-digit-% performance improvement, but I want to rerun these tests anyway.

0x53A:
I think another quick win may be ordering the Requirements so that fixed versions are resolved first.

E.g. I was testing https://github.com/fsprojects/Paket/issues/2294#issuecomment-297918543 and noticed the following warning:

```
 Could not resolve package NETStandard.Library 1.6:
   - Amazon.Lambda.APIGatewayEvents 1.1.0 requested package NETStandard.Library: >= 1.6
   - Amazon.Lambda.AspNetCoreServer 0.8.1-preview1 requested package NETStandard.Library: >= 1.6
   - FluentAssertions 4.19.2 requested package NETStandard.Library: >= 1.6
   - Microsoft.AspNetCore.Mvc.Core 1.1.2 requested package NETStandard.Library: >= 1.6.1
   - Microsoft.Extensions.Configuration.Json 1.1.0 requested package NETStandard.Library: >= 1.6.1
   - xunit.extensibility.execution 2.3.0-beta1-build3642 requested package NETStandard.Library: >= 1.6
```

But logically speaking, that warning is backwards.
NETStandard.Library is pinned and doesn't need resolving, the other packages need to be resolved for it.

forki:
ordering is very important. In general the heuristic was to resolve things with high conflict potential first. then fixed versions.

forki:
can we please get a mergeable PR before you start new work? Just to get this into the bank.

0x53A:
I'm not really happy with how the current check works - it just compares the name of the conflicting requirements to the names of the previously solved requirements and if it can't find one it just re-solves the previous one (which is what it did before). So there may (and surely are) still degenerate dependency files which will take forever to solve.

The correct way would be to walk the whole dependency chain and check if any conditions actually conflicts.

My plan (for this evening) was to split this in two PRs:

* basic performance improvements (like the Comparison method)
* the change to the Resolver

Now I don't think the resolver change is bad - the unit tests pass, so it shouldn't actually make it worse - but it is far from perfect, so it may be worth it to pull now anyway and improve it later.

cloudRoutine:
I'll clean this up and get it merge ready

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -681,7 +726,7 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
...
                        
-                        step (Inner((continueConflict,lastStep,lastRequirement),priorConflictSteps)) stackpack lastCompatibleVersions  lastFlags 
+                        step (Inner((continueConflict,lastStep,lastRequirement),priorConflictSteps)) stackpack lastCompatibleVersions  flags
```
0x53A: this was left-over from one of my tests, I didn't mean to commit it. flags is created above, but never used, to it looked suspicious.

What do you think @cloudRoutine?

Unit tests are green WITH this change, I plan to revert it and rerun the unit tests ...

cloudRoutine: yea I was wondering about this, it's been a few months so I'm not totally sure what it should have been, but I think I may have meant to write
```fsharp
let flags =StepFlags(lastFlags.Ready,lastFlags.UseUnlisted,lastFlags.HasUnlisted,true,lastFlags.FirstTrial,lastFlags.UnlistedSearch)
                        
step (Inner((continueConflict,lastStep,lastRequirement),priorConflictSteps)) stackpack lastCompatibleVersions  flags 
```

0x53A: Since the unit tests seem to be green either way - how would this affect execution?

cloudRoutine: the unit tests aren't are reliable check on restore's functionality, the full integration test suite is needed to put it through it's paces

0x53A: Yes, but the serverbuild was green before this PR, and it is now green with this PR, so unless this is also affected by the change to ``fuseConflicts`` it looks like it doesn't matter.

Or are the integration tests not run on PRs?

cloudRoutine: now that I actually have the code in front of me in VS, that's the force break flag, so that should just bail us out of resolution faster and it probably shouldn't make much of a difference what the other flags are (don't quote me on that ;P)

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -649,14 +652,56 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
...
+                    | [] -> currentConflict
+                    | priorConflictSteps ->
+                        match findMatchingStep currentConflict priorConflictSteps with
+                        | Some((lastConflict,lastStep,lastRequirement,lastCompatibleVersions,lastFlags), priorConflictSteps) ->
+                            let continueConflict = 
+                                { currentConflict with VersionsToExplore = lastConflict.VersionsToExplore }        
+                            step (Inner((continueConflict,lastStep,lastRequirement),priorConflictSteps))  stackpack lastCompatibleVersions lastFlags
+                        | None ->
+                            // could not find a specific package - go back one
+                            let (lastConflict,lastStep,lastRequirement,lastCompatibleVersions,lastFlags) :: priorConflictSteps = priorConflictSteps
```
cloudRoutine: Is this supposed to pull the head off of `priorConflictSteps` as `(lastConflict,lastStep,lastRequirement,lastCompatibleVersions,lastFlags)` and then alias it's tail as `priorConflictSteps` ?

0x53A: exactly, this uses the previous behaviour as a fallback if it can't find a specific one

0x53A: This was required to fix the three failing tests

cloudRoutine:
@0x53A yea you should probably make a new PR and just cherry pick what you need onto it, all the commits between your initial ones and now made rebasing a big pain.

0x53A:
If you don't mind I would just squash it all into one

0x53A:
It's not safe to compare tuples by reference (see https://github.com/Microsoft/visualfsharp/issues/2923), so I have modified findMatchingStep slightly. It now gets the index first.

Do you know of a nicer way to remove a single index from a list?

0x53A:
I did a quick benchmark, and it looks like foldback has a relatively high overhead: https://gist.github.com/0x53A/d6d30523e04c7e1c448668546c6d69f9

forki:
Do we have any numbers in how perf improves?

0x53A:
The Resolving is only cpu bound if the algorithm degrades. That means the really relevant part was the change to ``fuseConflicts``, the other performance optimizations would be unnecessary if the algorithm was perfect.

Currently it still degrades in a few situations, but it is already better than before.

Especially the last two commits are completely unnecessary performance-wise, my main motivation was readability and extracting it to a helper function, the benchmark was just curiosity.

---------------------

For performance testing, it would be great if you had any other dependencies files which did correctly solve before (so that we can actually compare numbers), and took maybe a minute or two, so that we don't need to wait too long.

For the next round of improvements I want to set up a benchmark script that checks out a git sha, builds it, and runs a few paket installs against a few different paket.dependencies.

----------------

Before: https://github.com/fsprojects/Paket/releases/tag/5.0.0-alpha007
Now: commit f5deb485

-----------------

1) from https://github.com/fsprojects/Paket/issues/2289

<details>
  <summary>#2289</summary>

redirects: on
framework: net46

nuget Akka.Persistence.SqlServer
nuget SQLProvider
nuget Microsoft.FSharpLu.Json
nuget Microsoft.Net.Http
nuget Microsoft.AspNet.WebApi.Client
nuget FSharp.Core redirects: force
nuget FSharp.Configuration
nuget FSharp.Formatting
nuget FSharp.Data
nuget FSharp.Data.TypeProviders
nuget Newtonsoft.Json
nuget FAKE
nuget SourceLink.Fake
nuget Akka
nuget Akka.FSharp
nuget Akka.Persistence
nuget Akka.Persistence.FSharp
nuget Akka.Persistence.Sql.Common
nuget Akka.Persistence.Query
nuget Akka.Persistence.Query.Sql
nuget Akka.Logger.Serilog
nuget Akka.Monitoring
nuget Akka.Monitoring.StatsD
nuget Akka.TestKit.NUnit
nuget Akka.TestKit.Xunit2
nuget Akka.Serialization.Wire
nuget FsPickler
nuget Serilog
nuget Serilog.Sinks.ElasticSearch
nuget Nest ~> 5
nuget RabbitMQ.Client
nuget SSH.NET
nuget Topshelf
nuget Topshelf.Serilog
nuget NUnit ~> 2
nuget NUnit.Runners
nuget NUnitTestAdapter
nuget TickSpec.NUnit content: none
nuget FsCheck
nuget FsCheck.Nunit
nuget System.Collections.Immutable
nuget Suave
nuget OctopusTools
nuget Suave.Swagger
nuget Quartz
nuget Akka.Quartz.Actor
nuget Oracle.ManagedDataAccess

</details>


Before: didn't finish after an hour
Now:  ``44 seconds - ready.``

----------------------------

2)  from https://github.com/fsprojects/Paket/issues/2294#issuecomment-298120468

The original paket.dependencies from the issue fails with an error, but this reduced one resolves correctly. (see https://github.com/fsprojects/Paket/pull/2305 for the failure)

```

nuget NETStandard.Library <= 1.6.0
nuget Microsoft.AspNetCore.Mvc.Core
```

Now: ``5 minutes, 1 second - ready.``
Before: still running after aprox > 10 minutes, will update when / if it finishes.

matthid:
Just a note on performance: We have two  resolver states since I introduced the runtime stuff. It might make sense to measure them separately... (or at least it's hard to compare them with previous paket versions)

0x53A:
With an older version (I assume pre 5.alpha) cloudroutine could resolve the second file in 2 minutes, now it takes me 5 minutes on this branch, so this may be worth investigating.

ref https://github.com/fsprojects/Paket/issues/2294#issuecomment-298120468

But it seems unrelated to this PR, because with 5.0alpha I'm still waiting for it to finish, and 5 minutes is better than never =).

0x53A:
@forki @cloudRoutine @matthid if you have some time, could you please test if you find any dependencies files that get slower after this pr?

If you just send me the paket.dependencies then I can also test it myself.

0x53A:
regarding the last commit:

before:
image

image

The name array was recalculated for each iteration.

now:
image

src/Paket.Core/Versioning/Requirements.fs
```diff
@@ -791,25 +791,30 @@ type PackageRequirement =
...
+        let c = compare boostX boostY
+        if c <> 0 then c else
+        let c = -compare x.VersionRequirement y.VersionRequirement
+        if c <> 0 then c else
+        let c = compare x.Settings.FrameworkRestrictions y.Settings.FrameworkRestrictions
+        if c <> 0 then c else
+        let c = compare x.Parent y.Parent
+        if c <> 0 then c else
+        let c = compare x.Name y.Name
+        if c <> 0 then c else 0
```
matthid: How much does this block yield? Because its really ugly :P

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -787,8 +815,7 @@ let Resolve (getVersionsF, getPackageDetailsF, groupName:GroupName, globalStrate
 
                 | stackpack, Some(alreadyExplored,exploredPackage) ->
                     let hasUnlisted = exploredPackage.Unlisted || flags.HasUnlisted
-                    let flags = 
-                        StepFlags(flags.Ready,flags.UseUnlisted,hasUnlisted,flags.ForceBreak,flags.FirstTrial,flags.UnlistedSearch)
+                    let flags = { flags with HasUnlisted = hasUnlisted }
```
cloudRoutine:
```fsharp
let flags = { flags with HasUnlisted = exploredPackage.Unlisted || flags.HasUnlisted }
```

matthid:
@0x53A After #2307 We could compare performance of these improvements with older paket versions as well (by setting the env var `PAKET_DISABLE_RUNTIME_RESOLUTION` to `true` ).
Currently the comparison is not really "fair" because we actually do a second resolver run...

matthid:
I combined your improvements with #2307 in https://github.com/matthid/Paket/tree/combine_fixes but couldn't see any real benefits (But I didn't really measure anything)

0x53A:
In most cases, paket shouldn't be cpu bound, so any performance improvements in this area won't actually matter.

It will only actually improve performance if there are a lot of conflicts, and the solving algorithm degrades.

src/Paket.Core/Versioning/Requirements.fs
0x53A: Because the old code was a thing of beauty? :D

According to my benchmark (take it with a truckload of salt) it is slightly faster: https://gist.github.com/0x53A/b8bb15e040d152742fd067ce257e80a2

 Method |     Mean |     Error |    StdDev |  Gen 0 | Allocated |
------- |---------:|----------:|----------:|-------:|----------:|
    Old | 656.1 ns | 6.2845 ns | 4.9066 ns | 0.0508 |     264 B |
    New | 105.5 ns | 0.3383 ns | 0.2641 ns |      - |       0 B |

PR #2301, opened 2017-04-28
add FixNuspecs to public api

cloudRoutine:
- Add FixNuspecs as a method on `Dependencies`
- Change argu DU to accept a list and mark args as mandatory to simply implementation


forki:
Breaking change?

cloudRoutine:
the name is different

forki:
Yes but was it already released? If so then can we create a function with old name that redirects to keep it stable?

cloudRoutine:
I think you're the only one who's used it in
```
   <Exec Command='$(PaketCommand) fix-nuspec file "@(_NuspecFiles)" references-file "$(PaketReferencesFilePath)" ' />
```

forki:
So this makes it everyone on core

cloudRoutine:
don't the targets files get updated when paket gets updated?

forki:
Nope. Only on install

cloudRoutine:
maybe we should change that, I'll add the obsolete arg regardless

forki:
Yes but even then people need to update the bootstrapper first.

cloudRoutine:
checked to make sure, backwards compatibility still works

PR #2316, opened 2017-05-06
Docs: Update FAQ to avoid confusion between install and restore

cdrnet:
It seems the explanation on why the lock file should usually be committed is confusing install and restore. To my understanding this is relevant mostly for restore, much less for install (also, you usually don't run install in CI).


forki:
Thx

PR #2319, opened 2017-05-07
align with spec requirements, meaning no leading zeroes on semVers

baronfel:
per the spec:

>A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and **MUST NOT** contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.

Emphasis mine

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/InstallModel.fs
+++ b/src/Paket.Core/PaketConfigFiles/InstallModel.fs
@@ -12,16 +12,16 @@ type UnparsedPackageFile = Paket.NuGet.UnparsedPackageFile
 type Tfm = PlatformMatching.ParsedPlatformPath
 
 //type Rid = Paket.Rid
-type FrameworkDependentFile = { 
+type FrameworkDependentFile = {
     Path : Tfm
     File : UnparsedPackageFile
-    Runtime : Rid option 
+    Runtime : Rid option
 }
 
-type Library = { 
+type Library = {
     Name : string
     Path : string
-    PathWithinPackage : string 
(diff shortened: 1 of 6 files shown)


baronfel:
Actually found this when I was integrating this module into FAKE, thanks to the tests over there.

cdrnet:
Note that `0` starts with `0` as well, and `0.1.0` is a perfectly valid version number.

forki:
We also need to support 0 itself since it is used in “=> 0" all over the
place

Am 08.05.2017 8:28 vorm. schrieb "Christoph Ruegg" <notifications@github.com

forki:
ping

baronfel:
Sorry this has taken so long, I'm in vacation without a computer but can
take a look on Saturday.

baronfel:
ok this is looking fixed up now.  I'm super curious what it would take to move us over to [SemVer.net](https://github.com/xyncro/semver.net) instead of a home grown parse, and it looks like we mostly have a 
1) semi-convention around the 'name' of the prerelease segment being the first alpha segment,
1) various tostring functions that normalize for nuget (which may not be necessary anymore if the 3.4+ documentation is correct in that they strip leading zeroes, and trailing 4th segment zeroes)

forki:
@baronfel the issue is always to support old released nupkg that don't conform to semver

forki:
we recently made paket more conform to semver. So hope it's better now.

PR #2322, opened 2017-05-09
use SourceLink for local builds too

ctaggart:
I [noticed](https://github.com/ctaggart/SourceLink/issues/181#issuecomment-299630165) that the published nupkg didn't have source link enabled. It is only enabled by default on build servers like AppVeyor. This enables it when run via FAKE. Once a [FAKE helper](https://github.com/ctaggart/SourceLink/issues/212) is added for `dotnet sourcelink test`, it would be good to add a test build step.


forki:
thanks!

Issue #1123, opened 2015-10-09
Comments are breaking framework restrictions in paket.references


vbfox:
Let's take this `paket.references` file copy-pasted directly [from the documentation](https://fsprojects.github.io/Paket/references-files.html) :

```text
Newtonsoft.Json framework: net35, net40  // .NET 3.5 and .NET 4.0
DotNetZip framework: >= net45            // .NET 4.5 and above
```

And observe the generated .csproj file :

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Choose>
    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And ($(TargetFrameworkVersion) == 'v2.0' Or $(TargetFrameworkVersion) == 'v3.0' Or $(TargetFrameworkVersion) == 'v3.5' Or $(TargetFrameworkVersion) == 'v4.0' Or $(TargetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersion) == 'v4.5.1' Or $(TargetFrameworkVersion) == 'v4.5.2' Or $(TargetFrameworkVersion) == 'v4.5.3' Or $(TargetFrameworkVersion) == 'v4.6')">
      <ItemGroup>
        <Reference Include="Ionic.Zip">
          <HintPath>..\packages\DotNetZip\lib\net20\Ionic.Zip.dll</HintPath>
          <Private>True</Private>
          <Paket>True</Paket>
        </Reference>
      </ItemGroup>
    </When>
  </Choose>
  <Choose>
    <When Condition="$(TargetFrameworkIdentifier) == '.NETFramework' And $(TargetFrameworkVersion) == 'v3.5'">
      <ItemGroup>
        <Reference Include="Newtonsoft.Json">
          <HintPath>..\packages\Newtonsoft.Json\lib\net35\Newtonsoft.Json.dll</HintPath>
          <Private>True</Private>
          <Paket>True</Paket>
        </Reference>
      </ItemGroup>
    </When>
  </Choose>
</Project>
```

Paket acted as if the file didn't contain the last element on each line :

```text
Newtonsoft.Json framework: net35
DotNetZip
```

forki:
looks like we don't support comments yet.

would change this to: "remove the comments from docs" - PR very welcome.

Implementing comments might be also possible, but it's not that easy as the implementation in the dependencies file shows.

PR #2348, opened 2017-05-18
remove unsupported comments from code example

pr-yemibedu:
remove unsupported comments from code example.
similar to https://github.com/fsprojects/Paket/issues/1123
also cleaned up the styling in that section to keep same information.

### DIFF
--- a/docs/content/references-files.md
+++ b/docs/content/references-files.md
@@ -47,13 +47,16 @@ If you don't want to import `.targets` and `.props` files you can disable it via
 
 ## No content option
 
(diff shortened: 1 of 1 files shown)


pr-yemibedu:
Hello,
 I don't know if this file touches the web page for [no content option](https://fsprojects.github.io/Paket/references-files.html#No-content-option) as that is what people see first. If not, that would also need a to be changed. I can make an additional PR if needed. Thank you. Good day.

forki:
thanks!

PR #2380, opened 2017-05-29
downgrade to tooling 1.0

0x53A:
see https://docs.microsoft.com/en-us/dotnet/standard/library

ref https://github.com/fsprojects/Paket/issues/2378


0x53A:
I didn't run any tests locally, let's see what the CI says.

0x53A:
@forki What's up with the build? :D

matthid:
Can you pick the MSBuild fixes from https://github.com/fsprojects/Paket/pull/2362?
Sorry https://github.com/fsprojects/Paket/pull/2362 is not ready to merge jet, and I need a couple of days to fix ... I will take care of it if you can't

0x53A:
I would prefer to just add the nuget FSharp.Compiler.Tools instead of hardcoding a few different paths. What do you think?

matthid:
Yes if it works I'm fine with that. nice!

0x53A:
Build now works, but one unit test failed. On it!

The great thing about the nuget is that this will allow compatibility with VS 2015 and F# vLatest. For example, if you use struct tuples, you will get red squiggles in the ide, but can still build and run / debug.

matthid:
Yes very nice! btw you can test with the appveyor artefact how this change will impact your scenario (sorry i'm on mobile today and tomorrow...) or just build locally obviously

0x53A:
wp_ss_20170531_0001
Av is green, travis failed with a weird error.

Please review if the unit test changes are actually correct.

tests/Paket.Tests/InstallModel/ProcessingSpecs.fs
```diff
@@ -154,7 +154,7 @@ let ``should understand reference folder``() =
         model.GetRuntimeAssemblies RuntimeGraph.Empty (Rid.Of "unix") (SinglePlatform (DotNetFramework FrameworkVersion.V4_6_3))
         |> Seq.map (fun f -> f.Library.Path)
     refs |> shouldNotContain @"..\System.Security.Cryptography.Algorithms\runtimes\win\lib\net46\System.Security.Cryptography.Algorithms.dll"
-    refs |> shouldContain @"..\System.Security.Cryptography.Algorithms\lib\netstandard1.6\System.Security.Cryptography.Algorithms.dll"
+    refs |> shouldNotContain @"..\System.Security.Cryptography.Algorithms\lib\netstandard1.6\System.Security.Cryptography.Algorithms.dll"
     refs |> shouldNotContain @"..\System.Security.Cryptography.Algorithms\ref\netstandard1.6\System.Security.Cryptography.Algorithms.dll"
```
matthid: Can we add a similar test that checks if the rid stuff works, but now with a lower netstandard?
Because now all those changed to shouldnotcontain here so the test no longer tests that...

matthid:
Looks good. Travis failure is not related to this PR.
Merging because this fixes the build with MSBuild 15 in a particular nice way.

Issue #2040, opened 2016-11-23
Add support for setting the specific version attribute in paket.references


gautebrakstad:
### Description

Add support to set SpecificVersion on Reference in .csproject files via paket.references
As it is now you have to manually go into each Project file and update the Reference.

### Wanted result
```
<Reference Include="ccc.dll">
          <HintPath>..\..\..\packages\current\ccc\lib\net45\ccc.dll</HintPath>
          <SpecificVersion>False</SpecificVersion>
          <Private>True</Private>
          <Paket>True</Paket>
</Reference>
```

batzen:
Would also like to see this feature added.
If you develop a base library which should support version ranges for it's own dependencies it's quite a lot of work to add `<SpecificVersion>False</SpecificVersion>` every time.

batzen:
@forki I got this in my fork and already created PR #2413.

enricosada:
Closing, done by @batzen , doc in https://fsprojects.github.io/Paket/nuget-dependencies.html#Referencing-specific-versions-in-projects

Issue #1427, opened 2016-01-23
paket changes line endings and overrides FodyWeavers.xml using Costura.Fody


punker76:
Found hopefully a small issue using paket with Fody and Costura.Fody.

`paket update` always overrides my FodyWeavers.xml

``` xml
<?xml version="1.0" encoding="utf-8"?>
<Weavers>
  <Costura>
  </Costura>
</Weavers>
```

after updating the `Costura` tag is removed and the project doesn't compile.

also the line endings are changed hard to `LF` for
- FodyWeavers.xml
- FodyWeavers.xml.install.xdt
- FodyWeavers.xml.uninstall.xdt

I'm using the paket version 2.47.1.0

forki:
http://fsprojects.github.io/Paket/nuget-dependencies.html#No-content-option

did you try `content:once` ?

should be working in paket.references as well (will update docs)

forki:
http://fsprojects.github.io/Paket/references-files.html#No-content-option done

punker76:
@forki Thanks, just tested it out, works only if I do this in the `paket.dependencies` file, it has no effect with the `paket.references`.

```
source https://www.nuget.org/api/v2/
framework >= net45

nuget Costura.Fody ~> 1.3.3.0 content: once
```

forki:
Are you sure? Mhm will check later

forki:
mhm integration tests in d9aab5d
 suggest in works from the ref file as well

punker76:
@forki Just make a retest with  2.47.2.0 and all works fine now, maybe I'm doing some wrong stuff...

wallymathieu:
It looks like these tests broke when I submitted a pull request. I'm starting to look into porting the tests to avoid using xml transformation files by adding a test package (https://github.com/wallymathieu/Paket.TestPackage.Foody).
The branch I'm working on is the following:
https://github.com/wallymathieu/Paket/tree/1427_tests

derwasp:
@forki @punker76 can you please reopen this? since the issue is still there

forki:
@derwasp can you please give us a repro?

derwasp:
@forki I just tried one more time with cleaned everything and it didn't happen. I swear it was not the first time I tried that. My apologies that it looks like my repo wasn't clean or something.

However, Fody also adds an import to `.*proj` file:

```
[...]

PR #1174, opened 2015-10-28
Makes conflicts fail faster

mrinaldi:
This PR makes conflicts fail faster.

Take this `paket.dependencies` for example:
```
source https://nuget.org/api/v2

nuget Microsoft.AspNet.SignalR.JS
nuget Nancy.Serialization.JsonNet
nuget SourceLink.Fake
nuget Ninject.Extensions.Logging.Log4net == 3.2.0
nuget Ninject.Extensions.Interception.Linfu == 2.2.1.2
```

This is how `paket` is trying to resolve in 2.18.0:
```
Paket version 2.18.0.0
Resolving packages for group Main:
 - Ninject.Extensions.Logging.Log4net 3.2.0
 - Ninject.Extensions.Interception.Linfu 2.2.1.2
 - Microsoft.AspNet.SignalR.JS 2.2.0
 - Nancy.Serialization.JsonNet 1.3.0
 - SourceLink.Fake 1.1.0
 - jQuery 2.1.4
 - Nancy 1.3.0
 - Newtonsoft.Json 7.0.1
 - Ninject.Extensions.Interception 2.2.1.2
 - Ninject 2.2.1.4
 - Ninject.Extensions.Logging 3.2.3
 - Ninject.Extensions.Logging 3.2.2
 - Ninject.Extensions.Logging 3.2.1
 - Ninject.Extensions.Logging 3.2.0
 - Newtonsoft.Json 6.0.8
 - Newtonsoft.Json 6.0.7
 - Newtonsoft.Json 6.0.6
 - Newtonsoft.Json 6.0.5
 - Newtonsoft.Json 6.0.4
 - Newtonsoft.Json 6.0.3
 - Newtonsoft.Json 6.0.2
 - Newtonsoft.Json 6.0.1
 - Newtonsoft.Json 5.0.8
 - Newtonsoft.Json 5.0.7
 - Newtonsoft.Json 5.0.6
 - Newtonsoft.Json 5.0.5
 - Newtonsoft.Json 5.0.4
 - Newtonsoft.Json 5.0.3
 - Newtonsoft.Json 5.0.2
 - Newtonsoft.Json 5.0.1
 - Newtonsoft.Json 4.5.11
 - Newtonsoft.Json 4.5.10
 - Newtonsoft.Json 4.5.9
 - Newtonsoft.Json 4.5.8
 - Newtonsoft.Json 4.5.7
 - Newtonsoft.Json 4.5.6
 - Newtonsoft.Json 4.5.5
[...]

PR #2413, opened 2017-06-11
Adding support for SpecificVersion attribute on assembly references

batzen:
This PR adds support for `SpecificVersion` on assembly references.
This was requested in #2040.

- [X] Implement feature
- [X] Fix unit and integration tests (partially done, see below)
- [X] Add new tests to verify feature

I'm currently having a hard time trying to find out which integration tests i broke with my change as they often ~~fail because either the path is too long (exceeds 260 characters for file or 248 for directory)  (solved by moving to c:\paket) or they~~ fail with 

>  Expected: "...argetFrameworkVersion) == 'v4.6.1' Or $(TargetFrameworkVer..."
  But was:  "...argetFrameworkVersion) == 'v4.5' Or $(TargetFrameworkVersi..."

The tests failing with this error are:
- #1427 install content
- #1427 install content once from dependencies file
- #1427 install content once from dependencies file removes paket tag
- #1427 install content once from dependencies file stays stable
- #1427 install content once from dependencies file stays stable 2 installs
- #1427 install content once from references file
- #1427 won't install content when content:none

The following tests timeout:
- #1174 Should find Ninject error

I don't think those fails are caused by my changes, but i can't be sure...

Please tell me if i did anything wrong as this is the first time i am in contact with F#.


matthid:
can you rebase/merge? Currently master is green so it might solve the spurious issues you encountered (there was a problem with the integration test suite which was solved)...

matthid:
Nevermind you were faster :)

forki:
Is this flag enabled by default?

batzen:
The default is `True` which is the same as omitting it in assembly references.

forki:
We should not emit it then in default case to reduce impact in csproj

batzen:
I thought that i should add it to csproj as private is also added by default.
If you don't want it to be added i will remove it if it's not specified or equals true.

matthid:
@forki If we get this into paket 5 I guess it's OKisch, because it will already rewrite a lot because of new restriction system, netcore and xamarin support...

forki:
Okish yes, but still lets keep it small if possible.

Am 11.06.2017 4:13 nachm. schrieb "Matthias Dittrich" <
notifications@github.com>:

batzen:
So which way should i go?
Omit it if it's not specified (`None`) or also omit it if it's specified with the default value of `True`?

forki:
Omit only if not explicitly specified. That's what we always do.

Am 11.06.2017 4:18 nachm. schrieb "Bastian Schmidt" <
notifications@github.com>:

batzen:
Ok.
Will make the changes accordingly and revert most changes to tests.

batzen:
Should be as expected now + tests are present now too.

forki:
💋

forki:
Urgs I forgot to ask for docs and already released it in rc006. Can you please send a PR to the docs as well?

I think it needs to go to https://github.com/fsprojects/Paket/blob/master/docs/content/nuget-dependencies.md#copy_local-settings and https://github.com/fsprojects/Paket/blob/master/docs/content/references-files.md#copy_local-settings

batzen:
Of course. Didn't know about that. Will get it done till friday.

PR #2429, opened 2017-06-16
Make command line GNU-compatible

agross:
This is work in progress, hopefully to gather feedback. 

I tried to change the command line to remove some nuisances when you're used to deal with other programs that use GNU-style arguments (e.g. `paket add Castle.Windsor --group Main` instead of `paket add nuget Castle.Windsor group Main`) and align the parameter style (`--createnewbindingfiles` -> `--create-new-binding-files`). The original issue is #1974.

On top of that I plan to provide zsh completion for all Paket commands. This is currently tracked [in my dotfiles repo](https://github.com/agross/dotfiles/blob/master/paket/functions/_paket) and will be updated alongside this PR.

**Perhaps anyone experienced writing bash completions will chime in because I believe most people use bash. (e.g. Git for Windows)**

All old-style arguments are still supported (e.g. for Ionide), but not advertised in the help text. Users [will be warned](https://asciinema.org/a/124886) when using the old style with hints how to resolve.


baronfel:
This is great, and I'm looking forward to this discussion.

One note: if providing completions for many shells is going to be difficult, it may be more efficient to create bash completions and then use zsh's `bashcompinit` funtion to allow zsh to use bash completions. Documentation for this feature can be found [here](https://linux.die.net/man/1/zshcompsys)

agross:
@baronfel Thanks for the suggestion. I'm not sure bash completion is as comprehensive as the zsh completion system (never wrote one, but I've been told zsh is superior to bash). Check some of the [videos I created from the WIP](https://asciinema.org/~agross) here to see what's possible with zsh.

matthid:
> I'm not sure bash completion is as comprehensive as the zsh completion system 

Isn't this another reason to consider the suggestion of @baronfel?
I'd rather change the cli now in a way that it works well in the most limited system rather than a second time later.

But whatever we decide I'm ok with this change if this means I can get auto completion or least it is easier to implement - as I'm using git bash... (though to be completely honest I like 'paket update group blub' which you haven't touched jet?)

agross:
> Isn't this another reason to consider the suggestion of @baronfel?

Actually quite the opposite: As a zsh user I'd like best fidelity I can get. (A much better completion is one of the reasons I prefer zsh over bash.)

> I'd rather change the cli now in a way that it works well in the most limited system rather than a second time later.

The CLI is changing right now and it'll be GNU-compatible which means bash will be able to complete it.

> though to be completely honest I like 'paket update group blub' which you haven't touched jet?

I go through the commands in ascending order now, but I might willing to implement `paket update --group blubb` first if you like.

agross:
FYI, I transferred all commits from my dotfiles repo to this PR. There's also a bash completion skeleton!

forki:
Wow. How much would we break?

agross:
No breaking changes until now!

forki:
So we could do a 5.5 alpha?

agross:
I'd say not yet because there are still 6 or 7 commands that I haven't touched.

agross:
@matthid Just pushed `update` ;-)

agross:
I'm done reworking the commands. The commands look a lot more like standard commands now, even if I say so myself. There's some docs left to review (everything that's not a command i.e. not generated from the `--help` output).

Looking forward to somebody testing this beyond CI!

If a zsh and Paket user reads this, please look [no further than here](https://github.com/agross/Paket/blob/gnu/completion/README.zsh.md).

agross:
I think I'm done. While reviewing the docs I fixed some links that led to 404s. Some I could not fix, e.g. the GitHub links [from here](https://fsprojects.github.io/Paket/reference/fake-globbing.html).

You can run a local external link check using the new `check-doc-links.sh` shell script (requires node).

And the docs now have a favicon!

agross:
If Argu releases https://github.com/fsprojects/Argu/pull/84 we're all set. It's just a minor issue, though (`--from-bootstrapper` is displayed in the `--help` output despite being `Hidden`).

forki:
do we still need argu release?

agross:
The next Argu release will only remove `--from-bootstrapper` parameter from the command help outputs of subcommands (`paket add --help`). Technically it's not required, but nice for cosmetic reasons.

forki:
Awesome work! Let's see how it works in alpha!

agross:
Thank you. That means a lot, coming from you. 🎉

forki:
giphy

Released in Paket 5.5.0-alpha005

Issue #2430, opened 2017-06-16
IOException when paket.reference has File:<package> <folder> link:false


leandrosalgado:
I have in my paket.reference the below reference:

`File:angular-chart.min.js Scripts link:false`

When I run `paket install` or `paket update` I get this error:

> Paket failed with:
> -> IOException: The file '\<My Project Folder\>\Scripts\angular-chart.min.js' already exists.

### Expected behavior

I was expecting that paket would overwrite the destination file or ignore it if no changes were found.

### Actual behavior

Throw an error: 

> Paket failed with:
> -> IOException: The file '\<My Project Folder\>\Scripts\angular-chart.min.js' already exists.

forki:
Can you please create a zip with a repro? Thanks

Am 16.06.2017 17:04 schrieb "leandrosalgado" <notifications@github.com>:

matthid:
Or at least run with the environment variable `PAKET_DETAILED_ERRORS` set to `true` or the verbose option. To get a more usefull error message.

leandrosalgado:
I believe I found the fault code at [Paket.Core/InstallProcess](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Installation/InstallProcess.fs#L480):

`File.Copy(remoteFilePath,targetFile.FullName)`

My suggestion:
```fsharp
// Set the overwrite flag true
File.Copy(remoteFilePath,targetFile.FullName, true)
```

@forki I reckon with the full error log it isn't necessary anymore to create a repro, thank you.
@matthid Thank you lots for the tip to run paket verbose! 👍 

#### Stacktrace:
```powershell
FileName: angular-chart.min.js
Performance:
 - Disk IO: 1 second
 - Runtime: 20 seconds
Paket failed with:
-> IOException: The file '<Project Folder>\Scripts\angular-chart.min.js' already exists.
   StackTrace:
        at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
        at System.IO.File.InternalCopy(String sourceFileName, String destFileName, Boolean overwrite, Boolean checkHost)
        at Paket.InstallProcess.gitRemoteItems@453.Invoke(Tuple`2 tupledArg)
        at Microsoft.FSharp.Collections.IEnumerator.map@114.DoMoveNext(b& )
        at Microsoft.FSharp.Collections.IEnumerator.MapEnumerator`1.System-Collections-IEnumerator-MoveNext()
[...]

PR #2433, opened 2017-06-17
Overwrite target file when link option is false

leandrosalgado:
Set overwrite flag true when link option is false, see more details in the issue [#2430](https://github.com/fsprojects/Paket/issues/2430)


forki:
thx

PR #2477, opened 2017-06-30
Allow comments in the references file.

### LINKED ISSUE #2371: Allow comments in paket.references
### Description

Its frustrating experience if you add a comment and a long running update then fails because of

```
Paket failed with:
        Unable to parse C:\PROJ\FSharp.Formatting\src\FSharp.Formatting.Common\paket.references
```

fsoikin:
Fixes #2371.

### DIFF
--- a/src/Paket.Core/Common/Utils.fs
+++ b/src/Paket.Core/Common/Utils.fs
@@ -900,6 +900,15 @@ let removeFile (fileName : string) =
 let normalizeLineEndings (text : string) = 
     text.Replace("\r\n", "\n").Replace("\r", "\n").Replace("\n", Environment.NewLine)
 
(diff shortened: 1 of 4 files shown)


forki:
Cool!

fsoikin:
@forki note that this may become a breaking change under some very specific circumstances: before this, the `removeComment` function allowed comments to start anywhere, but now comments have to start either at the beginning of the line or after a space.

Also, I have realized that a comment can start after a space, but not after a tab. Do you think it would be a good idea to update it to allow tabs as well? Or perhaps anything that is `Char.IsWhitespace`?

forki:
Char.iswhitespace sounds good

fsoikin:
Got it. #2479

PR #2478, opened 2017-07-01
find-package-versions shouldn't require paket.dependencies to be present as long as a source is explicitly specified.

### LINKED ISSUE #2332: paket.exe find-package-versions requires paket.dependencies although it doesn't need to
### Description

Running paket.exe find-package-versions fails with message:

Paket failed with: Could not find 'paket.dependencies'....

It's by design, I understand, but only looking up packages in repository should not put any requirements on local workspace because at that point no changes are being made anywhere.

### Repro steps

1. cd out of local paket repository (e.g. cd c:\temp) or rename your paket.dependencies to something else

2. paket.exe find-package-versions nuget Paket source https://www.nuget.org/api/v2 max 1 -s

### Expected behavior

Should print the latest version of Paket, e.g. "5.0.0-alpha023".

### Actual behavior

Paket failed with:
        Could not find 'paket.dependencies'. To use Paket with this solution, please run 'paket init' first.
If you have already run 'paket.init' then ensure that 'paket.dependencies' is located in the top level directory of your repository.
Like this:
MySourceDir
  .paket
  paket.dependencies

fsoikin:
Fixes #2332.

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -31,8 +31,11 @@ type Dependencies(dependenciesFileName: string) =
         emptyDir (Constants.NuGetCacheFolder)
         emptyDir (Constants.GitRepoCacheFolder)
 
-    /// Tries to locate the paket.dependencies file in the current folder or a parent folder.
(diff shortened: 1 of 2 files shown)


forki:
thanks a lot!

PR #2482, opened 2017-07-04
Fix for usage of the new csproj with targetFramework #2450

### LINKED ISSUE #2450: Paket pack doesn't properly parse OutDir when you use $(Targetframework)
### Description

When using `TargetFrameworks`, the default value for OutDir is `./bin/$(Configuration)/$(TargetFramework)`. When packaging, paket will look for the requisite .dlls in `./bin/Release/<Dll Name>` instead of `./bin/Release/<Target Framework>/<Dll Name>`
### Repro steps

1. Create a .fsproj file that uses `<TargetFrameworks>` and set `<OutDir>` to `./bin/$(Configuration)/$(TargetFramework)/`

2. Run `paket pack`

### Expected behavior

Paket properly creates a package

### Actual behavior

Paket fails with 
```
Paket failed with:
-> FileNotFoundException: Could not find file "<Path to Project>/bin/Release/<ProjectName>.dll".
```
### Known workarounds

None

bhugot:
Hi it's a quick fix of the problem with dotnet pack on the new csproj when using template with type project.
It should fix #2450 

The issue with TargetFrameworks is still there for now. But it's a bigger rework as for now we only return one outputpath where we should return multiple one.

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/ProjectFile.fs
+++ b/src/Paket.Core/PaketConfigFiles/ProjectFile.fs
@@ -993,6 +993,10 @@ module ProjectFile =
     let getTargetFrameworkIdentifier (project:ProjectFile) = getProperty "TargetFrameworkIdentifier" project
 
     let getTargetFrameworkProfile (project:ProjectFile) = getProperty "TargetFrameworkProfile" project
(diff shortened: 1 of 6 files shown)


forki:
thanks!

PR #2483, opened 2017-07-04
find-packages shouldn't require paket.dependencies to be present

agross:
Related to #2478, same fix for `find-packages`.


forki:
thanks!

PR #2486, opened 2017-07-05
Restore cli tools via paket

forki:
/cc @enricosada @alfonsogarciacaro @davkean @dsplaisted

This PR allows `dotnet restore` to create a `obj/PROJECTNAME.fsproj.paket.props` file with the following content:

    <?xml version="1.0" encoding="utf-8" standalone="no"?>
    <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
      <PropertyGroup>
        <MSBuildAllProjects>$(MSBuildAllProjects);$(MSBuildThisFileFullPath)</MSBuildAllProjects>
      </PropertyGroup>
      <ItemGroup>
        <DotNetCliToolReference Include="dotnet-fable" Version="1.1.6" /> // or whatever cli tool we find
      </ItemGroup>
    </Project>

With this we can get rid of maintaining DotNetCliToolReference in our precious csproj/fsproj.
This became an issue since version number of dotnet-fable and Fable.Core need to match. Which involved manual work and some degree of giving a fuck.

used in https://github.com/fable-compiler/fable-suave-scaffold/pull/132

Open question:

* how do we really detect a cli tool. At the moment we check package name starts with "dotnet-"

### DIFF
--- a/src/Paket.Core/Dependencies/PackageResolver.fs
+++ b/src/Paket.Core/Dependencies/PackageResolver.fs
@@ -42,6 +42,9 @@ type ResolvedPackage = {
     member self.HasFrameworkRestrictions =
         getExplicitRestriction self.Settings.FrameworkRestrictions <> FrameworkRestriction.NoRestriction
 
+    member self.IsCliToolPackage() =
+        self.Name.ToString().StartsWith "dotnet-"
+
     member private self.Display
         with get() =
             let deps =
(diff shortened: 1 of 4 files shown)


enricosada:
> how do we really detect a cli tool. At the moment we check package name starts with "dotnet-"

It need to be explicit, trust the user 
cc @livarcocc of `dotnet/cli` team too, for feedback

In theory, there is a flag to be set in nuspec (also for discriminate the `dotnet new` templates) but is pratically unused in real world.

any `netcorepp1.0` console packaged will work afaik.
The only rule is: executable name must have prefix `dotnet-`, so `dotnet-app.dll` for `dotnet app` not the package itself, the executable
so if exists `lib\netcoreapp1.0\dotnet-{something}.dll` can be used as cli tool
Can exists more executables in same package btw

that said, lot of cli tool, doesnt use `dotnet-` prefix in package name, so need to add a conf in paket somehow.

enricosada:
uops @forki you was too fast 😄

forki:
Discussion is still open. I just merged it into alpha version so that we
can experiment downstream

Am 05.07.2017 15:32 schrieb "Enrico Sada" <notifications@github.com>:

PR #2489, opened 2017-07-05
Add default NuGet source back to find-packages

agross:
As noted by @fsoikin in https://github.com/fsprojects/Paket/commit/aa5f1b7a5f7ef050e6466b8eeac479daf144acbd#commitcomment-22938166 I unintentionally removed the default NuGet source. Thanks for the heads-up, @fsoikin!

We should probably release 5.4.6 such that Paket doesn't stay broken until 5.5-alpha is released. I apologize for the trouble!

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -308,7 +308,8 @@ let findPackages silent (results : ParseResults<_>) =
         | Some source, _ ->
             [PackageSource.NuGetV2Source source]
         | _, Some dependencies ->
-            dependencies.GetSources() |> Seq.map (fun kv -> kv.Value) |> List.concat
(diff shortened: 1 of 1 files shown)


forki:
ok released as hotfix and in alpha channel. Thx

fsoikin:
Since `find-packages` has a default source prepended, should `find-package-versions` be the same? In fact, perhaps it makes sense to factor out the common code from them both? Common sense would suggest that these two commands ought to use the same package-finding logic.

agross:
@fsoikin This sounds useful! I never considered the case, I was recently more concerned about the command line in general :)

fsoikin:
Ok, I'll send a PR

PR #2490, opened 2017-07-06
[docs] paket works with new .net sdk not just dotnet cli

enricosada:
paket works also for msbuild 15, not just dotnet cli

add command i found useful

### DIFF
--- a/docs/content/paket-and-dotnet-cli.md
+++ b/docs/content/paket-and-dotnet-cli.md
@@ -1,7 +1,8 @@
-# Paket and dotnet CLI
+# Paket and .NET Sdk (dotnet cli and msbuild 15)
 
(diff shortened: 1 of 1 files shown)


docs/content/paket-and-dotnet-cli.md
```diff
@@ -40,25 +46,46 @@ This step is absolutely the same as with traditional .NET projects.
 
 Like with traditional .NET projects you also need to put a [`paket.references` files](references-files.html) alongside your MSBuild project files.
 
+After `paket.references` files are created, to update your existing projects, just run
+
+```sh
+.paket/paket.exe install
```
enricosada: @forki that's the best command to run?

forki: part of new workflow is that you don't need to run install after a references file change. everything now happens at "dotnet restore". So you need to run that instead

PR #2494, opened 2017-07-06
docs: review, naming, linking, wrapping, formatting

agross:
Mostly related to .NET SDK. Do we need a technical writer?


forki:
💋

PR #2527, opened 2017-07-15
run paket install with paket version 5.6.12

0x53A:
This should not actually change anything.

The reason is I want to reduce the diff of my other pr.


forki:
image

is that related?

/cc @matthid

PR #2528, opened 2017-07-15
[Really Ready] dotnet sdk: disable implicitly adding system.valuetuple and fsharp.core

0x53A:
disable implicitly adding system.valuetuple and fsharp.core
fail when a sdk f# project does not contain a reference to fsharp.core

context: https://github.com/Microsoft/visualfsharp/issues/3335#issuecomment-315516168


forki:
> fail when a sdk f# project does not contain a reference to fsharp.core

I don't think we need to do that. This is basically a thing that will happen later in build. We should not have this in paket.exe itself.

0x53A:
> I don't think we need to do that. This is basically a thing that will happen later in build. We should not have this in paket.exe itself

Thats the big issue - if you dont' pass any FSharp.Core to the compiler, it will silently pick any random one.

forki:
> if you dont' pass any FSharp.Core to the compiler, it will silently pick any random one

how? I think the targets file changes forbid that!?

0x53A:
Please test it, maybe i missed something, but afaik it does not error out.

forki:
If you have a repro? Can you please report that to visualfsharp? It should not implicitly add things when you explicitly set the flag, right?

0x53A:
Will do

forki:
can you please send a second PR that only sets the properties in the targets file?

0x53A:
I don't think that is a good idea.

When you do dotnet new and paket convert, the resulting project will be broken (and still compiling, so the user doesn't notice it).

I would say let's wait for a response on that issue, maybe there is a compilerflag to disable the magic core reference?

0x53A:
Great, locally I have the same three ``PathTooLongException`` failures, but I also have one that may be more interesting:

image

matthid:
It feels like this test needs to be reworked. It's not working on travis either, see [here](https://github.com/fsprojects/Paket/pull/2530#issuecomment-315615472)

0x53A:
I need to test if it actually works (probably doesn't, I know myself =) ), and a few unit tests would also be nice, but from my side this is mostly finished.

As it stands, it adds FSharp.Core on ``convert-from-nuget`` and warns (I downgraded it from an error) on install if a v15 F# project does not reference FSharp.Core.

What do you think?

0x53A:
Manual test:

1) ``dotnet new lib -lang F#``
2) ``paket convert-from-nuget``
3) Check that both paket.dependencies and paket.references contain FSharp.Core ✔
4) manually remove FSharp.Core from paket.references
5) Check that a warning was printed ✔
image

forki:
what is the status?

0x53A:
Ready from my side.

CI errors seem unrelated, but please verify.

I've rebased it on master, which wiped the CI history, so here is a link to the previous CI result: https://ci.appveyor.com/project/SteffenForkmann/paket/build/0.0.1.9005

0x53A:
AppVeyor: PathTooLongException
Travis: each stage passed at least once ;-)
image

matthid:
Maybe we can add long-path support for the test suite on appveyor (via App.Config). We don't really care about too long paths there. As the appveyor is green on master and our only check right now I think we need it to be green here as well. Or we wait for https://github.com/fsprojects/Paket/pull/2584?

0x53A:
Kay, let me shorten the scenario names a bit ...

0x53A:
@matthid green

src/Paket.Core/PackageManagement/NugetConvert.fs
```diff
@@ -227,6 +227,22 @@ let createPackageRequirement sources (packageName, versionRange, restrictions) d
        TransitivePrereleases = false
        Graph = Set.empty }
 
+let private isFSharpproject (projectFileName:string) =
```
matthid: isFSharp**P**roject ?

0x53A:
image

If I understand it correctly, `Pack` does a `Restore`, which goes to github, and doesn't seem to be cached.

matthid:
Would be interesting to see WHAT is not cached. We had some problems with the bootstrapper in the past (where dotnet was calling the bootstrapper in magic mode).
Maybe we need to set some verbose flags:
 - `PAKET_BOOTSTRAPPER_TRACE=true` allows to set the bootstrapper into verbose mode.
 - you maybe can temporarily add "-v" to the paket targets file.

This is when this error is actually reproducible.

0x53A:
no idea - I think this may not actually be an issue with paket, because the build in my fork also failed with the same error. So it seems the AppVeyor Datacenter as a whole hit the limit.

forki:
thanks!

PR #2544, opened 2017-07-19
Fix1925: improve paket.local docs wrt versions and restore command sp…

daz10000:
This is a suggested doc fix for Issue 1925.   The changes highlight the default behavior for paket.local files with respect to requiring exact version matches,  that only the restore command is affected, and a suggested use case is provided in the package version override section for testing upgraded versions of packages with an example.   I hope it addresses the areas I found a little confusing the first time I used the paket.local file.

### DIFF
--- a/docs/content/local-file.md
+++ b/docs/content/local-file.md
@@ -3,9 +3,11 @@
 The `paket.local` file is used for testing changes before releasing new version
 of a dependent project.
 
-**Note:** You *should not commit* `paket.local`. This file should always be
-added to `.gitignore`, because it specifies file paths that are only accessible
-on your machine.
+> Notes: 
+>
+> * You **should not commit** `paket.local`. This file should always be added to `.gitignore`, because it specifies file paths that are only accessible on your machine.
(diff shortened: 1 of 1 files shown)


docs/content/local-file.md
```diff
@@ -132,6 +134,23 @@ is determined by your build server, and `.nupkg`s created locally have zero
...
+
+```paket
+    // Argu is searched in specific version, grab a specific version
+    nuget Argu -> source C:\github\Argu\bin version 1.0.0
+```
+
+>Notes
+>
+>* `version` is currently supported only for [Source override](#Source-override).
+>* The syntax allows only a specific version, **not** inequality expressions like >= .
```
theimowski: Instead `inequality expressions like >=` let's have an already used term for that: `version constraint` http://fsprojects.github.io/Paket/nuget-dependencies.html#Version-constraints

theimowski:
Added a comment - apart from that LGTM

daz10000:
Thanks guys - have changed the term to version constraint and included the link, tested and updated PR.

forki:
thanks!

PR #2553, opened 2017-07-21
Update FSharp.Core

forki:


forki:
ok hanging test is: should fallback to timeoutexception when task never canceles

It can be reproduced in VS.
@matthid it's a lot of Task cancellation stuff going on. I assume we need to extract that from the resolver and submit as test case to VF#. /cc @eiriktsarpalis @dsyme

PR #2562, opened 2017-07-26
Fix endsWithIgnoreCase

0x53A:
if text is smaller than suffix, -1 (not found) passes the condition

----

Note: I have not verified if the changed test is actually correct, or not!


0x53A:
Is this a bug in the Mono allocator, or "just" an out-of-memory exception?

image

matthid:
@0x53A To be honest I stopped looking at travis, all stuff like this which prints a native stack is categorized as mono bug (by me) and I'm working on getting rid of mono completely to make travis useful again...

PR #2577, opened 2017-08-01
Do not gc if we have a project file

forki:


forki:
not really sure how it got there. especially with that framework <> None filter

PR #2587, opened 2017-08-04
[ignore this] no-op

0x53A:
I have local test failures on __master__.
Let's see what CI says.

image


0x53A:
bah, one git clean later the tests are green again `` ¯\_(ツ)_/¯ ``

Issue #2599, opened 2017-08-10
Issue with DevExpress nuget feed


jcchalte:
Hi,

We have a project that use their products that we switched to paket three days ago, as a test before migrating our whole codebase. The migration was a blast, paket is VERY promising for our needs.

But we are encountering an issue with one of our supplier Nuget feed : DevExpress. DevExpress is a company that sells UI components for web & desktop applications. They usually package their components through an installer, but they started to distribute them through a private nuget server a few months ago. 

[https://www.devexpress.com/Support/Center/Question/Details/T466415/devexpress-nuget-packages](https://www.devexpress.com/Support/Center/Question/Details/T466415/devexpress-nuget-packages)

As we switched our package management system, we also tried to switch to the DevExpress nuget repository, but we cannot find a way to install the packages, as Paket seems to have issues with version numbers.

My paket.depencies file is as follows :

```
framework: auto-detect
redirects: on

source .\CustomNugetRepository
nuget EntityFramework = 6.2.0-alpha

source https://api.nuget.org/v3/index.json
nuget bootstrap 
...much more dependencies...
nuget WindowsAzure.Storage

source https://nuget.devexpress.com/MyPrivateKey/api
nuget DevExpress.Web 17.1.5.0
```

If I specify the DevExpress.Web latest version (17.1.5.0, or even `== 17.1.5.0`), `paket install` fails because it tries to resolve `DevExpress.Core 17.1.5` which does not exists on the server (the available version is '17.1.5.0'). If I specify an older version (`17.1.4.0`), I have another weird issue, paket does not seems to recognize one available version on the server

```
There was a version conflict during package resolution.
     Conflict detected:
      - Dependencies file requested package DevExpress.Web: == 17.1.4.0
      - Available versions:
        - (17.1.5.0, [https://nuget.devexpress.com/MyPrivateKey/api])
        - (17.1.4.0, [https://nuget.devexpress.com/MyPrivateKey/api])
        - (17.1.3.0, [https://nuget.devexpress.com/MyPrivateKey/api])
       ... And much more...
     Please try to relax some conditions or resolve the conflict manually (see http://fsprojects.github.io/Paket/nuget-dependencies.html#Use-exactly-this-version-constraint).
```

It might be an issue with the nuget server, but couldn't it be an issue with the way Paket manages version numbers with 4 parts ?

_I can obviously not expose publicly my private nuget key, but if this is required, I can send it privately._

jcchalte:
[...]

PR #2600, opened 2017-08-10
#2599 : Package detail request fallback when no package detail in main feed.

jcchalte:
NuGetV2 :
- Corrected an edge case where the Nuget Feed does not expose the package detail in the root feed but let the user access details directly through the detail URL.
Should correct #2599

### DIFF
--- a/src/Paket.Core/Dependencies/NuGetV2.fs
+++ b/src/Paket.Core/Dependencies/NuGetV2.fs
@@ -243,6 +243,34 @@ let parseODataEntryDetails (url,nugetURL,packageName:PackageName,version:SemVerI
 let getDetailsFromNuGetViaODataFast auth nugetURL (packageName:PackageName) (version:SemVerInfo) =
     async {
         let normalizedVersion = version.Normalize()
+        let fallback6 () =
+            async {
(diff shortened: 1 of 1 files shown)


forki:
(don't worry if the Travis goes red. We currently experience a bit of mono trouble.)

jcchalte:
Well, I have looked at the codebase for another  "Packages(Id=" and you are right, this case is already used but only on URL that appears to be tfs or vsts related (eg. if `urlSimilarToTfsOrVsts` function returns true).

So we could either change this test to include more cases such as the devexpress case (but I can't see any way to do this just from the URL without adding one more hard-coded test, and nothing in the feed seems really relevant at first sight) or maybe remove the TfsOrTst branch and let those queries bubble up to the 5 ou 6th fallback function ? But it will slow down those queries....

forki:
I will think about it and discuss it with @matthid. I think we will release
your pr as hotfix and then try to merge @matthid's perf improvements.

Sorry @matthid - this will need another rebase then. :P

But I assume we can then remove that other code path completely. Because
perf comes back with @matthid's tricks

Am 11.08.2017 09:05 schrieb "Jean-Christophe Chalté" <
notifications@github.com>:

matthid:
Can you all please retry with  5.86.0-alpha002. Otherwise we might break you on next release.

PR #2610, opened 2017-08-13
Friendlier warnings about obsolete syntax

TeaDrivenDev:
When using the "old" syntax for Paket commands, such as `paket add nuget <package>`, Paket shows a warning saying
> Please use the new syntax: nuget -> (omit, option is the new default argument)

or
> Please use the new syntax: project -> --project

These are concise but not very friendly, and the first one requires some interpretation to make sense of.

This PR makes these messages more human-readable:
> Please use the new syntax: 'nuget' is the default argument and should be omitted.

and

> Please use the new syntax: 'project' has been replaced by '--project'.

Also, the error message when using old and new syntax together has been changed from

> You cannot use the old and new syntax at the same time: project <-> --project

to
> You cannot use the old and new syntax at the same time: 'project' has been replaced by '--project'. You cannot mix the two.

### DIFF
--- a/src/Paket/Program.fs
+++ b/src/Paket/Program.fs
@@ -104,25 +104,36 @@ let processWithValidation silent validateF commandF (result : ParseResults<'T>) 
 let processCommand silent commandF result =
     processWithValidation silent (fun _ -> true) commandF result
 
-let warnObsolete o n =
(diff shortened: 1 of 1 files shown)


forki:
thx!

PR #2614, opened 2017-08-15
Make sure we don't forget empty results

forki:


forki:
for nuget and myget we should break earlier if we get Emptyesult. we don't need to check all fallbacks here

matthid:
Ideally we trust empty results when we got real results from a particular url before...

forki:
Yes. That is next step. We should definitely trust myget and nuget

Am 15.08.2017 13:52 schrieb "Matthias Dittrich" <notifications@github.com>:

matthid:
We should do https://github.com/fsprojects/Paket/pull/2582 for GetDetails as well instead of adding special cases again :/

forki:
Yes definitely.

Am 15.08.2017 13:54 schrieb "Matthias Dittrich" <notifications@github.com>:

PR #2625, opened 2017-08-17
Fixes mac install script by switching it to https for nuget

loganmac:
I was trying to install on Mac, and ran into the issue saying it couldn't connect to "http://www.nuget.org".  After changing it to use HTTPS, the script worked.


forki:
Sounds good. Thx

PR #2637, opened 2017-08-20
Mono boringssl: retry up to 10 times / show *which* path is too long, if _FileInfo_ fails

0x53A:
This should improve the travis test results

both commits were extracted from my other PRs


0x53A:
try 1: 4/6 green
image

Failures:
image

image

matthid:
Nice try, but it seems like mono just doesn't like paket

0x53A:
The nightly is a lot better. With latest / 5.2, there are assertions and crashes in malloc, etc, all over the place, but with nightly, I could only find issues with boringssl / "no performance data received".

0x53A:
Currently, Travis easily gets a long backlog:

image

I propose to drop 5.2, this would cut the build time by 1/3

matthid:
Kudos for not giving up :P

0x53A:
It still failed with that boring-ssl bug, but I couldn't see whether my check was incorrect, or whether it really failed 10 times in a row.

That's why I added a throw with a custom message, but now it doesn't fail anymore at all :\

Issue #2653, opened 2017-08-23
NewSdk: PackageReference Version may not be fixed


0x53A:
I have not tested the following theory. But I wanted to create an issue before I forget about this again ;)

In our target we add the PackageReference in the form
``<PackageReference Include="ExamplePackage" Version="6.1" />``.

https://github.com/fsprojects/Paket/blob/b7e31e3c798b72cfca0ab062605103e136eaa83a/src/Paket/embedded/Paket.Restore.targets#L40-L42

But according to nuget, this means any version 6.1 __or up__: https://docs.microsoft.com/en-us/nuget/reference/package-versioning.

Shouldn't we instead use ``[%(PaketReferencesFileLinesInfo.PackageVersion)]``?

Edit:

I realized that this is a non-issue, because nuget will prefer the lowest matching, apparently:

https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files

> In the example above, 3.6.0 means any version that is >=3.6.0 with preference for the lowest version, as described on Package versioning.

PR #2654, opened 2017-08-23
Paket.Restore.targets: use fixed version ([1.0])

0x53A:
fixes https://github.com/fsprojects/Paket/issues/2653

When the version is specified without the braces, then it will choose any version equal or higher, with preference for the lowest.


0x53A:
I don't think the previous behavior would have caused any issues, but safe is safe ...

The only case I can think of is if the nuget cache for some reason didn't contain the exact version (paket restore should have put it there), but a higher one. Then nuget would have silently chosen the higher one.

0x53A:
Aaand github is down again?

image

image

forki:
thx!

0x53A:
damn, sorry, can you please revert this?
This changes the behavior of dotnet pack.

Before it would generate a constraint >=, now it generates a constraint ==.

I don't think it is possible to tell nuget to use (==) for resolution, but use (>=) for the nuspec.
Since paket restore puts the exact version into the cache anyway, I don't think there is any realistic chance for something to go wrong.

PR #2664, opened 2017-08-25
[Ready] Make paket build on a clean windows

0x53A:
### Windows

* Install Windows 10 Pro from ISO
* install git
* clone paket
* run build.cmd

image
(Note: IntegrationTests were skipped in this screenshot (because they take so much time), but they were also green!

### Linux

We still rely on a global mono and a global MSBuild, because the bundled version is not compatible with linux and crashes hard on startup.

__Notes:__
* msbuild is provided by 'RoslynTools.MSBuild'
* ~ReferenceAssemblies are currently checked-in. I need to either find a nuget package, or create one myself. (I had planned to use 'RoslynTools.ReferenceAssemblies', but that doesn't contain the ones for 4.5)~ I created a [nuget package](https://www.nuget.org/packages/0x53A.ReferenceAssemblies.Paket/) using https://github.com/jaredpar/xcopy-msbuild. This contains only the two directories it needs (NETFramework\v4.5 to compile against net45 AND .NETPortable\v4.5 to ___consume___ portable assemblies, for example FsCheck)
* ~MSBuild doesn't seem to support Directory.build.props in combination with old-sdk projects? I had wanted to use that to set ``FrameworkPathOverride``, but it didn't work until I manually modified all fsprojs~ Only the first Directory.Build.props is used, and there was already one in /src/. Also, Directory.build.targets does not yet work: https://github.com/Microsoft/msbuild/issues/1721
* ~there is something funky with referencing portable stuff, I had to disable the FsCheck tests, or I would get the following errors:~ fixed after adding ref-assembly directory ``.NETPortable\v4.5``

```

### CODE REVIEW

**Comment by 0x53A** [comment]
what binaries do you mean? ;D

File: paket.dependencies [hunk 1]
```diff
@@ -44,3 +44,9 @@ group Build
   nuget ILRepack
   
   github fsharp/FAKE modules/Octokit/Octokit.fsx
+  
+  nuget 0x53A.ReferenceAssemblies.net45 0.1.0
```
matthid: So this is for net45 support, correct? because in the future this is no longer required as soon as we are fine with net461 compat?

0x53A: No, these are the reference assemblies. When we move to net461 we need to update to the correct reference assemblies for that.

The reference assemblies are the "contract assemblies" that the compiler compiles against. They are metadata only.

I used https://github.com/jaredpar/xcopy-msbuild, this basically just zipped up the content of ``C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5``

matthid: But once we are on net461 all we need is the netstandard build, so we could drop that in theory?
I don't think this is the correct thing to do long term (though I haven't really decided yet).

I thought we will have all the reference assemblies as "official" nuget packages now?

0x53A: I haven't found _any_ reference assemblies as nuget packages.

0x53A: Otherwise I'm pretty sure the roslyn build would use these, instead of packaging this stuff up themselves.

0x53A: > But once we are on net461 all we need is the netstandard build

at the least paket.exe needs to compile against a "real" netfx version, so we still need the ref assemblies. paket.core _could_ be compiled against netstandard, but why bother?

matthid: Hm, that's what I'm doing with FAKE 5, I don't depend on anything for compilation but the reference assemblies in the nuget packages (`ref` folder).

But I don't know if they are available for net45 as well, but I assume they are. Just look in the netstandard16 packages or the Netstandard.Library2.0 package

matthid: > why bother?

I just thought there is a "better" way by using official infrastructure. Might be wrong ...

0x53A: What I meant is, why do 

``Paket (net461) -> Paket.Core (netstandard1.6)``

instead of

``Paket (net461) -> Paket.Core (net461)``

?

The second way surely causes less issues.

0x53A: > Just look in the netstandard16 packages or the Netstandard.Library2.0 package

I haven't looked, but I suspect these would be the reference assemblies for netstandard.

At the end of the day, paket.exe needs to target full-fx in some version, so we need the reference assemblies for full-fx.

matthid: For example https://www.nuget.org/packages/System.Linq/ contains a reference assembly in the ref/net463 folder.

But I have no idea where they went for sdk20

matthid: To back off a step: My understanding was that you can download the SDK and build everything - not only netcore app but also full framework applications. Is there really a requirement to install Visual Studio to get reference assemblies?

0x53A: These are all just ref assemblies for the implementation assemblies.

For example, in System.Linq: it contains neither ref nor implementation assembly for net45 (and both for net463).

We need the ref-assemblies for the assemblies of the net framework. Any nuget package would just contain a large list of ref assemblies without any implementation assemblies.

image

0x53A: > My understanding was that you can download the SDK and build everything - not only netcore app but also full framework applications. Is there really a requirement to install Visual Studio to get reference assemblies?

No, but you DO need to install the targeting packs

matthid: Yeah I get that and like that. Can you elaborate/link the RoslynTools.ReferenceAssemblies you are referring to above (I cannot find it)? And also why it doesn't work for us (just trying to figure out if we can make it work or if it is viable for other scenarios)?

Thanks for your work on this and being patient until I get it ;)

0x53A: > Can you elaborate/link the RoslynTools.ReferenceAssemblies you are referring to above (I cannot find it)?

source https://dotnet.myget.org/F/roslyn-tools/api/v3/index.json

> And also why it doesn't work for us (just trying to figure out if we can make it work or if it is viable for other scenarios)?

It does not contain the ref-assemblies for 4.5:

image

0x53A: Hm, maybe I can use this?
image
Will try it ...

matthid: Maybe should just upgrade to net46? I'm not sure about the portable stuff could be a subset..

0x53A: So, I did a test, and I am pretty sure we need .NETFramework\v4.5 to compile against net45 AND .NETPortable\v4.5 to ___consume___ portable assemblies, for example FsCheck.

0x53A: I have created a package that contains only these two. Reason: it is smaller.

0x53A:
The tool ``dotnet-mergenupkg`` was compiled against netcore1.0, so we need the netcore1.0.x runtime to execute it ...

The issue is that Fake only allows the installation of a single runtime, I tried installing two after each other but the second deleted the first.

0x53A:
Hi @enricosada, do you think it would be possible to update [dotnet-mergenupkg](https://www.nuget.org/packages/dotnet-mergenupkg/) and provide a netcoreapp2.0 binary?

Otherwise usage in the Sdk 2.x is a bit clumsy, because I still need to install the 1.0.x runtime in parallel somehow.

0x53A:
image

paket.dependencies
```diff
@@ -44,3 +44,10 @@ group Build
   nuget ILRepack
   
   github fsharp/FAKE modules/Octokit/Octokit.fsx
+  
+  nuget 0x53A.ReferenceAssemblies.Paket 0.2.0
```
matthid: So every project now bundles their own reference assemblies :P

matthid: Or put differently, if you think this is a good way forward (which I think it is). We might want to have another documented project on how this stuff works (Then others can use it and read about it). Anyway that's just a suggestion (I'd like to maybe add that to FAKE as well)

build.cmd
```diff
@@ -5,4 +5,7 @@ if errorlevel 1 (
   exit /b %errorlevel%
 )
 
+set MSBUILD=packages\build\RoslynTools.MSBuild\tools\msbuild
```
matthid: Maybe we could set that in the FAKE script.

0x53A: sure, that may also fix unix, which currently fails (because it still runs the bundled msbuild).

How would you set it inside Fake? Just set the env vars? or is there some config switch?

paket.dependencies
0x53A: Yeah, I would prefer to use an official package, also.

The one advantage this has is that it only bundles the directories it actually needs, so the size is a bit smaller. I _could_ have given it a specific name like "ReferenceAssemblies.Net45", but this way it is clear that it is used by paket and will change as soon as we update.

Building the package is dead easy, clone https://github.com/jaredpar/xcopy-msbuild, modify which frameworks to include https://github.com/jaredpar/xcopy-msbuild/blob/006e1bb65d031854f38eaac251b5f799ce281e37/build-reference-assemblies.ps1#L23-L30
and run the powershell script.

build.cmd
matthid: Good question. I think in FAKE 4 the way would be to change the mutable field with the path.
But yeah I don't know how a good FAKE API for this would look like...

0x53A:
what a bunch of crap:

image

matthid:
Oh yeah ;)

0x53A:
well, one issue was that ubuntu 17.x still references some ancient 4.x mono ...

0x53A:
does paket not use caching on unix?

Maybe I'm wrong ,but it seems to download everything every time

0x53A:
wait - paket now finished successfully???????

I only had to start the build script 20 times!

forki:
Mono? It's short for "Mondays..."

Am 26.08.2017 6:12 nachm. schrieb "Matthias Dittrich" <
notifications@github.com>:

matthid:
Well, if we don't find a global cache we fallback to some local directory. Maybe that is what you are seeing?

0x53A:
Idk, I was probably confused. Each time I started it, it downloaded some, then failed. But each time it probably got a little bit further...

You may want to mute this thread, i will probably rage a bit until I either have this fixed, or given up ;)

forki:
Some of the integration tests also run -f

Am 26.08.2017 6:21 nachm. schrieb "Lukas Rieger" <notifications@github.com>:

0x53A:
good news: I got it to use the msbuild from the package.

bad news: it crashes hard on startup

image

0x53A:
Ready from my side, as soon as AppVeyor and at least one travis leg are green

paket.dependencies
matthid: Maybe we should add that as comment then?

src/Directory.Build.props
```diff
@@ -1,9 +1,12 @@
 <Project>
+  <!-- MSBuild only includes the first Directory.Build.props, so we need to manually include the root one -->
+  <Import Project="$(MSBuildThisFileDirectory)..\Directory.Build.props" />
```
matthid: Maybe `/` like above?

0x53A: msbuild does a normalization, the relevant change was the casing. I think I will go back to backslash for everything

0x53A:
> I'd love to see a bit less "WIP"-Dependencies but overall that's a good step forward.

The version numbers are arbitrary. If it makes you feel better I can push a 1.0 right now ;D

If you want, we can also change to a "non-branded" package that is owned by you guys or the fsfoundation.

"0x53A.ReferenceAssemblies.Paket" was just so that I could get started.

Similar, I forked ``dotnet-mergenupkg`` to ``dotnet-0x53A-mergenupk``g so that I could work on this _now_, we can wait with actually pulling this until enrico has updated his original package to netcoreapp2.0.

src/Directory.Build.props
0x53A: > Maybe we should add that as comment then?

I had a comment in the actual commit, but yes, will add a source comment

matthid:
Note that it's not a blocker for me :). I trust you to cleanup eventually. That's why approved the changes ;)

0x53A:
There wasn't any real need for this PR, I mostly wanted to test this out so I can also move my work solution to this model. So it is up to you whether we pull it now and clean up later, or wait a bit.

forki:
> The version numbers are arbitrary

Welcome to netcore ecosystem ;-)

build.fsx
```diff
@@ -364,7 +364,7 @@ Target "MergeDotnetCoreIntoNuget" (fun _ ->
 
     let runTool = runCmdIn "tools" dotnetExePath
 
-    runTool """mergenupkg --source "%s" --other "%s" --framework netstandard1.6 """ nupkg netcoreNupkg
+    runTool """0x53A-mergenupkg --source "%s" --other "%s" --framework netstandard1.6 """ nupkg netcoreNupkg
```
matthid: I think this fixes https://github.com/fsprojects/Paket/issues/2645. This is bad. We didn't even realize that we still depend on SDK1.0 to be installed. Thanks :)

0x53A: You are welcome ;)

0x53A:
image

0x53A:
Note: I would like to add an actual CI leg for this, but afaik AppVeyor does not provide "clean" windows machines, so I don't know of a way to test this.

matthid:
Maybe with travis a minimalistic netcore only build?

0x53A:
> Maybe with travis a minimalistic netcore only build?

On linux, I still rely on the system-installed msbuild because the one from RoslynTools.MSBuild crashes hard

matthid:
"dotnet msbuild" could work. I mean there is a bit of work to do to make this happen but it's not too unrealistic as paket/fake are basically ported. I think the only thing I know is that the netcore version of paket cannot push for some reason

0x53A:
Officially, the dotnet cli does NOT support build old-sdk projects.

matthid:
then we need to update sooner or later anyway, because I really want to get rid of mono. We can still have tests and stuff working with mono but the core stuff will probably move soon :)

0x53A:
yes, it would be great to move paket itself to netcoreapp, maybe even standalone so the dotnet runtime does not need to be installed.

Do you know what the status of the linker is? Single-file standalone would be awesome.

matthid:
I would be so happy already if stuff would be working, so I didn't think about a linker yet ;)

But yeah it would be awesome.

forki:
AWESOME!!!!

Issue #2733, opened 2017-09-06
Paket restore large HTTP dependencies 


sergey-tihon:
### Description

I cannot restore large http dependencies (looks like 100s timeout) 
for https://github.com/sergey-tihon/Stanford.NLP.NET/issues/74

### Repro steps

run `mono .paket/paket.exe install`

```
http http://www.frijters.net/ikvmbin-8.1.5717.0.zip
http http://nlp.stanford.edu/software/stanford-corenlp-full-2017-06-09.zip
http https://nlp.stanford.edu/software/stanford-ner-2017-06-09.zip
http https://nlp.stanford.edu/software/stanford-parser-full-2017-06-09.zip
http https://nlp.stanford.edu/software/stanford-postagger-full-2017-06-09.zip
http https://nlp.stanford.edu/software/stanford-segmenter-2017-06-09.zip
```

### Expected behavior

zip archives downloaded and unzipped.

### Actual behavior

![2017-09-06_2220](https://user-images.githubusercontent.com/1197905/30130407-ab9342d8-9351-11e7-86e2-b8bf5ae97e5b.png)

### Known workarounds

No =( Maybe try to find some old version

sergey-tihon:
This looks like default timeout of HttpClient https://msdn.microsoft.com/en-us/library/system.net.http.httpclient.timeout%28v=vs.110%29.aspx?f=255&MSPPError=-2147217396

sergey-tihon:
fixed in 5.92.8

PR #2737, opened 2017-09-07
Fix for #2733

sergey-tihon:
remove default 100sec timeout for Http dependencies download #2733


src/Paket.Core/Dependencies/RemoteDownload.fs
```diff
@@ -266,11 +266,13 @@ let downloadRemoteFiles(remoteFile:ResolvedSourceFile,destination) = async {
     | Origin.HttpLink(origin), _ ->
         let url = origin + remoteFile.Commit
         let authentication = auth remoteFile.AuthKey url
+        let timeout = (Some(TimeSpan.FromDays(1.0)))
```
forki: o_O

sergey-tihon: It is kind of `Infinity`, sometimes internet connection is really pure... =) I can spend `20min` just to download `8Mb` of new `paket.exe` from Github. Do you have better default for `infinity` timeout?

PR #2760, opened 2017-09-13
Resolves #2719 Print package version in "paket why"

### LINKED ISSUE #2719: Print package version in "paket why"
We often ask people to use "paket why" to find out where .NETStandard.Library comes from.
Like @haf did, users often report this back to us. It would be very helpful (even for ourselves) if that output would also print the package versions ;-)

t-krag:
paket-why


forki:
thanks a lot!

Issue #1848, opened 2016-08-04
include-referenced-projects lead to strange changes in nuspec


lexarchik:
I have two projects ProjA, ProjB. 
Both have paket.template files. 
ProjA have reference to ProjB (in .csproj file)

I execute `paket pack output test` in solution folder 
In result package ProjA have a dependence with min version specified to package ProjB.

But if i execute `paket pack output test include-referenced-projects`
result package ProjA have a dependence without specified version to package ProjB
(`pin-project-references` do not change situation in this case)

Expected behavior: ProjA have a dependence with min version to ProjB (or with fixed version if `pin-project-references` was specified)

Sample solution in attached archive
[PackIncludeRefProj.zip](https://github.com/fsprojects/Paket/files/402862/PackIncludeRefProj.zip)

jam40jeff:
@forki This behavior existed before the introduction of `pin-project-references`.  I am not too familiar with `include-referenced-projects` as I have not used it, but it seems to have always worked this way (which doesn't seem correct).  However, I am not sure I see much benefit to this command line option, as including a `paket.template` file for each project in the solution is a one-time cost which also provides other benefits.

PR #2766, opened 2017-09-14
fixes #2765

### LINKED ISSUE #2765: Paket pack seems to evaluate all templates it finds even when using the --template switch
### Description

When running `paket pack` with the `--template` switch it seems that paket still goes looking at other `paket.template` files than the one provided. This is all fine as long as all other paket.template files it can find are ok, but if for some reason they arent (in this case because of only partially compiling the solution) paket still fails even though the packing of the template file given by the `--template` switch would work fine. 

### Repro steps

Check out the following reproduction repo: https://github.com/wastaz/paket-template-repro
Run `./build.sh targetHost=ProjectA`

### Expected behavior

This project contains two projects, ProjectA and ProjectB, each of these has its own paket.template file (and is meant to be pushed as two separate nugets). They have no references to each other.
By running `./build.sh targetHost=ProjectA` it should build only ProjectA and then pack only ProjectA. So I expect this to correctly produce a ProjectA nuget package.

### Actual behavior

However paket fails with the following message.
I've also tried running this without FAKE, and I get the exact same output and problem.

```
Starting Target: ProGetPack (==> Build)
C:\oss-code\PaketTemplateRepro\.paket\paket.exe pack output "../build"  --version  0.1.0.0 --template  ./ProjectA/paket.template
Paket version 5.96.2
Please use the new syntax: 'output' is the default argument and should be omitted.
Performance:
 - Runtime: 923 milliseconds
Paket failed with
-> DirectoryNotFoundException: Could not find a part of the path 'C:\oss-code\PaketTemplateRepro\ProjectB\bin\Release\ProjectB.dll'.
Running build failed.
Error:
System.Exception: Error during packing ..
   at Fake.Paket.Pack@128-4.Invoke(String message) in D:\code\FAKE\src\app\FakeLib\PaketHelper.fs:line 128
   at Fake.Paket.Pack(FSharpFunc`2 setParams) in D:\code\FAKE\src\app\FakeLib\PaketHelper.fs:line 128
   at FSI_0005.Build.clo@54-8.Invoke(Unit _arg4)
   at Fake.TargetHelper.runSingleTarget(TargetTemplate`1 target) in D:\code\FAKE\src\app\FakeLib\TargetHelper.fs:line 626
```

Notice how what it is complaining about is that it could not find ProjectB.dll, which is of course correct since we have not compiled ProjectB and we do not need ProjectB in order to pack ProjectA.

### Known workarounds

Always compile everything or remove all other paket.template files...

forki:


lexarchik:
Tests will fail due to `.template` files in integtration tests was deleted in 02150c58f064d4c329f30818f02ecaf4674eefcb.
I choose wrong name for files in integration tests in #2753. It seems like `.template`  must be `.templatetemplate`.
But i can't see anything related in .gitignore :(

lexarchik:
And this fix will break #1848 for `pack --template templatename --include-referenced-projects` case.
Paket can't choose "referenced-projects" dependencies correctly without analysis of that projects.

forki:
@lexarchik ok I'm closing this one. Any chance you can give it a try to fix as well?

lexarchik:
Yes, but which way?) (https://github.com/fsprojects/Paket/issues/2765#issuecomment-329487095).
Third option is to revert part of my fix and leave `--include-referenced-projects` broken for `pack --template` case.
We don't use such case but i don't want to leave issue half-resolved.

PR #2773, opened 2017-09-17
[Build] dotnet-0x53A-mergenupkg => dotnet-mergenupkg

0x53A:
ref https://github.com/fsprojects/Paket/pull/2664#issuecomment-326571495


forki:
Thx

PR #2781, opened 2017-09-20
Adding feature to verify the URL and credential correctness before storing them in paket.config

ruhullahshah:
While adding a source and its credentials to paket.config, it often happens that there are typos or the credentials do not authenticate, in that case adding these credentials and sources to paket.config becomes pointless. So we propose to have --verify option to augment the config command. It verifies the URL for the source and the credentials are correct.

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -165,11 +165,17 @@ type Dependencies(dependenciesFileName: string) =
                                               projectName, installAfter, runResolver))
 
     /// Adds credentials for a Nuget feed
+    member this.AddCredentials(source: string, username: string, password : string, authType : string) : unit =
+        RunInLockedAccessMode(
+            this.RootPath,
+            fun () -> ConfigFile.askAndAddAuth source username password authType false |> returnOrFail )
+
+     /// Adds credentials for a Nuget feed
     member this.AddCredentials(source: string, username: string, password : string, authType : string, verify : bool) : unit =
         RunInLockedAccessMode(
             this.RootPath,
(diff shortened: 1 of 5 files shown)


src/Paket.Core/PublicAPI.fs
```diff
@@ -165,10 +165,10 @@ type Dependencies(dependenciesFileName: string) =
                                               projectName, installAfter, runResolver))
 
     /// Adds credentials for a Nuget feed
-    member this.AddCredentials(source: string, username: string, password : string, authType : string) : unit =
+    member this.AddCredentials(source: string, username: string, password : string, authType : string, verify : bool) : unit =
```
forki: this one is in the PublicAPI and would be breaking.
We need a second function and keep the old one

src/Paket.Core/Versioning/ConfigFile.fs
```diff
@@ -229,4 +229,15 @@ let askAndAddAuth (source : string) (username : string) (password : string) (aut
             if input = "" then "basic" else input
         else
             authType
-    AddCredentials (source.TrimEnd [|'/'|], username, password, authType)
+
+    let authResult = 
+        if verify then 
+            Console.WriteLine "Verifying the source URL and credentials..."
```
forki: should use tracef

ruhullahshah:
Thanks. I will update the code with the requested changes.

src/Paket.Core/PublicAPI.fs
ruhullahshah: Done

forki:
thx!

ruhullahshah:
You are welcome and thanks for the quick review, subsequent merge

PR #2813, opened 2017-10-02
Remove dnxcore50 moniker from lock file

### LINKED ISSUE #2810: Remove dnxcore50 moniker from lock file
That thing leaks everywhere and is not used anymore.

forki:
fixes #2810


src/Paket.Core/Dependencies/NuGetCache.fs
```diff
@@ -176,7 +179,8 @@ let tryGetDetailsFromCache force nugetURL (packageName:PackageName) (version:Sem
                 if (PackageName cachedObject.PackageName <> packageName) ||
                     (cachedObject.Version <> version.Normalize())
                 then
-                    traceVerbose (sprintf "Invalidating Cache '%s:%s' <> '%s:%s'" cachedObject.PackageName cachedObject.Version packageName.Name (version.Normalize()))
+                    if verbose then
```
matthid: macht den check nicht auch traceVerbose schon?

forki: yes, but the right side (sprintf ...) would be evaluated even if verbose is false.
This was serious perf problem in the past

forki:
if we move it into extractPlatforms then we get the annoying warnings that dnxcore is not real moniker

src/Paket.Core/Dependencies/NuGetCache.fs
matthid: Should we remove `traceVerbose` in that case ?

forki:
I did that a while ago, but I'm not remembering  why it didn't work fully

Am 02.10.2017 12:21 schrieb "Matthias Dittrich" <notifications@github.com>:

matthid:
Tests need to be adapted. The current failure looks a bit suspicious, but it might be that the error message is just incomplete and everything is fine.

forki:
Working on the tests. Will come later today

Am 02.10.2017 12:24 schrieb "Matthias Dittrich" <notifications@github.com>:

PR #2816, opened 2017-10-02
Stricter parsing in Requirements.parseRestrictionsLegacy

### LINKED ISSUE #2803: parse error in framework restrictions
deps file with:

    framework: netstandard 2.0

leads to lock file with:

    RESTRICTION: == net2

which is clearly wrong. The parser should throw an error instead.

Correct would be:

    framework: netstandard2

csmager:
This more strictly requires expressions matching the syntax in the examples below:

| Value | Result |
| --- | --- |
| net45 | Exactly net45 |
| = net45 | Exactly net45 |
| >= net45 | At least net45 |
| >= net45 <= net46 | Between net45 and net46 |
| portable-windows8+net45+wp8 | At least Profile78 |
| >= portable-windows8+net45+wp8 | At least Profile78 |

Note that due to expectations in lock file parsing tests like [this one](https://github.com/fsprojects/Paket/blob/8a7070f055c2366bd138b2c0d0dbcdeb9c57277c/tests/Paket.Tests/Lockfile/ParserSpecs.fs#L282-L339), we also accept `>= net45 < net46` as meaning 'Between net45 and net46'.

Fixes #2803

### DIFF
--- a/src/Paket.Core/PackageManagement/NugetConvert.fs
+++ b/src/Paket.Core/PackageManagement/NugetConvert.fs
@@ -285,8 +285,8 @@ let createDependenciesFileR (rootDirectory : DirectoryInfo) nugetEnv mode =
                     |> Option.toList 
                     |> List.map (fun fw ->
                         let restrictions, problems = Requirements.parseRestrictionsLegacy false fw
-                        for problem in problems do
-                            Logging.traceErrorfn "Could not detect any platforms from '%s' in %O %O, please tell the package authors" problem.Framework name version
+                        for framework in problems |> Seq.choose (fun x -> x.Framework) do
+                            Logging.traceErrorfn "Could not detect any platforms from '%s' in %O %O, please tell the package authors" framework name version
                         restrictions)
                 | _ -> []
             let restrictions =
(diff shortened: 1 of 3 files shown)


forki:
This lead to #2822 so I reverted for now. Let's investigate if this was intentional

csmager:
Sure. Would guess the code I didn’t touch only parses ‘framework’ and passes ‘s: net45’ along to be parsed. 

I guess we need to decide whether we want to allow ‘frameworks’ or not. Either way, it needs a better error message I think.

forki:
I think since we "broke" something at day 1 I would prefer to allow it.
Just in favor of the users...

Am 04.10.2017 19:38 schrieb "Charles Mager" <notifications@github.com>:

csmager:
Agreed. Will add a few more tests with ‘framework’ with and without various combinations of ‘s’ and ‘:’

forki:
Thank you. That would probably help a lot.

Am 04.10.2017 19:45 schrieb "Charles Mager" <notifications@github.com>:

csmager:
I've checked this out, was as I expected - we find 'framework', read the text after it, remove any occurrence of `:` and then pass that along to the parser. In the case of #2822, we would try to parse `s net40`, and `s` isn't a valid framework identifier. In the previous 'loose' parsing, we would assume the `s` was `=` and parse `net40`. 

That also means that e.g. `frameworks: >= net40` and `frameworks: auto-detect` wouldn't have worked in the past.

I've made a change that will check for `frameworks` and then `framework`, which means they are now interchangeable. Will open a new PR shortly.

Issue #2822, opened 2017-10-04
frameworks vs framework


ncave:
in paket.dependencies, this worked before:
`frameworks: net45`

but seems to fail with 1.0.6:
```
Paket failed with
-> Error in paket.dependencies line 1
     Could not parse framework 's'. Try to update or install again or report a paket bug.
```
it works if changed to: `framework: net45`
which makes sense I guess, if the property was changed, so feel free to close if by design.

forki:
Urgs. Reverting that last change to give us time to  investigate
/cc @agross @matthid

Am 04.10.2017 18:57 schrieb "ncave" <notifications@github.com>:

matthid:
Looks intentional, seems like it previously was parsed correctly "by accident"

forki:
Ok. But question is where is it used. What do we break

Am 04.10.2017 19:21 schrieb "Matthias Dittrich" <notifications@github.com>:

ncave:
@forki One place is [here](https://github.com/fsharp/FSharp.Compiler.Service/blob/master/fcs/paket.dependencies#L1).

forki:
Oh breaking FCS. Nice one. ;-)
Anyway we decided to roll back (should already be available).

Am 04.10.2017 20:08 schrieb "ncave" <notifications@github.com>:

ncave:
@forki Yes, it works now, closing as fixed ;-)

matthid:
what the hell ;)

ncave:
@matthid Sorry, just tried to help, please reopen if you need it for tracking.

matthid:
nah everything is fine. I was just stating my feeling about how users managed to build a feature out of a bug/misspelling the keyword oO.

ncave:
@matthid That was my OP, to better understand if it's a bug/misspelling in FCS that needs to be fixed there, or something else. Thanks for your quick response, @forki and @matthid

matthid:
yes but I agree with @forki: if even fcs has this we should probably get over it and make it a proper (maybe undocumented) feature now ...

forki:
We should still send a pr to things where we see it.

Am 04.10.2017 22:31 schrieb "Matthias Dittrich" <notifications@github.com>:

forki:
Also TIL we have a dependencies file in official visualfsharp repo

PR #2824, opened 2017-10-05
Implement stricter parsing in Requirements.parseRestrictionsLegacy (take 2)

csmager:
This supersedes reverted PR #2816, which caused issue #2822 

It seems some existing files in the wild incorrectly used e.g. `frameworks: net40` instead of `framework: net40`, and the previous 'loose' parsing would parse the resulting expression `s net40` as `= net40`. The stricter version previously merged would bail as `s` isn't a valid framework.

This PR changes to first check for `frameworks` and then `framework` meaning the two are now interchangeable and existing files such as those described in #2822 will continue to work.


forki:
thank you so much!

Issue #2839, opened 2017-10-12
paket not extracting package with local override


daz10000:
### Description

Using paket.local to pull in a package from a local build,  the nupkg file is copied but not extracted

### Repro steps

Messy to reproduce without the entire project, but 

1. Built upstream project with project scaffold ./build.sh BuildPackage

2. add paket.local file to dependent project

`nuget GslCore group Build -> source c:\extproj\GslCore\bin`

Run .paket/paket.exe restore on dependent project

Build dependent project (and have it work)

### Expected behavior
I would expect paket to copy over the GslCore.nupkg file and unpack it.  

### Actual behavior

When the dependent project is built, it is referring to packages/build/GslCore/lib/net461/GslCore.dll    and those references all fail because the lib and lower folders are missing as paket is not extracting the nupkg.    If I remove the whole packages folder and children, then do `paket.exe restore --verbose --force` and look at the output, the only steps for this library reported are

```
$ grep GslCore out
paket.local override: nuget GslCore group Build -> c:\extproj\GslCore\bin
Cleaning C:\src\dem_gslc\packages\build\GslCore
GslCore 0.4.28 already extracted
```

It looks like it is (wrongly?) concluding that the GslCore package is already extracted and failing to extract it.

### Known workarounds

I'm stuck - I won't be able to use paket.local unless I can fix this.

daz10000:
So I worked out why this is happening by instrumenting the code that checked isExtracted.   It's looking for files that aren't the original package or the license file, and in my case the original package file is capitalized (GslCore.0.4.28.nupkg) and the file it's excluding is gslcore.0.4.28.nupkg so it sees the package file and incorrectly thinks it is extracted.  I will check all my naming, but everything should be referred to by the camelcaps form, so I think paket must be forcing lowercase somewhere, and the fix would be to lowercase the files as well as they are checked.  I'll try that now as a workaround, but is that a safe fix?

Darren


Debug: directory=C:\src\dem_gslc\packages\build\GslCore licenseFile=C:\src\dem_gslc\packages\build\GslCore\gslcore.0.4.28.license.html  // ignores license file
 fullName=C:\src\dem_gslc\packages\build\GslCore\gslcore.0.4.28.nupkg  // ignores this file
Debug:  f=C:\src\dem_gslc\packages\build\GslCore\GslCore.0.4.28.nupkg  // finds this file

daz10000:
[...]

PR #2842, opened 2017-10-12
BUG: isExtracted function was falsely returning true to comparison of…

daz10000:
Re issue #2839,  this patch should rectify. It makes file name comparison lowercase.


daz10000:
Danke!

Von meinem iPhone gesendet

PR #2845, opened 2017-10-13
Use maps instead of lists in why command

### LINKED ISSUE #2844: Paket why is very slow
### Repro steps

1. git clone https://github.com/SAFE-Stack/SAFE-BookStore.git
1. cd SAFE-BookStore
1. .paket\paket.exe why ncrontabs

### Expected behavior

Blazing fast response

### Actual behavior

image

forki:
fixes #2844


forki:
from 48s down to 7s on my PC (while dottrace attached)

Issue #2823, opened 2017-10-04
paket install slow


TheAngryByrd:
### Description

`paket install` seems to take a long time with dotnet core projects.  On my machine this repo took over 10 minutes.  Maybe my ssd is dying, maybe my internet is throttled? 

```
Paket version 5.108.0
...
Performance:
 - Resolver: 10 minutes, 28 seconds (2 runs)
    - Runtime: 10 minutes, 23 seconds
    - Blocked (retrieving package details): 891 milliseconds (21 times)
    - Blocked (retrieving package versions): 4 seconds (6 times)
    - Not Blocked (retrieving package details): 174 times
    - Not Blocked (retrieving package versions): 189 times
 - Disk IO: 450 milliseconds
 - Average Download Time: 2 seconds
 - Number of downloads: 1
 - Average Request Time: 159 milliseconds
 - Number of Requests: 212
 - Runtime: 10 minutes, 38 seconds
```

### Repro steps

https://github.com/TheAngryByrd/paket-install-slow

### Expected behavior

[Insert Sonic]

### Actual behavior

[Insert not Sonic]

### Known workarounds

Obtain a hobby between calls to install

### Other
```
Mono JIT compiler version 5.2.0.215 (d15-3/da80840 Thu Jul 20 16:43:07 EDT 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           normal
	SIGSEGV:       altstack
	Notification:  kqueue
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	LLVM:          yes(3.6.0svn-mono-master/8b1520c)
	GC:            sgen (concurrent by default)
```

```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
 Commit SHA-1 hash:  cdcd1928c9

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.0.0/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```

sandeepc24:
I have this issue on Windows 10. I avoid upgrading my packages but when I have to add a new package to my project it takes a long time as for some reason I see all projects listed in console.

[...]

PR #2846, opened 2017-10-13
memoize simplify

forki:
references #2823

Before: paket update for bookstore 5min 6s
After:  paket update for bookstore 1min 12s


forki:
Ok this does not work

PR #2851, opened 2017-10-17
Minor perf improvement for why command

theimowski:
I got from 6s down to 3s on my box for the original case #2844 

This replaces `Map` with `IDictionary` and cycle dependency prevention is a bit faster now

### DIFF
--- a/src/Paket.Core/PackageAnalysis/Why.fs
+++ b/src/Paket.Core/PackageAnalysis/Why.fs
@@ -1,6 +1,7 @@
 module Paket.Why
 
 open System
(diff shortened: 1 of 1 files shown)


forki:
Marking it twice as fast is not "minor" ;-)

Am 17.10.2017 23:00 schrieb "Tomasz Heimowski" <notifications@github.com>:

forki:
in https://github.com/fsprojects/Paket/pull/2853 I replaced the set with a hashset. Not sure if it's still correct?! Can you please review?

PR #2853, opened 2017-10-18
Use a hashset

forki:
references #2844 

brings original why sample down to 1s

### DIFF
--- a/src/Paket.Core/PackageAnalysis/Why.fs
+++ b/src/Paket.Core/PackageAnalysis/Why.fs
@@ -18,15 +18,19 @@ and LblPathNode<'a> =
 | LblPathLeaf of PackageName * 'a
 
 module AdjLblGraph =
-    let rec paths start stop visited (g: AdjLblGraph<_>) : list<LblPath<_>> =
-        let adjacents = 
-            g.[start]
-            |> Seq.filter (fun kv -> not (Set.contains kv.Key visited))
-        [ for kv in adjacents do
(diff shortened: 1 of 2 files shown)


forki:
so from 21s down to 1s - good team work @theimowski!

theimowski:
LGTM, pity there are no tests to be sure there's no regression.
Can we name the nested rec function `paths'` to better distinguish from outer fun?

forki:
@theimowski my question is: are we ver going back here? In my optimization we only add to the set and never reuse old sets.

theimowski:
Hm crap you're right. Now it might skip some chains ...
E.g. when we have a diamond:
```
A
  B
B
  C
  D
C
  E
D
  E
E
```

Then it would output just `A->B->C->E` and skip `A->B->D->E`
So it changes behavior - but aside maybe such behavior is ok?

forki:
so revert?

theimowski:
That's up to you - behavior changed, but I'm not sure whether it's ok or not.
By that I mean the output shown when `--details` is specified might get very long for bigger trees

forki:
@matthid  ?

theimowski:
Do we know how does yarn handle such cases?

forki:
Dunno. But I'm fine with revert. 3s is still usable.

theimowski:
Ok then how about revert and create issue to cover the code with some tests and optionally then improve perf

PR #2871, opened 2017-10-28
generate nuspecs in IntermediateOutputPath

### LINKED ISSUE #2865: Paket.restore.targets creates nuspec in wrong folder
I just upgrade a solution to the new csproj format, but have a problem with packaging with `dotnet pack`
 afterwards. As I understand `paket.Restore.targets` override the packaging process by generating nuspec files which are modified afterward by `paket fix-nuspecs`. The problem is that these nuspec files are duplicated in `obj` and `obj/$(Configuration)`, and only last one is correct. There is one project which does not have a `paket.references` file and it is correctly generated.

As I work-a-round I tried deleting the entire section in the  `paket.Restore.targets` and now everything works as expected. Dependencies are also correct.

johlrich:
this is so they go in obj/Configuration/ instead of obj/

closes #2865

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -136,15 +136,21 @@
   </Target>
 
   <Target Name="PaketOverrideNuspec" AfterTargets="GenerateNuspec" Condition="('$(IsPackable)' == '' Or '$(IsPackable)' == 'true') And Exists('$(MSBuildProjectDirectory)/obj/$(MSBuildProjectFile).references')" >
+    <ItemGroup>
+      <_NuspecFilesNewLocation Include="$(BaseIntermediateOutputPath)$(Configuration)/*.nuspec"/>
+    </ItemGroup>
+
     <PropertyGroup>
       <PaketProjectFile>$(MSBuildProjectDirectory)/$(MSBuildProjectFile)</PaketProjectFile>
       <ContinuePackingAfterGeneratingNuspec>true</ContinuePackingAfterGeneratingNuspec>
       <UseNewPack>false</UseNewPack>
       <UseNewPack Condition=" '$(NuGetToolVersion)' != '4.0.0' ">true</UseNewPack>
+      <AdjustedNuspecOutputPath>$(BaseIntermediateOutputPath)$(Configuration)/</AdjustedNuspecOutputPath>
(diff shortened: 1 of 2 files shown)


forki:
Pinging @matthid because I'm on vacation

forki:
@johlrich  can you please take a look at the broken test in appveyor
(travis is expected to be red)

johlrich:
Looks like I didn't test with the IntermediateOutputPath variable enough. That one includes the configuration and the target framework in the path. I switched back to `$(BaseIntermediateOutputPath)$(Configuration)/` and will keep an eye on the appveyor result.

forki:
Mhm still red?

forki:
Maybe the test is no good?

johlrich:
Hmmmmm the test passes locally after switching. The test also fails here without the changes complaining about no description, and when looking at the in the scenario temp folder I see the same situation as the issue: an extra blank nuspec at obj root. Not sure why it's passing now.

Strange that I'm getting the opposite results as appveyor. Not really sure of a good way to troubleshoot atm. @forki any ideas? 

My only thought was to somehow capture the scenario temp folder from the appveyor build to take a closer look in both scenarios, but not sure how to do that tbh.

forki:
Maybe different dotnet cli versions?

Am 29.10.2017 15:35 schrieb "Jonathan Ohlrich" <notifications@github.com>:

johlrich:
@forki I think that's it. I isolated that one test and then ran it with `DOTNET_EXE_PATH` set to both the preview one that the build downloads (2.1.0-preview1-007002) and the one I have installed (2.0.2). The preview one fails, the install one passes.

Going to change to 2.0.2 so appveyor gets kicked off again with it to prove that out there. If that works then we'll have to see how you want to handle that. 🤞

johlrich:
More weirdness, the full build is currently churning through the integration tests on my machine with 2.0.2 but appveyor can't seem to get past restoring Paket.Core.preview3.

forki:
Yes changing the dotnet cli version is not the solution here

Am 29.10.2017 20:10 schrieb "Jonathan Ohlrich" <notifications@github.com>:

johlrich:
Ok, it is now checking for existence in the new location and if it couldn't find it, falling back to the old location. Build finally green and worked with both cli versions here.

Not super familiar with MSBuild so let me know if there was a simpler way to do that since Exists doesn't seem to do wildcards

Issue #70, opened 2014-09-09
Ignore *.*projs that do not look like MSBuild projects


agross:
I have inherited a Git repository that contains a zreproj (no idea what it is) that is definitely not MSBuild. Paket fails to modify this file because it does not contain a Project node. Please ignore such files.

szul:
@forki Hate to bring up a long dead issue, but I'm working on an IronPython project that uses a *.pyproj project file extension. Has there been any other movement/discussion on alternative project file extensions? I'd like to be able to have paket update this project file, but right now, I have to constantly change the extension back and forth to do so, it seems.

forki:
Just grep for fsproj and csproj in the repo here and add your stuff in a
PR. Will be fastest way ;-)

Am 02.11.2017 20:54 schrieb "Michael Szul" <notifications@github.com>:

PR #2885, opened 2017-11-04
pyproj support

szul:
*.pyproj project file support as per my [comments](https://github.com/fsprojects/Paket/issues/70#issuecomment-341538948) on issue #70.

Tests are passing from what I see. I used the subsequent output of the build in a local project, and it populated the *.pyproj file appropriately.

I did not write specific tests for this, as it is pretty basic support. Let me know if that is highly desired. I just needed to get Paket to recognize *.pyproj files as a file type that needed to be updated with the installed packages.


forki:
cool. thanks

PR #2901, opened 2017-11-16
Remove possibly nonexistent extension safely

### LINKED ISSUE #2900: Paket failed: String cannot be of zero length (a file has no extension issue again)
### Description
Seems to be the same issue as #2405. When running paket install, it fails to remove the nonexistent extension of a `LICENCE` file contained in a nuget package and fails to install it.

### Repro steps

1. Create an empty project

2. Add the following to paket.dependencies

```
nuget Zstandard.Net
```

3. Run paket install.

### Expected behavior
Installs Zstandard.Net

### Actual behavior
Fails to install with a paket error:
 ```
Paket failed with
-> ArgumentException: String cannot be of zero length.
   Parameter name: oldValue
```

I've tracked it down to line 59 of InstallModel.fs

domdom:
Fixes #2900

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/InstallModel.fs
+++ b/src/Paket.Core/PaketConfigFiles/InstallModel.fs
@@ -56,7 +56,10 @@ type MsBuildFile = {
 module MsBuildFile =
     let ofFile (f:FrameworkDependentFile) =
         let fi = FileInfo(normalizePath f.File.FullPath)
-        let name = fi.Name.Replace(fi.Extension, "")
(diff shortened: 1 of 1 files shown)


domdom:
I think there is actually a better solution, which is to use the Path.GetFileNameWithoutExtension method which seems to suit the use case perfectly. I can also replace other occurrences of this in the code base.

forki:
thanks!

domdom:
I'm happy to help, thanks for the work on paket :)

Issue #2889, opened 2017-11-08
Running "paket pack" fails if the project file uses wildcards for compile includes


ruhullahshah:
### Description
Using `<Compile Include="*.cs" />` in the .csproj file leads to paket runtime error ArgumentException: Illegal characters in path.

It might be related to https://github.com/fsprojects/Paket/issues/1462

ruhullahshah:
I think 

let sourceFile =
                compileNode
                |> getAttribute "Include"
                |> fun attr -> attr.Value
                |> normalizePath
                |> fun relPath -> Path.Combine(projectFolder, relPath)

from GetCompileItems, ProjectFile.fs is wildcard agnostic. Shouldn't it return a list of source files if there are wildcards in the "Include" attribute of the <Compile> node?

enricosada:
Additional scenario from https://github.com/fsprojects/Paket/issues/1462:

```xml
<Compile Include="GENERATED_CS_LINKED\**\*.cs">
      <Link>GENERATED_CS_LINKED\%(RecursiveDir)%(FileName).cs</Link>
</Compile>
```

and 

```xml
<Compile Include="GENERATED_CS\**\*.cs" />
```

Issue #2068, opened 2016-12-05
Wishlist: Reduce diffs to csproj files after "paket install"


vivainio:
### Description

csproj files are sometimes changed by Paket in redundant ways:

- <Reference> items move down a few items
- Whitespace normalization happens

etc. etc.

Proposal: analyze whether the semantics of xml document change in any way, and do not do the change the csproj if that didn't happen.

Vilmir:
We would benefit from this improvement. paket install generates frequently noisy csproj file changes in our codebase as well

smoothdeveloper:
Most of the file changes are happening if no framework restriction is defined, anytime a new fancy .net platform appears, or whenever the nuget packages support a new fancy .net platform, the conditional will changes.

In my experience this doesn't happen much when you set framework restriction (such has `framework: >= net40`).

I understand that those are not points you listed, and having semantics analysis would be nice (also avoids unecessary project reload when solution is opne), but I'm unsure it would be easy to get this done right.

vivainio:
@smoothdeveloper easy is relative, but comparing the before/after XML element trees for semantic differences should be quite realistic.

smoothdeveloper:
@vivainio sorry if I sounded pessimistic, I know for sure that in my developer life, diffing on xml file changes can be a headache but I agree it can be achieved. :smile:

vivainio:
@smoothdeveloper the trick is to compare parsed XElement form, not text

jbaehr:
I have another proposal to address this issue: Use the API `Microsoft.Build.Construction` to manipulate the DOM of msbuild files. This approach could also be used to solve #1949.

Issue #1949, opened 2016-10-07
Question: Evaluating MSBuild-Properties from paket pack


Dirk-Kuypers:
We want to split up a huge C# project with around 400 projects into smaller components which should be delivered as nuget packages. I am just looking into Paket because it seems to be very promising, so please be gentle if this question seems dumb.

### Description

In our C# project files we have a global property called $(RootCodeDir) which is set to the root directory of our software if not set from outside:

    <RootCodeDir Condition=" '$(RootCodeDir)' == '' ">$(MSBuildThisFileDirectory.Substring(0, $(MSBuildThisFileDirectory.IndexOf(`repo`))))</RootCodeDir>

With this trick we can ensure that every assembly is written to the same output directory, f.i.:
    <OutputPath>$(RootCodeDir)\repo\bin\debug\</OutputPath>

And if we move around the folder of the project we do not have to change the project file and fiddle with changing the path at multiple locations from ..\..\.. to ..\..

In the example below I have removed the $(RootCodeDir) from the OutputPath, but as pack symbols tries to pack some global files with $(RootCodeDir) in their path, it fails.

c:\repo\path\to\my\project>..\.paket\paket.exe --verbose pack symbols output out buildplatform x86 buildconfig Release_Package
Paket version 3.21.0.0
found: c:\repo\paket.dependencies
Parsing c:\repo\paket.dependencies
Loading assembly metadata for c:\repo\path\to\my\project\bin\x86\Release_Package\my.dll
Paket failed with:
        Could not find a part of the path 'c:\repo\path\to\my\project\$(RootCodeDir)\repo'.
StackTrace:
     at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileSystemEnumerableIterator`1.CommonInit()
   at System.IO.FileSystemEnumerableIterator`1..ctor(String path, String originalUserPath, String searchPattern, SearchOption searchOption, SearchResultHandler`1 resultHandler, Boolean checkHost)
   at System.IO.Directory.GetFiles(String path, String searchPattern)
   at <StartupCode$Paket-Core>.$ProjectFile.GetCompileItems@1626.Invoke(CompileItem compileItem)
   at Microsoft.FSharp.Collections.IEnumerator.map@111.DoMoveNext(b& )
   at Microsoft.FSharp.Collections.IEnumerator.MapEnumerator`1.System-Collections-IEnumerator-MoveNext()
[...]

PR #2913, opened 2017-11-24
Support wildcards in MSBuild Compile Includes using the MSBuild API

ruhullahshah:
Fixes issue #2889 

- Use MSBuild Evaluation API to evaluate the `include` inside the `Compile` node.  Enables support for diverse project files and better maintainability.

### DIFF
--- a/src/Paket.Core.preview3/Paket.Core.fsproj
+++ b/src/Paket.Core.preview3/Paket.Core.fsproj
@@ -97,5 +97,9 @@
       <PackagePath>lib/netstandard1.6</PackagePath>
     </Content>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Build" Version="15.3.409" />
+	<PackageReference Include="Microsoft.Build.Framework" Version="15.1.548" />
+  </ItemGroup>
   <Import Project="..\..\.paket\Paket.Restore.targets" />
 </Project>
--- a/src/Paket.Core/Paket.Core.fsproj
+++ b/src/Paket.Core/Paket.Core.fsproj
@@ -144,6 +144,9 @@
     <None Include="Prototypes\script-generation.fsx" />
   </ItemGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.Build">
+      <Private>False</Private>
+    </Reference>
     <Reference Include="mscorlib" />
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
--- a/src/Paket.Core/PaketConfigFiles/ProjectFile.fs
+++ b/src/Paket.Core/PaketConfigFiles/ProjectFile.fs
@@ -1943,32 +1943,30 @@ type ProjectFile with
 
     member this.GetCompileItems (includeReferencedProjects : bool) cache = 
         let getCompileRefs projectFile =
-            projectFile.Document
-            |> getDescendants "Compile"
-            |> Seq.map (fun compileNode -> projectFile, compileNode)
-
-        let getCompileItem (projectFile, compileNode) =
+            let project = Microsoft.Build.Evaluation.Project(projectFile.FileName)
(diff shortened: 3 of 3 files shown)


forki:
I don't think it's a good idea to reference that lib. At least we would need to get it from nuget or something and ilmerge it. Is there no other way?

ruhullahshah:
I used the dotnet add reference (https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-add-reference). I am not so well versed with dotnet tool but will try to figure out some other way for adding it then.

jbaehr:
@forki what are your concerns against using `Microsoft.Build` for the `.proj`-parsing and evaluation? Is it unavailable on certain platforms you're targeting? I've seen that paket has reimplemented those bits (at least partially)...

In my eyes it make reading and manipulating the msbuild DOM easier, with less code and could address issues like #1949, or #2068, too.

It is available via nuget.org, btw: https://www.nuget.org/packages/Microsoft.Build/ (for .NETFramework 4.6 and .NETStandard 1.5)

forki:
@jbaehr yes we had many many issues with that in things like ProjectCracker. People like @Krzysztof-Cieslak and @enricosada can tell you long stories

ruhullahshah:
IMO, this pull request aims at changing a minimum part of the Paket code to achieve a functionality that would be otherwise be hard and non-maintainable if implemented without using the MSBuild API. For example, parsing include paths with *, ** and ?. Maybe it is not the right choice overall for parsing a .csproj, which I cannot comment upon at this moment but I think that a blend of custom parsing code combined with MSBuild API usage can lead to maintainable and robust code.

forki:
The problem is it binds us to a very specific MSBuild version and we
introduce GAC issues and assembly loading issues.

If we really need to go down this road then we need to use the nuget
package and ilmerge the dll

Am 27.11.2017 12:47 schrieb "Ruh Ullah Shah" <notifications@github.com>:

IMO, this pull request aims at changing a minimum part of the Paket code to
achieve a functionality that would be otherwise be hard and
non-maintainable if implemented without using the MSBuild API. For example,
parsing include paths with *, ** and ?. Maybe it is not the right choice
overall for parsing a .csproj, which I cannot comment upon at this moment
but I think that a blend of custom parsing code combined with MSBuild API
usage can lead to maintainable and robust code.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/fsprojects/Paket/pull/2913#issuecomment-347159067>, or mute
the thread

ruhullahshah:
I agree on changing the dependency to nuget-based rather than tying it to a specific version. I will update the PR as soon as I have access to my development machine later this week.

Krzysztof-Cieslak:
Not only it introduces the dependency on GAC, it depends on specific MSBuild version (which may or may not be present on the machine), it opens all the problems connected with different MsBuild versioning on the mono and problems with assembly redirects. 

I strongly believe it's a bad idea given the limited needs for the project file evaluation Paket needs for the most cases. Also, with my experience with MSBuild (:heart:) if you decide to fix the problem with MSBuild you end up having two problems and I strongly believe the maintenance cost of such PR way outweighs the maintenance cost of some additional custom code for better parsing of project files.

enricosada:
Do not include msbuild evaluation **as library**.

is really a bad choice long term, because it tie you to a specific msbuild version (see projectcracker pain). Also you dont get the environemtn fully initialized, some properties are missing (see `VisualStudioVersion` used to choose F# target)

If you want a feedback, if you really really want to do that, use **msbuild as process** .

So invoke `msbuild myproj.csproj /t:Something` to gather info.

see https://github.com/enricosada/dotnet-proj-info this handle all msbuild version, because it ASK installed msbuild about info, not use that as library

ruhullahshah:
Closing PR due to restrictions posed by Microsoft.Build. Looking into other feasible cross platform methods for .csproj parsing other than custom code

PR #2923, opened 2017-12-02
Prefer Requirements that were already resolved to get conflicts early

forki:


tests/Paket.Tests/Resolver/PrereleaseSpecs.fs
```diff
@@ -286,30 +286,4 @@ nuget PackageA prerelease
     let resolved = ResolveWithGraph(cfg,noSha1, VersionsFromGraphAsSeq graph, PackageDetailsFromGraph graph).[Constants.MainDependencyGroup].ResolvedPackages.GetModelOrFail()
     getVersion resolved.[PackageName "packageA"] |> shouldEqual "1.0"
     // we need to take the stable as the alpha package has no resolution
-    getVersion resolved.[PackageName "packageB"] |> shouldEqual "1.0"
-
-[<Test>]
-let ``should not depend on resolution order``() =
```
matthid: Is it a good idea to get rid of this test?

forki: the changed resolver found another valid resolution. One that became valid after we redefined what prerelease means

matthid: Ähm and which one? I thought I designed the test in a way that there is only a single valid resolution (which needs to contain `packageD` in Version `1.0`:

D-1.1alpha -> B-1.1-alpha doesn't exist
So D-1.0 is set.

Honestly I think taking B-1.1 is wrong, because `nuget PackageD prerelease` should not allow prereleases in a completely different subtree (which is A -> B -> C (prerelease), note that D doesn't depend on A).

But even if we now allow that for some reason (because we allow D -> B -> C) I feel like we should change the test accordingly and add comments to represent this behavior.

So I guess the result should now be:
 - D-1.0
 - B-1.1
 - C-1.1-alpha
 - A-1.0

correct?

forki: Yes we need to decide what to do here. If we really want to allow this. But it's unrelated to this pr. The pr only shows that we have this issue

forki:
I really don't understand why this integration test now runs into timeout. @matthid did you add this one? Any ideas?

matthid:
Yes the test was a workaround for #2294 I guess it cannot find a solution in the given time after these changes? This is a difficult puzzle for paket (and I think I have seen this test failing before occasionally)...

forki:
Ok thanks. So it must be fixed with the other charges I'm planning.

Am 02.12.2017 13:42 schrieb "Matthias Dittrich" <notifications@github.com>:

tests/Paket.Tests/Resolver/PrereleaseSpecs.fs
matthid: The PR removes the test ;)

forki: Yes. Will not merge it. It's just to let ci check the integration tests and not break early on unit tests

forki:
the thing is: we don't use the set order anymore.
Take a look at getCurrentRequirement - it's calculating the minimal element manually with a loop. Set order is completely ignored

PR #2928, opened 2017-12-04
Boost conflicts during fusion - fixes #2922

### LINKED ISSUE #2922: "paket install" and the loop of doom
### Description

On conflicts paket goes into "loop of doom" not providing too much information what actually might have gone wrong. CPU utilization goes to 100% (more precisely 100% of one core plus a little bit more).

### Repro steps

[paket-conflict.zip](https://github.com/fsprojects/Paket/files/1523334/paket-conflict.zip)

Unpack archive, and then:
```
cd paket.conflict.lib
paket install
dotnet build
dotnet pack --output ..
cd ../paket.conflict.app
paket install
```
(~~or just use bang.cmd~~)

...now sit back and observe paket going into "loop of doom". Never got to the end of this loop, but what I can say for sure, for real project (with more than 3 dependencies) I left it for 10 hours and it did not end.

So, what triggers it?

There is a library (`paket.conflict.lib`) which depends on 3 things:
```
nuget Microsoft.EntityFrameworkCore
nuget Pomelo.EntityFrameworkCore.MySql
nuget Moq
```

There is also an application (`paket.conflict.app`) which depends on mentioned library but also on another 3 things:

```
nuget paket.conflict.lib
nuget Hangfire.MySql.Core
nuget MySqlConnector < 0.30
nuget Moq
```

You can notice restriction on `MySqlConnector` as latest (2.1.5) `Hangfire.MySql.Core` does not work with newer versions. At the same time, newest `Pomelo.EntityFrameworkCore.MySql` (2.0.1) requires `MySqlConnector > 0.32`. So yes, we have a conflict, as, transitively, library wants `> 0.32` while app wants `< 0.30`. 

There is no clue from paket what might have happened, it actually tries to scan all versions of `Moq` which has nothing to do with the problem.

I do understand that this is a algorithmicly complex problem (NPC? NPH?) but even a warning line saying "trying earlier version of XXX because ..." might be useful. Maybe it does print it on the end but I never waited long enough.

**NOTE**: this is just for 3 dependencies. Add 'Microsoft.AspNotCore.*` and watch it then.

### Expected behavior

Better conflict detection, better logging?

### Actual behavior

Endless (?) loop of doom.

### Known workarounds

None, but take a closer look at your `paket.lock`/`paket.dependencies` files, especially with `less than` restrictions.

forki:
fixes "loop of doom" by using conflict boost

@matthid for some reason we forgot to boost when we report conflicts.
This is now dramatically faster and even reports the conflict along the way


0x53A:
Just wanted to confirm that I hit the same issue - paket update running forever - which was fixed by building from this branch.

forki:
the problem with this branch is that I'm still hitting couple of other edge cases

PR #2936, opened 2017-12-12
Don't lock files in the bootstrapper that we are only reading

vbfox:
Create lot of bugs when the bootstrapper is executed in parallel
(like for dotnet restore)


forki:
giphy 3

PR #2940, opened 2017-12-14
Avoid replacing load script files if the contents are the same

### LINKED ISSUE #2939: Load script generation should touch/replace files only if the file really changes
See title

This would decrease locking between ionide and cmd line builds

smoothdeveloper:
This fixes #2939.


forki:
thanks!

PR #2952, opened 2017-12-30
Paket shows parsing errors while doing BindingRedirects - #2408

MrLuje:
This is not fixing the issue but at least indicates which file is causing the issue.
(Example comes from a web.config from npm package selenium-webdriver)


forki:
thanks. Hope you find the parsing error now.

PR #2961, opened 2018-01-03
Remove redundant example from binding redirect docs

chrnola:
This section of the page already explains how to use the
'redirects: force' setting, so remove the less descriptive example.


forki:
thx

PR #2966, opened 2018-01-05
Using Git, GitHub and Http sources in PublicAPI

frelon:
Right now there is no way (that I'm aware of) to add a git/github/http dependency using the PublicAPI.

This is a first draft to be able to specify those dependency-types when adding a package.

Is PackageRequirementKind the correct type for this? I used it because the .Package member was used to specify the "nuget" string...

There are some edge-cases that probably won't work with this implementation (github-dependencies is defined with "<repo>:<version>" instead of "<package> <version>").

### DIFF
--- a/src/Paket.Core/PackageManagement/AddProcess.fs
+++ b/src/Paket.Core/PackageManagement/AddProcess.fs
@@ -6,6 +6,7 @@ open System
 open System.IO
 open Paket.Domain
 open Paket.Logging
+open Requirements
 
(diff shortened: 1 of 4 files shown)


enricosada:
@frelon the `PackageRequirementKind` was for nuget packages. git/github etc are not nupkg, so ihmo you shouldnt use that.
all that resolution is for nupkg, doesnt really apply to git stuff

frelon:
Alright, I'll open up an issue to discuss this instead! :)

PR #2990, opened 2018-01-17
add parameter `--type` to `paket add`

enricosada:
enable `paket add dotnet-fable --type clitool` or `paket add dotnet-fable -t clitool`

default is `nuget`, so `paket add Argu` is the same as `paket add Argu -t nuget`


forki:
Cool. Kudos buddy

PR #2994, opened 2018-01-18
Reduce next-link warning to once per method/endpoint, omitting the query, fixes #2848

### LINKED ISSUE #2848: Allow to disable repeated warnings for misbehaving sources
### Description

Allow to disable warnings resulting from a misbehaving package source, ideally using a command line parameter, like `--nowarn`, optionally supporting warning numbers to suppress. 

### Repro steps

Use a misbehaving package source, like Artifactory with their broken "next" links. 

### Expected behavior

Paket should warn about unexpected behavior unless told otherwise, but only once per package & source combination. 

### Actual behavior

Paket produces warnings for each single package, e.g. in our case with Artifactory: 
`At least one 'next' link (index 0) returned a empty result ...` 
flooding the console/logs with known and currently useless information. 

### Known workarounds

None, as we can neither fix Artifactory nor use different server.

viktor-svub:
Warnings for empty next-link result are repeated for every single occurrence, which caused console/log flooding when we use Artifactory for all of our internal dependencies (repository behavior fix or replacement does not seem negotiable). 

Reduced the warnings to once per specific API endpoint and shortening the report by trimming the query, which only contains redundant information.

### DIFF
--- a/src/Paket.Core/Dependencies/NuGetV2.fs
+++ b/src/Paket.Core/Dependencies/NuGetV2.fs
@@ -73,7 +73,11 @@ let private followODataLink auth url =
 
             match atLeastOneFailed with
             | Some i ->
-                traceWarnfn "At least one 'next' link (index %d) returned a empty result (noticed on '%O'): ['%s']" i url (System.String.Join("' ; '", linksToFollow))
+                let mutable uri = null // warn once per apecific API endpoint, but try to cut the query
+                let baseUrl = if Uri.TryCreate(url, UriKind.Absolute, &uri) then uri.AbsolutePath else url
+                traceWarnIfNotBefore baseUrl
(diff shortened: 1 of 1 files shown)


src/Paket.Core/Dependencies/NuGetV2.fs
```diff
@@ -73,7 +73,11 @@ let private followODataLink auth url =
 
             match atLeastOneFailed with
             | Some i ->
-                traceWarnfn "At least one 'next' link (index %d) returned a empty result (noticed on '%O'): ['%s']" i url (System.String.Join("' ; '", linksToFollow))
+                let mutable uri = null // warn once per apecific API endpoint, but try to cut the query
```
forki: typo

PR #2995, opened 2018-01-18
ci should always be green

enricosada:
flacky tests should be moved to ci jobs allowed to fail

- [X] fix appveyor
- [x] fix travis

miscellaneous:

- [x] test results in appveyor test tab (log are too long, results give also test duration hints)
  - netcore integration tests (also unit maybe) are not added yet, need to rename the assembly too (otherwise are mixed with net version)
- [x] is `flaky` not `flacky`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -196,6 +196,7 @@ System.Management.Automation.dll
 docs/tools/FSharp.Formatting.svclog
 docs/content/paket.exe
 TestResult.xml
+*.trx
 tests/Paket.Tests/NuGetConfig/PasswordConfig.xml
 paket.local
 dotnetcore
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,28 +1,34 @@
 language: csharp
+dist: trusty
+
+sudo: false
 
 env:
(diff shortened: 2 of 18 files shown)


enricosada:
ref https://bugzilla.xamarin.com/show_bug.cgi?id=57516 about mono ssl error

enricosada:
Ok @forki this is ready to merge.

both travis and appveyor run exactly same fake targets (goal is parity) and are splitted in a similar build matrix

appveyor now populate the `test` tab with test result, useful mostly to check timining.

i moved some failing tests to `flaky` , we can use that as temporary dump for failing tests instead of ignore, so at least we see the failure message.

### appveyor

both .net and .net core paket pass all non flaky tests
afaik atm is enough, we can split integration tests in two later (just adding another category for long tests) when we reach appveyor build max time of 60min.

### travis

- .net core paket tests should be always green, are not allowed to fail and shouldnt
- .net paket tests may fails because of a bug https://github.com/fsprojects/Paket/issues/3009 who sporadically (< 4%) make random test failing while making web request (usually from https nuget api). while waiting for a fix may be good enough, that will not fix out issue, ihmo we should retry when that specific expection, so is more robust for users too.

forki:
Awesome work!

PR #2997, opened 2018-01-18
Trace not-found and blacklist warnings as actual warnings

viktor-svub:
Specific explicitly handled and repeated "errors" are changed to "warn if not before" to avoid stderr writes and/or console/log flooding when the resolver actually continues and can succeed. 
Trace not-found and blacklist warnings as actual warnings, and only once per query and source combination.


src/Paket.Core/Dependencies/NuGet.fs
```diff
@@ -301,7 +312,7 @@ let rec private getPackageDetails alternativeProjectRoot root force (parameters:
                             return! tryV3 nugetSource force
                         with
                         | exn ->
-                            eprintfn "Possible Performance degradation, V3 was not working: %s" exn.Message
+                            traceWarnfn "Possible Performance degradation, V3 was not working: %s" exn.Message
```
forki: /cc @matthid good idea?

forki:
wow you are on fire!

PR #2998, opened 2018-01-18
Only overwrite NuGet metadata cache when needed, cleanup and exception handling

viktor-svub:
Cleaned up NuGet metadata cache access based on production experience: 

1/ with explicit "force", do not re/query and overwrite the metadata cache twice 

2/ stale cache file cleanup should not be performed on every single read -- doing this only on "force" seems like a good idea, there is "paket clear-cache" in case more "forceful" solution is needed 

3/ do not overwrite cached metadata on each access with "force", if the content already equals -- the read and comparison is faster and does not cause file-access violations

### DIFF
--- a/src/Paket.Core/Dependencies/NuGetCache.fs
+++ b/src/Paket.Core/Dependencies/NuGetCache.fs
@@ -144,16 +144,22 @@ type NuGetPackageCache =
 
 let inline normalizeUrl(url:string) = url.Replace("https://","http://").Replace("www.","")
 
-let getCacheFiles cacheVersion nugetURL (packageName:PackageName) (version:SemVerInfo) =
+let getCacheFiles force cacheVersion nugetURL (packageName:PackageName) (version:SemVerInfo) =
     let h = nugetURL |> normalizeUrl |> hash |> abs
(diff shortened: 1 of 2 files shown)


src/Paket.Core/Dependencies/NuGetCache.fs
```diff
@@ -204,15 +208,26 @@ let tryGetDetailsFromCache force nugetURL (packageName:PackageName) (version:Sem
...
-    let cacheFile, oldFiles = getCacheFiles NuGetPackageCache.CurrentCacheVersion nugetURL packageName version
-    for f in oldFiles do
-        File.Delete f
+    let cacheFile = getCacheFiles force NuGetPackageCache.CurrentCacheVersion nugetURL packageName version
     let get() =
         async {
             let! result = get()
             match result with
             | ODataSearchResult.Match result ->
                 File.WriteAllText(cacheFile.FullName,JsonConvert.SerializeObject(result))
```
forki: I assume this line now needs to go

forki: will remove it myself.

forki:
Awesome. Keep them coming ;-)

PR #3023, opened 2018-02-01
Add GitHub sources in PublicAPI and CLI

frelon:
work in progress for implementing Git, GitHub, Gist and Http sources in CLI tool and PublicAPI

### DIFF
--- a/src/Paket/Commands.fs
+++ b/src/Paket/Commands.fs
@@ -69,7 +69,7 @@ type AddGithubArgs =
     | [<ExactlyOnce;MainCommand>] Repository of repository_name:string
     | [<Unique;AltCommandLine("-V")>] Version of version_constraint:string
     | [<Unique;AltCommandLine("-g")>] Group of group_name:string
-    | [<Unique;AltCommandLine("-file")>] File of file_name:string
+    | [<Unique>] File of file_name:string
 with
     interface IArgParserTemplate with
         member this.Usage =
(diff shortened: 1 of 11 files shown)


enricosada:
about cli, @forki @matthid what do you think?

```
paket git add .. // and list remove etc
paket github add ..
paket http add ..
```

like `git remote`?

add lot of cmd `add-git` `remove-git` etc is annoying

Asking because i'd like something like that for repotool too

frelon:
@enricosada I changed the cli and added the functions in AddProcess etc. Now you can run

`paket github add fsprojects/paket`

and it generates the paket.dependencies and paket.lock file correctly. It also supports the version, file and group parameters.

forki:
so no longer WIP?
@matthid  @enricosada can you please review?

frelon:
I could clean up the empty functions AddHttp, AddGit etc if you want to merge only the Github-functionality. 
Then it's no longer WIP, no.

src/Paket/Commands.fs
```diff
@@ -625,6 +647,7 @@ with
         member this.Usage =
             match this with
             | Add _ -> "add a new dependency"
+            | Github _ -> "commands to manipulate gihub repository references"
```
agross: Could you please change `gihub` to `GitHub`?

frelon: Absolutely! Opened #3100 for this.

PR #3030, opened 2018-02-07
Extended SemVer v2 compliance and reliability improvements

viktor-svub:
- updated Dependencies parser to accept semver2 twiddle-wakka constraints -- versions like `1.2.3-beta.4` ended up with invalid upper constraint (`>= 1.2.3-beta.4 < 1.2.3-beta.0 (beta)`)
- changed version comparison to be more obvious and to follow semver2 rules more closely
- changed extended SemVer Build to `bigint`, as string builds resulted in "unique" versions, not properly resolvable or comparable to anything else
- extended SemVer parser to accept well-formed semver2 strings without misinterpreting embedded numbers in alphanumeric segments -- which is now special-cased for for semver1 support
- updated SemVer parser to interpret malformed versions as close as possible to correct ones -- until now, they were accepted as-is without warnings or errors, but could not be properly compared, e.g. version like `1.2.3.beta.4` was accepted on all paths, but considered neither prerelease, nor part of `1.2.3` range
- changed transitive prerelease versions of prerelease packages to use `Concrete` name where possible -- with "diamond dependencies" in graph, having transitive dependencies on `All` caused resolver loops, and broke the expectation of using only the named prerelease, unless `prerelease` was specified explicitly
- added more prerelease and its transitivity information to conflict report
- added tests and new cases to cover previously undefined edge cases


src/Paket.Core/Dependencies/DependenciesFileParser.fs
```diff
@@ -25,29 +26,36 @@ module DependenciesFileParser =
...
-            let parsed, number = Int32.TryParse values.[index]
-            if parsed then values.[index] <- (number + 1).ToString()
-            if values.Length > 1 then values.[values.Length - 1] <- "0"
-            values
-
-        let parts = minimum.Split '.'
-        let penultimateItem = Math.Max(parts.Length - 2, 0)
-        let promoted = parts |> promote penultimateItem
-        String.Join(".", promoted)
+    let twiddle (minimum:SemVerInfo) =
```
viktor-svub: changed to utilize existing SemVer parser, to handle multi segment prerelease ranges correctly
original code changed the last numeric segment and ended up with invalid upper constraint

src/Paket.Core/Dependencies/DependenciesFileParser.fs
```diff
@@ -25,29 +26,36 @@ module DependenciesFileParser =
...
+                let items = texts |> List.filter ((<>) "") |> List.distinct
+                match items with
                 | [] -> 
                     versions
                     |> List.collect (function { PreRelease = Some x } -> [x.Name] | _ -> [])
                     |> List.distinct
                     |> function [] -> PreReleaseStatus.No | xs -> PreReleaseStatus.Concrete xs
                 | [x] when String.equalsIgnoreCase x "prerelease" -> PreReleaseStatus.All
-                | _ -> PreReleaseStatus.Concrete texts
+                | _ -> PreReleaseStatus.Concrete items
```
viktor-svub: assuming the filter for empty strings is necessary, we should use the filtered result

src/Paket.Core/Dependencies/PackageResolver.fs
```diff
@@ -181,31 +181,31 @@ module ResolutionRaw =
...
         | conflicts ->
-            let hasPrereleases = Seq.exists (fun r -> r.VersionRequirement.PreReleases <> PreReleaseStatus.No) conflicts
 
             errorReport.AddLine (sprintf "  Conflict detected:")
 
             let getConflictMessage req =
                 let vr = formatVR req.VersionRequirement
-                let pr = formatPR hasPrereleases req.VersionRequirement
+                let pr = formatPR true req.VersionRequirement
+                let tp = if req.TransitivePrereleases then "*" else ""
```
viktor-svub: prerelease transitivity hint

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -47,42 +53,72 @@ type PreRelease =
...
     static member TryParse (str : string) = 
-        if String.IsNullOrEmpty str then None 
+        if String.IsNullOrEmpty str then None
         else
-            let name = Regex("^(?<name>[a-zA-Z]+)").Match(str).Value
-
+            let getName fromList =
+                match fromList with
+                | AlphaNumeric(a)::_ -> a
+                | _::AlphaNumeric(a)::_ -> a // fallback to 2nd
```
viktor-svub: fallback to 2nd place, for cases where users want to move from non-compliant versions like `1.2.3.4-beta1` to valid semver2 like `1.2.3-4.beta.1` to keep similar precedence rules

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -47,42 +53,72 @@ type PreRelease =
...
             let parse segment =
                 match bigint.TryParse segment with
-                | true, bint -> Numeric bint
-                | false, _ -> AlphaNumeric segment
+                | true, number when number >= 0I -> Numeric number
+                | _ -> AlphaNumeric segment
+                
+            let notEmpty = StringSplitOptions.RemoveEmptyEntries
+            let name, values = 
+                match str.Split([|'.'|],notEmpty) with
```
viktor-svub: heuristics: versions with dots in prerelease are semver2, which allows hyphen as part of identifiers; 
only when having no dots we try to interpret the embedded numbers and separate "prerelease name" via regex

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -47,42 +53,72 @@ type PreRelease =
...
-        | :? PreRelease as y -> x.Origin = y.Origin
+        | :? PreRelease as y -> x.Equals(y)
         | _ -> false
-
+        
     override x.ToString() = x.Origin
     
     override x.GetHashCode() = hash x.Origin
+    
+    member x.CompareTo(yobj) =
```
viktor-svub: changed to tail-recursive list comparison as zipOpt was slower, not debuggable, and keeping semver2 compliance was nontransparent

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -96,19 +132,18 @@ type SemVerInfo =
       /// The optional PreRelease version
       PreRelease : PreRelease option
       /// The optional build no.
-      Build : string
+      Build : bigint
```
viktor-svub: there was no case in unit or integration tests actually using string builds; though they could have been created by accident (we actually did)

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -125,40 +160,83 @@ type SemVerInfo =
     
     member x.AsString
         with get() = x.ToString()
+        
+    member x.Equals(y) =
+        x.Major = y.Major && x.Minor = y.Minor && x.Patch = y.Patch && x.Build = y.Build && x.PreRelease = y.PreRelease
```
viktor-svub: comparing BuildMetaData is breaking semver2 compliance, and was not used anywhere

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -125,40 +160,83 @@ type SemVerInfo =
...
-        | :? SemVerInfo as y -> 
-            x.Major = y.Major && x.Minor = y.Minor && x.Patch = y.Patch && x.PreRelease = y.PreRelease && x.Build = y.Build && x.BuildMetaData = y.BuildMetaData 
+        | :? SemVerInfo as y -> x.Equals(y)
         | _ -> false
     
-    override x.GetHashCode() = hash (x.Minor, x.Minor, x.Patch, x.PreRelease, x.Build)
+    override x.GetHashCode() = hash (x.Minor, x.Minor, x.Patch, x.Build, x.PreRelease)
+    
+    member x.CompareTo(y) =
+        let comparison =
```
viktor-svub: comparing each part only once and using the result is more transparent (and faster)

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -177,55 +255,59 @@ module SemVer =
...
+                    | n when n = version.Length - 1 -> ""
+                    | n -> 
+                        let content = validContent.Match(version.Substring(n + 1))
+                        if content.Success then content.Value else ""
+
+                let fragments = expand [ '.'; '-' ] versionStr
+                /// matches over list of the version fragments *and* delimiters
+                let major, minor, patch, revision, suffix =
+                    match fragments with
+                    | (Int M)::"."::(Int m)::"."::(Int p)::"."::(Big b)::tail -> M, m, p, b, tail
```
viktor-svub: match including delimiters allows to parse as much of valid version parts as possible, and processing the rest as prerelease segments

src/Paket.Core/Versioning/VersionRange.fs
```diff
@@ -31,26 +32,92 @@ type VersionRange =
...
         match other, this with
         | Minimum v1, Minimum v2 when v1 <= v2 -> true
         | Minimum v1, Specific v2 when v1 <= v2 -> true
+        | Minimum v1, Range(_, min2, max2, _) when v1 <= min2 && v1 <= max2 -> true
         | Specific v1, Specific v2 when v1 = v2 -> true
-        | Range(_, min1, max1, _), Specific v2 when min1 <= v2 && max1 >= v2 -> true
         | GreaterThan v1, GreaterThan v2 when v1 < v2 -> true
         | GreaterThan v1, Specific v2 when v1 < v2 -> true
+        | GreaterThan v1, Range(_, min2, max2, _) when v1 < min2 && v1 < max2 -> true
+        | Range(lower, min1, max1, upper), Specific v2 ->
```
viktor-svub: covering more cases allows faster resolver passes, and may prevent looping in some cases

src/Paket.Core/Versioning/VersionRange.fs
```diff
@@ -31,26 +32,92 @@ type VersionRange =
...
+                | VersionRangeBound.Excluding, VersionRangeBound.Including -> min1 < min2
+                | _ -> min1 <= min2
+            let upperMatch = 
+                match upto1, upto2 with
+                | VersionRangeBound.Including, VersionRangeBound.Excluding -> max2 < max1
+                | _ -> max2 <= max1
+            lowerMatch && upperMatch
         | _ -> false
 
+    member this.GetPreReleaseStatus =
```
viktor-svub: unified prerelease status from the constraints' parts

src/Paket.Core/Versioning/VersionRange.fs
```diff
@@ -161,20 +228,20 @@ type VersionRequirement =
     static member Parse text =
         if String.IsNullOrWhiteSpace text || text = "null" then VersionRequirement.AllReleases else
 
-        let prereleases = ref PreReleaseStatus.No
+        let prereleases = List<string>()
```
viktor-svub: packages with transitive dependencies like `>= 1.0.1-beta.1 < 1.0.1` -- which are valid and can be produced automatically -- ended up with No prerelease status, as it was overwritten by the second part
also, allowing All for transitive dependencies of Concrete prerelease packages ended up with unresolvable graphs in some cases (diamond dependencies on named prerelease)

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -19,26 +20,31 @@ type PreReleaseSegment =
     | AlphaNumeric of string
     | Numeric of bigint
 
+    member x.CompareTo(y) =
+        match x, y with
+        | AlphaNumeric a, AlphaNumeric b -> compare a b
+        | Numeric a, Numeric b -> compare a b
+        | AlphaNumeric a, Numeric b -> 1
+        | Numeric a , AlphaNumeric b -> -1
```
baronfel: good move, much cleaner to make this logic part of the type itself 👍

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -19,26 +20,31 @@ type PreReleaseSegment =
...
-                | AlphaNumeric a, Numeric b -> 1
-                | Numeric a , AlphaNumeric b -> -1
+            | :? PreReleaseSegment as y -> x.CompareTo(y)                
             | _ -> invalidArg "yobj" "can't compare to other types of objects."
-
+            
     override x.GetHashCode() = hash x
+    
+    member x.Equals(y) =
+        match x, y with
```
baronfel: again, good thinking

src/Paket.Core/Versioning/SemVer.fs
baronfel: Nice change, can we remove `zipOpt` now? I should never have written that thing.

src/Paket.Core/Versioning/SemVer.fs
baronfel: I think the compelling case for `string` here instead of `bigint` is the case of embedding commit hashes as the build metadata. The spec (section 10) says that build metadata can have the following form:
> ...Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]...

and so I think we must allow string here.

src/Paket.Core/Versioning/SemVer.fs
baronfel: good catch, that never should have been there

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -125,40 +160,83 @@ type SemVerInfo =
...
 
     override x.Equals(yobj) = 
         match yobj with
-        | :? SemVerInfo as y -> 
-            x.Major = y.Major && x.Minor = y.Minor && x.Patch = y.Patch && x.PreRelease = y.PreRelease && x.Build = y.Build && x.BuildMetaData = y.BuildMetaData 
+        | :? SemVerInfo as y -> x.Equals(y)
         | _ -> false
     
-    override x.GetHashCode() = hash (x.Minor, x.Minor, x.Patch, x.PreRelease, x.Build)
+    override x.GetHashCode() = hash (x.Minor, x.Minor, x.Patch, x.Build, x.PreRelease)
```
baronfel: why the change in order here?

src/Paket.Core/Versioning/SemVer.fs
viktor-svub: As far as I understand it, build metadata is separate field, a semver compliant one, placed after a `+` sign, while the `Build` field is the 4th optional, non-semver, but NuGet supported, numeric segment, separated by dot, and having no meaningful comparison semantics for strings.

src/Paket.Core/Dependencies/DependenciesFileParser.fs
```diff
@@ -25,29 +26,36 @@ module DependenciesFileParser =
...
+        let inline isNumeric item = 
+            match BigInteger.TryParse item with
+            | true, number -> Some(number)
+            | false, _ -> None
+            
+        let mutable fragments = 
+            ((minimum.AsString.Split '-').[0].Split '.')
+            |> Array.map isNumeric 
+            |> Array.takeWhile (fun i -> i.IsSome) 
+            |> Array.choose (fun i -> i)
```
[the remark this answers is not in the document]
viktor-svub: thanks for the catch! I'll look into that

src/Paket.Core/Versioning/SemVer.fs
viktor-svub: Build is a number (or simple string) and may be part of release versions like `1.2.3.4` so the comparison preference should be kept regardless if there is PreRelease present or not; plus, this way it's faster :)

src/Paket.Core/Versioning/SemVer.fs
baronfel: ah, of course you are correct

src/Paket.Core/Versioning/SemVer.fs
baronfel: yes, correct again :)

src/Paket.Core/Dependencies/DependenciesFileParser.fs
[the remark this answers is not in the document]
viktor-svub: I believe the shorter version has different results - the intent is full stop on first non-numeric segment, not simply taking all the numeric ones

matthid:
This looks very good!

Just some words of warning: This is a delicate part of Paket with high potential of breaking stuff (especially since there are a lot of packages out there which are not actually semver compatible) .
Therefore I suggest that this at least needs a green CI (Just to explain: We have a long history of broken CI and ignoring them for PRs on some occasions, if the failures look unrelated :) )

Also feel free to send the same PR to FAKE as we have a `Fake.Core.SemVer` package with basically the same logic, and it would be nice to have the proper author on the history :)

If I find the time before this thing is merged I might review this more in-depth, thanks for sending this.

viktor-svub:
Thank you! And I definitely agree this needs a thorough review & passing CI :)
I'm aware of the risks, as this was created to resolve our very real production issues, and mainly to bridge the gap between semver2 and non-semver cases, supporting more of both.

forki:
I didn't yet find time to look at this, but for quite some time I wanted to to have an integration test which checks ALL packages on nuget and verify we can understand the version no.

I know it's a big thing to ask, but tbh it would increase my confidence into this change dramatically if you could try to come up with such a test.

viktor-svub:
Actually that's a quite good idea, it may just take few more hours to implement :)
I'll do that separately, to see the tests' results before the changes, and merge into this branch/PR later.

src/Paket.Core/Versioning/SemVer.fs
BillHally: Is this deliberately not including x.Major in the hash (though it does include x.Minor twice)?

viktor-svub: oh my, that's a very bad typo, thanks for the catch :)

src/Paket.Core/Versioning/SemVer.fs
```diff
@@ -177,55 +244,59 @@ module SemVer =
                 for s in version.Split([|'.'|]) do
```
matthid: Can you change the comment of this function.
I'm not sure if this PR changed this but the line

`///     parse "1.2.3-alpha002" > parse "1.2.3-alpha1"   // true`

is no longer correct.

forki:
Really really cool.

forki:
Should we merge and release as prerelease for couple of days?

matthid:
@forki can you elaborate what you meant with the “ALL packages“ test and how it would work. I don’t think we can download all versions from nuget within a single test run. But we can generate a list of test-cases (but it will be way too long to review and the F# compiler will probably choke as well).
Or I just don’t get the idea...

matthid:
We can maybe filter out “uninteresting” versions via regex.

viktor-svub:
As far as I understood and agreed on the "all of nuget" test idea, it should be possible to use the [NuGetv3 Catalog](https://docs.microsoft.com/en-us/nuget/api/catalog-resource) to get all versions of all packages. 
If they were in gallery order (need to confirm this), I could check our parsing & comparison is able to 1/ read all versions, 2/ return the latest release, and 3/ return the latest prerelease. 
It will take me few days to do this in F# by myself; Do we perchance already have some helper in Paket or FAKE for catalog access?

forki:
Yes my naive idea was to scrape all package versions of all packages.

Am 10.02.2018 10:57 schrieb "Viktor Svub" <notifications@github.com>:

src/Paket.Core/Versioning/SemVer.fs
viktor-svub: I'm conflicted about this :) the change is side effect of preferring SemVer2 ordering, but I really want to keep compatibility with SemVer1 workarounds -- not even M$ packages (netcore) use v2 format and rules in all cases, and there are many companies (incl. ours) having _thousands_ of improperly versioned packages in their repos. 
In this case, real-world semantics should win, I'd expect -- `alpha002` should be bigger than `alpha1`, as the zero-padded version was created as workaround of broken ordering  above `alpha9`, maybe after `alpha1` already existed. 
Luckily, the SemVer1/2 special case handler is one of the simpler parts of the code :)

matthid:
Yes the only thing we need to watch out for is that afaik NuGet will respond with different results depending on the client and the feed :)

forki:
@viktor-svub could you please take a look at https://github.com/fsprojects/Paket/issues/3032#issuecomment-364398018 I think we interpret that range wrongly. I fixed the issue with lowest_matching but the problem around twiddle-Waka is still open

matthid:
One more note on this:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Numeric identifiers MUST NOT include leading zeroes. Pre-release versions have a lower precedence than the associated normal version. A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version. Examples: 1.0.0-alpha, 1.0.0-alpha.1, 1.0.0-0.3.7, 1.0.0-x.7.z.92.

This means in fact that 
```fsharp
> parseAndPrint "1.0.01-pre-001";;
val it : string =
  "{ Major: 1, Minor: 0, Patch: 1, Prerelease: Some {Origin = "pre-001";
      Name = "pre";
      Values = [AlphaNumeric "pre"; Numeric 1];}, Build: 0, BuildMetadata: , Original: Some "1.0.01-pre-001" }"
```
Is not the correct result. Apparently items are only split by `.` and therefore we have `AlphaNumeric "pre-001"` (or `AlphaNumeric "pre"`; `AlphaNumeric "001"`). 

Leading zeroes in numerical prerelease identifiers like `1.0.0-alpha.01` are not allowed, but I'm not too concerned if we allowed that.

Maybe we could have a property on the "SemVerInfo` indicating whether the given version is fully semver compliant.

@viktor-svub What is your opinion on this?

viktor-svub:
@forki 
>  ... could you please take a look at #3032 (comment)
I'll play with this a bit more then -- we are creating constraints like this in automated scripts, it happens when you have a 4-segment package version like `Package-1.2.3.4` and want to have semver-compliant constraint without allowing history versions: `Package ~> 1.2 >= LOCKEDVERSION`. 
We should adjust the lower bound accordingly, and/or throw, if the second constraint part violates the first (twiddle) one.

src/Paket.Core/Versioning/SemVer.fs
matthid: Yes I agree, but this is about PRERELEASE only. And SemVer2 has been out for some time now. You should not depend on PRERELEASEs too long you know :). Can you give examples of currently used packages where this particular instance is a problem?

Personally I would have no problem with releasing this and look where and if it hurts someone and if we cannot figure out a workaround we can still revert and look for a solution.

matthid: I just verified that this PR did not change this behavior:
```
> oldParse "1.2.3-alpha002" > oldParse "1.2.3-alpha1";;
val it : bool = false
```

Therefore this change is already live and probably nothing to worry about.

viktor-svub:
@matthid I agree we should be SemVer2 compliant where feasible & useful, but the scheme pretty much ignores history -- packages like `xunit.core 2.3.0-beta3-build3705` are and will be in use for long time, and should be resolved in expected order.

I'd like Paket to be actively helpful if a package switches to the currently proper scheme -- semver2 today, but even NuGet.org does not _enforce_ semver (they even extend it), and we can safely assume there will be another scheme in foreseeable future (semver3? nuget4?) :) 

In these cases the ordering should as close to what was intended -> `xunit.core 2.3.0-beta3-build3705` may be followed by `xunit.core 2.3.0-beta.4.build.4567` and they should compare to each other. 

My take in this case is assuming a version without dots in prerelease _can_ be interpreted as semver1-like, where dots were not allowed and hyphens _were_ used to separate "segments", though formally there was only a single one. 

We could keep & extend the existing heuristic to facilitate for existing versions, support new ones, and maybe allowing users to use hints in `paket.dependencies` to ensure the version is interpreted strictly.

matthid:
@viktor-svub One important point to consider is that we don't want a different behavior than NuGet itself. afaik they will use a semver compliant ordering when using a recent client and V3.

While I agree with your reasoning I'm not sure if will matter that much in practice. For example consider your example `xunit.core 2.3.0-beta3-build3705` and `xunit.core 2.3.0-beta.4.build.4567`. In the current implementation you propose it is `2.3.0-beta3-build3705` > `xunit.core 2.3.0-beta.4.build.4567` already. And if you meant `xunit.core 2.3.0-beta4.build.4567` (note the missing dot) than it doesn't matter how we handle `-` or `.` the one or the other way as the result will be the same. So I'd argue that the change I propose will not change a lot in practice.

src/Paket.Core/Versioning/SemVer.fs
viktor-svub: As much as I like there currently is a widely accepted semver2, it does not cover all of use cases, and for some companies, named "prereleases" are the majority of their packages, being produced & consumed by named feature/bugfix branches. 

Similar approach can be seen in bigger companies, where there is simply no feasible fast release cycle (netcore could stay _years_ in prerelease). 

From my PoV, removing the old workaround of parsing numbers off alphanumeric segments makes transition no newer scheme harder -- I'd like to safely assume that `1.2.3-beta001` is less-than `1.2.3-beta.2` :)

matthid:
I'd even go further than that: Let's assume xunit-core reaches the place where their build number has 5 digits and therefore my change would introduce a subtle difference. I think that package will not be recognized as newer by NuGet and they will immediately re-release with a higher previous segment.

viktor-svub:
@matthid You are right I should not try to workaround everything :) 

I'm just really concerned about backwards compatibility -- AFAIK majority of NuGet packages still use semver1-like versions but with embedded zero-padded numbers to work around the lack of better solution at that time -- in may cases, they cannot be simply upgraded or replaced, and they should be interpreted as intended at the time of their creation. 

As for semver2 compliance -- I'd assume any version actually using the new possibility of multiple dot-separated segments is created with knowledge of the semantics and should not be re-interpreted. 

But we can harden/soften the rules as needed, and we should keep a foot in the door for next semver :)

matthid:
> in may cases, they cannot be simply upgraded or replaced, and they should be interpreted as intended at the time of their creation.

Yes, in fact I think doing the SemVer-compliant thing is actually the correct thing. Afaik NuGet always recognized the prerelease stuff as simple string (which was surprising and therefore the workarounds with zero padding). Applying SemVer compliance (ie only splitting by `.`) does exactly this: Parses this thing as raw string without any interpretation.

viktor-svub:
Having a "semverX" non/compliance hint & property on `SemVerInfo` could be better solution than heuristic, and better than accidentally breaking current use cases -- Paket's selling points are its _differences_  from NuGet -- specifically its ability to interpret embedded numbers and named prereleases -- I'd believe :)

matthid:
> Paket's selling points are its differences from NuGet -- specifically its ability to interpret embedded numbers and named prereleases -- I'd believe :)

Afaik we never said we would handle version parsing or comparisons differently than NuGet - at least I never heard of it ;)
Personally I'd consider this a bug :)

Just to be sure /cc @forki :)

viktor-svub:
@matthid I never thought of it that way :D Considering the heuristic parsing was even covered by tests, it might have been a bug-became-feature, but already in wide use for long time (we surely do) :)

matthid:
Wow good job!

> It cannot be submitted -- even with paged Catalog, full sweep takes more than an hour :/
With per-page cached responses, the whole test finishes in about 40 seconds,
but the per-page gzipped cache itself is more than 70MB -- a little too much.
I'm working on cached data reduction, what would you consider an acceptable size?

Good question, I think we at least would accept the code even if it takes a while to execute we could execute it from time to time if needed.
If we in-fact want to integrate it into the CI we could reduce the data a lot by removing all "common" cases and keeping the unusual ones.
At least for the ones where the old parser failed we should add a test into the CI (probably makes sense).

By the way it would be also interesting to see where the previous and the new parser changed ordering... But I think you have done enough to get this merged already. The only thing left is about the handling of `-` in the prerelease section.

forki:
It's now red!?

viktor-svub:
Working on the CI failure -- cached catalog size was reduced to "only" 18MB, and the merged branches pass on my machine in 30 seconds -- the CI server or jobs seem to behave differently.

forki:
looks like master build is broken as well. Something weird happen https://github.com/fsprojects/Paket/pull/3038 is canary and red

forki:
https://help.appveyor.com/discussions/problems/10253-maximum-allowed-artifact-storage-size-of-1000-mb-will-be-exceeded

viktor-svub:
As for the SemVer2 compliance, I believe we should support NuGet behavior described at 
https://docs.microsoft.com/en-us/nuget/reference/package-versioning#semantic-versioning-200 
where they use similar heuristics to allow semver1 and non-semver packages as fallback for non-compliant version strings, _and_ we may want to double-check our wildcard handling. 

We should also probably check discussion and issues described in https://github.com/NuGet/Home/issues/912 as Paket's approach was mentioned as viable solution over there :)

matthid:
I'm still talking about the handling of `-` in prerelease strings as opposed to `.`. I still feel like we should treat them the semver way. I'm yet to hear a reason against that. The link https://docs.microsoft.com/en-us/nuget/reference/package-versioning#pre-release-versions just shows that they do exactly what I'm proposing:

> When resolving package references and multiple package versions differ only by suffix, NuGet chooses a version without a suffix first, then applies precedence to pre-release versions in reverse alphabetical order.

I think the issue you link is orthogonal as we still can support a minimum prefix of the semver prerelease string in the dependencies file no matter they way we use to parse the versions? 

Or maybe I just don't get what you are trying to tell me, sorry :)

If you say for example the current implementation is easier that's a fine reason to leave it as is (as I think the impact is minimal either way)

viktor-svub:
Ah, now I understand, and you are right :) Indeed, I got over-zealous changed the semver1 parser -- maybe while solving a real-world case, but a change nevertheless -- I'll try to keep that closer to original.

src/Paket.Core/Dependencies/NuGetV3.fs
```diff
@@ -277,7 +282,8 @@ let private getCatalogPageDirectory(basePath:String,item:String) =
         | _ -> Uri.EscapeDataString(item)
     let cachePath = Path.Combine(basePath,hostPath)
     let directory = new DirectoryInfo(cachePath)
-    if directory.Exists |> not then 
+    if directory.Exists |> not then
+        traceWarnfn "Create \"%s\"" directory.FullName
```
forki: Is this warning expected?

forki: Is this warning expected?

viktor-svub: it's not, as the page cache for the only code using this -- the public catalog integration test --
 _should_ be committed (it is now) and the warning will trigger only if called without the path existing

forki:
My problem is that people regularly complain here in issue tracker when
they see warnings. And this one would be very weird, since it's not very
specific and there is nothing what they can do about it.

Am 12.02.2018 19:24 schrieb "Viktor Svub" <notifications@github.com>:

viktor-svub:
Oh, I see -- I can change that to detailed warning about re/building catalog cache from scratch and expecting very long operation/delay, would that be ok?

forki:
Mhm. I'm not sure what we are changing here. Can you please explain again
what is happening? I think I'm not fully understanding yet

Am 12.02.2018 19:34 schrieb "Viktor Svub" <notifications@github.com>:

viktor-svub:
Catalog support was created just for the "all of nuget" integration test, but placed alongside the other 
NuGet V3 "services" implementations -- this function is new and not used anywhere else. 

Cursor-based implementation like https://docs.microsoft.com/en-us/nuget/api/catalog-resource#iterating-over-catalog-items would be only reliable and efficient if I used a compact and append-only representation on client side -- so I went for a more resilient and tolerant per-page cache. 

Creating the page cache takes still more than an hour though, which is where the warning is appropriate -- missing directory means the cache will be created from scratch. 

It seems this may not be a viable approach in the end -- while locally subsequent test runs take about half a minute, the CI servers do not exactly like it.

viktor-svub:
Ok, the test overhead outgrew its potential usefulness, I'm removing it from this PR. 
I will try to come up with lightweight version, in another brach, as mixing the changes got messy.

viktor-svub:
AppVeyor RunIntegrationTestsNetCore failed with 
`
1) System.Exception: Test failed on "test" "integrationtests/Paket.IntegrationTests.preview3/Paket.IntegrationTests.fsproj" ...
Command exited with code 42
`
https://ci.appveyor.com/project/paket/paket/build/0.0.1.88/job/bp3w37f1hb14rp6x#L2812

No specific test failed, and on local machine, the whole build seems to pass.

forki:
yes have that locally as well. wonder what changed!? /cc @enricosada @matthid

viktor-svub:
So far it seems that failed or successful test runs differ in the machines' performance -- the latest failed run has startup time -- and almost any subsequent measured operation -- 2-3 times longer than the successful ones, end ends with job timeout :/

forki:
@enricosada ideas?

viktor-svub:
@forki you were completely right about the public catalog verification :-) there _was_ a regression and the latest commits (semver2 allows hyphens...) are the fix with new unit-tests

As for the verification -- this branch was merged into PR #3048 which _passed_:
 https://github.com/fsprojects/Paket/pull/3048/commits/20df660a3313a698c7ff03dc8bd04ea666f55c68 shows the _intentional_ changes (lines ending with `!0.0.0` were not parse-able and interpreted as zero in the original requirement parser)

Do we still need to wait for the same verification to pass here?

forki:
I'm fine with merging if @matthid and you are saying it's "probably good" ;-)

viktor-svub:
As there is no further activity planned here at the moment, 
unless you have any objections, feel free to merge anytime :)

matthid:
Yes it probably is fine, thanks @viktor-svub for going all the way with us :)

forki:
image


^^ in the latest builds

viktor-svub:
Seems like conflict with #3040 -- I should have probably rebased/merged it immediately after this was merged. Line 6778 is especially suspicious, as it does not look like a parser issue but rather an ordering one (which I did not encounter before) _or_ the catalog being updated accidentally during the test. 
Note that updating is done through an `TestCase("...",Explicit=true)` "test" and if the attribute was ignored in the CI build runner, it would create quite interesting results.

forki:
any advice how to fix?

viktor-svub:
After looking a bit, it rather looks like the error _expected_ to appear _before_ #3030. 
It's a merge/rebase issue most likely, _or_ I might gave accidentally pushed older version of the catalog, i.e. not adapted by the #3030 -- as the catalog is saved in the "current" ordering, to detect any change i parser/order without keeping their older versions around :)

viktor-svub:
Currently it can be fixed semi-manually as the updating "test" `#3030-9 update nuget catalog from public index` does too much -- re-ordered version of current catalog without update from gallery needs only `getCatalogCursor` to `setCatalogCursor` -- there will be only changed line, no added, and the flat-text file can be diffed and reviewed manually to see if the ordering was broken or rather fixed.

viktor-svub:
Note that when the `Expected` line contained an exclamation mark and the `But was` did not, it's _likely_ a fix rather error, as `!` delimits the original version string and its normalized form _if they differ_, and non-parsed versions were interpreted as `0.0.0` in constraints -- that's what I see in the picture.

forki:
ok. please please send a PR ;-)
thank you!

viktor-svub:
I can push updated catalog after I get to my devbox, as there was already supposed to be one -- not sure how the old ordering ended up in the updated PR :)

viktor-svub:
here #3056 is the catalog update to correct ordering :)

PR #3040, opened 2018-02-11
testing semver parser over public nuget catalog

viktor-svub:
intended as verification of #3030, allowed to fail on current version


viktor-svub:
Note this _can_ be merged separately from #3030 as the test failing without the new parser is currently written to print its errors as warnings and end up "inconclusive", see 
https://ci.appveyor.com/project/paket/paket/build/0.0.1.101/job/6g3hx4ju4g5qg2xy#L2465 
and https://travis-ci.org/fsprojects/Paket/jobs/341608465#L6785 
(if the almost 20MB flat text file holding the nuget.org catalog, in the last approved version ordering, does not bother you, that is)

forki:
carzy! thanks

PR #3048, opened 2018-02-14
[CI][no-merge]verification of #3030 semver2ext with catalog

viktor-svub:


viktor-svub:
Already merged via #3030, #3040 and #3056

PR #3051, opened 2018-02-16
fix integration test

enricosada:
use `createScenarioDir` to create a clean `temp` dir for integration tests

fix integration test `#3200 info should locate paket.dependencies`.
now test can be run multiple times, temp dir is cleaned up correctly


forki:
thanks

Issue #3047, opened 2018-02-14
Paket pack failing in multi-target projects


eiriktsarpalis:
### Description

It seems like the paket pack command is failing when attempting to bundle packages in multi-target projects. 

### Repro steps

1. Do a Release build of [this project](https://github.com/eiriktsarpalis/Repro/tree/acaa51c4f220fb587782af707e9e874f65a22817).

2. From the command line, call `.paket/paket.exe pack .`

### Expected behavior

Should construct a nuget package with supplied parameters.

### Actual behavior

Fails with the following error:

```
Paket failed with
-> Unable to find Release output path node in file C:\Users\eirik.tsarpalis\Source\Repos\Repro\Library\Library.fsproj for any known platforms
```

### Known workarounds

None known.

forki:
Haha @matthid this is hunting us now.

Am 14.02.2018 17:14 schrieb "Eirik Tsarpalis" <notifications@github.com>:

eiriktsarpalis:
@forki context? I'll see if I can write a fix for this over the weekend.

eiriktsarpalis:
I have been able to make it work in a non multi-target setting.

eiriktsarpalis:
Quite a bit of unnecessary boilerplate though: https://github.com/fsprojects/Argu/blob/35ff826bb0944d06cb2763f79d4a86d48d83b643/src/Argu/paket.template

forki:
Let's just say pack command needs a bit of help. But especially for dotnet
core people used dotnet pack with success, so it wasn't our highest
priority.

Am 14.02.2018 20:06 schrieb "Eirik Tsarpalis" <notifications@github.com>:

eiriktsarpalis:
Makes sense. I had evaluated dotnet pack, but wasn't too impressed by its configurability.

matthid:
What exactly is missing in dotnet pack?

eiriktsarpalis:
Couldn't find a way to pass version numbers or customize contents, e.g. to
include DLLs from other projects.

enricosada:
@eiriktsarpalis you can include any files, in any nupkg directory (renamed too) with something similar to

```
    <Content Include="build/FSharp.NET.Sdk.props">
      <PackagePath>build/FSharp.NET.Sdk.props</PackagePath>
      <Pack>true</Pack>
    </Content>
```

You can also include output of projectreferences with a different xml thing.

version number like what? `/p:Version` and `/p:PackageVersion` can customize some.

[...]

PR #3070, opened 2018-02-24
Attempt to fix paket pack for multi-target projects

eiriktsarpalis:
This PR attempts to address the issues described in #3047


eiriktsarpalis:
This approach is too naive and is breaking current behaviour. Closing, will come back with a more conservative fix.

PR #3073, opened 2018-02-25
Fix bug when attempting to pack multi-target frameworks

eiriktsarpalis:
This provides a more conservative fix than the one attempted in #3070. Hopefully this will not result in any regressions.

It should be noted that this doesn't implement proper support for multi-target packing, but lets package authors work around the problem by explicitly specifying package contents.


eiriktsarpalis:
Closed?

forki:
Merged but rebased because I fucked up. Will be rel with next release

Am 26.02.2018 09:30 schrieb "Eirik Tsarpalis" <notifications@github.com>:

eiriktsarpalis:
Ah, thanks

PR #3092, opened 2018-02-27
Faster "hot" restore

vbfox:
Remove two slow part from doing `paket restore` when a restore isn't needed

* Setting `ServicePointManager.SecurityProtocol` is costly as it initialize a lot of .Net subsystems we don't otherwise use so I split the `Utils` module in two. As the network isn't used if everything is there the static constructor of `NetUtils` is never called.
* Make filling `supportMap` with it's initial value lazy. It's costly and not used in such restores.


forki:
mhm it broke one integration test. weird

2018-02-27 21:42 GMT+01:00 Matthias Dittrich <notifications@github.com>:

forki:
ok test failure is unrelated. it's red on https://github.com/fsprojects/Paket/pull/3090 as well

forki:
ok seems https://www.nuget.org/packages/Microsoft.Diagnostics.Tracing.EventSource.Redist/ released a package 7h ago. I will fix the test

PR #3098, opened 2018-02-28
Safeguards loading of assemblies during load script generation. 

### LINKED ISSUE #3097: Generate Load Scripts fails with BadImageFormatException on Microsoft.TestPlatform.TestHost
### Description

When the package `Microsoft.TestPlatform.TestHost` is a dependency in `paket.dependencies` file and it is tried to create a load script for that dependency paket fails with a `BadImageFormatException`.

### Repro steps

See a repro repo here: https://github.com/ChrSteinert/Paket-3097
Do a `paket install` && `paket generate-load-scripts` do trigger the bug.

### Expected behavior

Paket ignores native dll files.

### Actual behavior

Paket tries on multiple occasions to load `.dll` files of a package. [Some](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Installation/InstallProcess.fs#L300) are guarded by `try…with`, but [some](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/PaketConfigFiles/DependencyCache.fs#L101) are not. Load Script generation hits the latter one and breaks if it tries to load native dlls from (for instance) the `Microsoft.TestPlatform.TestHost` package.

### Known workarounds

Do not generate Load Scripts for those packages.

ChrSteinert:
This is a super naive approach to the problem of loading native assembies that will fail with a `BadImageFormatException`.

Fixes #3097 .
I confirmed, that this fixes the error for https://github.com/ChrSteinert/Paket-3097

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/DependencyCache.fs
+++ b/src/Paket.Core/PaketConfigFiles/DependencyCache.fs
@@ -99,13 +99,12 @@ type DependencyCache (lockFile:LockFile) =
             installModel
             |> InstallModel.getLegacyReferences (TargetProfile.SinglePlatform framework)
             |> Seq.map (fun l -> l.Path)
-            |> Seq.map (fun path -> 
+            |> Seq.choose (fun path -> 
                 try 
                     (AssemblyDefinition.ReadAssembly path, FileInfo(path)) |> Some
                 with
                 | :? BadImageFormatException -> None
             )
-            |> Seq.choose id
             |> dict
 
         getDllOrder (dllFiles.Keys |> Seq.toList)
@@ -127,12 +126,11 @@ type DependencyCache (lockFile:LockFile) =
                         sysLibs.Add sysLib |> ignore
 
             let assemblyFilePerAssemblyDef = 
-                libs |> Seq.map (fun (f:FileInfo) -> 
+                libs |> Seq.choose (fun (f:FileInfo) -> 
                     try
                         (AssemblyDefinition.ReadAssembly (f.FullName:string), f) |> Some
                     with
                     | :? BadImageFormatException -> None)
-                |> Seq.choose id
                 |> dict
 
             let assemblies = 
--- a/src/Paket.Core/PaketConfigFiles/DependencyCache.fs
+++ b/src/Paket.Core/PaketConfigFiles/DependencyCache.fs
@@ -98,7 +98,13 @@ type DependencyCache (lockFile:LockFile) =
             installModel
             |> InstallModel.getLegacyReferences (TargetProfile.SinglePlatform framework)
             |> Seq.map (fun l -> l.Path)
-            |> Seq.map (fun path -> AssemblyDefinition.ReadAssembly path, FileInfo(path))
+            |> Seq.map (fun path -> 
+                try 
+                    (AssemblyDefinition.ReadAssembly path, FileInfo(path)) |> Some
+                with
+                | _ -> None
+            )
+            |> Seq.choose id
             |> dict
 
         getDllOrder (dllFiles.Keys |> Seq.toList)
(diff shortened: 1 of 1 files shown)


src/Paket.Core/PaketConfigFiles/DependencyCache.fs
```diff
@@ -98,7 +98,13 @@ type DependencyCache (lockFile:LockFile) =
             installModel
             |> InstallModel.getLegacyReferences (TargetProfile.SinglePlatform framework)
             |> Seq.map (fun l -> l.Path)
-            |> Seq.map (fun path -> AssemblyDefinition.ReadAssembly path, FileInfo(path))
+            |> Seq.map (fun path -> 
+                try 
+                    (AssemblyDefinition.ReadAssembly path, FileInfo(path)) |> Some
+                with
+                | _ -> None
```
matthid: I don't like that pattern, can we at least catch only the relevant exception or otherwise log the error as warning?

src/Paket.Core/PaketConfigFiles/DependencyCache.fs
```diff
@@ -121,7 +127,11 @@ type DependencyCache (lockFile:LockFile) =
 
             let assemblyFilePerAssemblyDef = 
                 libs |> Seq.map (fun (f:FileInfo) -> 
-                    AssemblyDefinition.ReadAssembly (f.FullName:string), f)
+                    try
+                        (AssemblyDefinition.ReadAssembly (f.FullName:string), f) |> Some
+                    with
+                    | _ -> None)
```
matthid: same here ;)

matthid:
Thanks for taking care of this :)

ChrSteinert:
My pleasure!

src/Paket.Core/PaketConfigFiles/DependencyCache.fs
ChrSteinert: Yes, sure. Good point.
I was "copying" from https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Installation/InstallProcess.fs#L300 :)

matthid: Yes, getting rid of them is hard, because you do not know what you have catched previously ;).
A PR or adding the other case is welcome as well ;)

src/Paket.Core/PaketConfigFiles/DependencyCache.fs
```diff
@@ -98,7 +99,13 @@ type DependencyCache (lockFile:LockFile) =
             installModel
             |> InstallModel.getLegacyReferences (TargetProfile.SinglePlatform framework)
             |> Seq.map (fun l -> l.Path)
-            |> Seq.map (fun path -> AssemblyDefinition.ReadAssembly path, FileInfo(path))
+            |> Seq.map (fun path ->
```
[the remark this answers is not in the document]
matthid: One more question: What path is it here? Why is someone putting invalid references into the `/lib` folder?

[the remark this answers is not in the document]
ChrSteinert: They are invalid in the sense that they are native x64 or x86 dlls that the assembly loader might not like to load.

And yes, the direct choose is obviously a better way.

src/Paket.Core/PaketConfigFiles/DependencyCache.fs
```diff
@@ -121,7 +128,11 @@ type DependencyCache (lockFile:LockFile) =
 
             let assemblyFilePerAssemblyDef = 
                 libs |> Seq.map (fun (f:FileInfo) ->
```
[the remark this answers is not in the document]
ChrSteinert: Obviously here as well.

forki:
thanks

PR #3102, opened 2018-03-08
Clarify how Including referenced projects works

Lavinski:
Paket automatically replaces inter-project dependencies with NuGet dependencies if the dependency has its own paket.template. However, if the working directory is the project directory (as opposed to the solution directory) the project dependencies will not be added.


forki:
thanks!

PR #3107, opened 2018-03-10
don't not add duplicated references

0x53A:
HACK: workaround for https://github.com/fsprojects/Paket/issues/2811
 * DO remove FW-References which are already referenced by the user
 * DO NOT remove package references, where the user already has a reference
Example where this is relevant:
1) Pre-existing reference to the FW-Assembly System.IO.Compression
2) user adds nuget package https://www.nuget.org/packages/System.IO.Compression/

Before:
 * the old reference is kept, the new one not added
 * no feedback during compile, assembly mismatch at runtime

After:
 * new reference is added
 * duplicated reference, warning at compile time
 * msbuild automatically takes the higher version, everything is good
 * user can manually remove old reference to silence warning

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/InstallModel.fs
+++ b/src/Paket.Core/PaketConfigFiles/InstallModel.fs
@@ -784,7 +784,13 @@ module InstallModel =
 
     let filterReferences (references:string Set) (this:InstallModel) =
         this
-        |> mapCompileLibReferences (Set.filter (fun reference -> Set.contains reference.Name references |> not))
+        // HACK: workaround for https://github.com/fsprojects/Paket/issues/2811
+        //  * DO remove FW-References which are already referenced by the user
+        //  * DO NOT remove package references, where the user already has a reference
+        // Example where this is relevant:
+        // 1) Pre-existing reference to the FW-Assembly System.IO.Compression
+        // 2) user adds nuget package https://www.nuget.org/packages/System.IO.Compression/
+        //|> mapCompileLibReferences (Set.filter (fun reference -> Set.contains reference.Name references |> not))
         |> mapCompileLibFrameworkReferences (Set.filter (fun reference -> Set.contains reference.Name references |> not))
 
     let addLicense url (model: InstallModel) =
--- a/src/Paket.Core/PaketConfigFiles/InstallModel.fs
+++ b/src/Paket.Core/PaketConfigFiles/InstallModel.fs
@@ -781,16 +781,13 @@ module InstallModel =
         { this with TargetsFileFolders = targetsFileFolders }
             |> addItem targetsFiles getMsbuildFile (addTargetsFile) (fun i -> i.TargetsFileFolders)
 
-
-    let filterReferences (references:string Set) (this:InstallModel) =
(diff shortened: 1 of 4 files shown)


src/Paket.Core/PaketConfigFiles/InstallModel.fs
```diff
@@ -784,7 +784,13 @@ module InstallModel =
...
     let filterReferences (references:string Set) (this:InstallModel) =
         this
-        |> mapCompileLibReferences (Set.filter (fun reference -> Set.contains reference.Name references |> not))
+        // HACK: workaround for https://github.com/fsprojects/Paket/issues/2811
+        //  * DO remove FW-References which are already referenced by the user
+        //  * DO NOT remove package references, where the user already has a reference
+        // Example where this is relevant:
+        // 1) Pre-existing reference to the FW-Assembly System.IO.Compression
+        // 2) user adds nuget package https://www.nuget.org/packages/System.IO.Compression/
+        //|> mapCompileLibReferences (Set.filter (fun reference -> Set.contains reference.Name references |> not))
```
matthid: I'm not sure this is the place to do it. InstallModel is / or should be independent from msbuild/project file quirks

matthid:
Just some notes on this:

- I'm pretty sure we did the current behavior on purpose. It is not something introduced by accident. (I guess there exists a documentation/issue around this - it would be nice to find that again).
- I'm fine with releasing your suggestion and to look what happens
- Another "solution" would be to let paket warn when we don't add something because of an existing fraemwork reference

0x53A:
Yeah, I don't like this "solution" either.

1) imo the status quo is almost worst case - it is broken silently without any feedback. 

2) nuget does __remove and replace__ them.

3) I think most people will just ignore a warning at install time, because there are already so many.

4) that's why I would prefer to either keep old and add new, which will create a warning at compile time, which is wayyy more visible to the user ...

5) ... or just do the same as nuget and remove them.

matthid:
I'm ok either way (removing or always adding). The only thing here is that the logic needs to be moved to `ProjectFile` instead of `InstallModel` IMHO.

0x53A:
Ok, Will do in the evening. Thank you.

src/Paket.Core/PaketConfigFiles/InstallModel.fs
0x53A: ``filterReferences`` is called from ProjectFile.fs:
image

This is the only place that calls it.

forki:
done?

0x53A:
https://github.com/fsprojects/Paket/pull/3107#discussion_r174355982

forki:
@matthid ?

src/Paket.Core/PaketConfigFiles/InstallModel.fs
matthid: 2 and 3 are fine by me. Basically it's less about where it is used but that the InstallModel-API speaks for itself and is consistent in its naming (it is already hard to understand as it is)

matthid:
@0x53A Thanks :)

PR #3109, opened 2018-03-12
Extract fuseConflicts

forki:


matthid:
I think this is more dangerous as it seems because we depend on tail recursion to happen, right?

matthid:
Didn’t we try that before and failed or do I remember something else?

forki:
It's still tail recursion

Am 12.03.2018 17:31 schrieb "Matthias Dittrich" <notifications@github.com>:

Issue #3014, opened 2018-01-26
Using Git, GitHub and Http sources in PublicAPI


frelon:
Right now there is no way (that I'm aware of) to add a git/github/http dependency using the PublicAPI.

Any ideas on implementation? I tried a draft on #2966

enricosada:
hi @frelon ! awesome to have that api too

I think as said in PR that was not the place, it need a new API function, just for that.

so a specific method like https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/PublicAPI.fs#L132-L133

`AddGithub`, `AddGit`, `AddHttp`.
Best one for each type, using just one make things just more complicated (different flags/etc).

And new methods in https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/PackageManagement/AddProcess.fs#L92 for the AddProcess. that will be good, because we can reuse these to add cli command like `paket add-github`, `paket add-git` etc


I think is enough to mimic the `add`, because that will just invoke `InstallProcess.Install` who does the hard work https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/PackageManagement/AddProcess.fs#L69

we can chat in slack/etc if you need hints about the codebase

frelon:
Hi @enricosada!

Thanks for a very informative reply! I will try implementing this again with your changes in mind!

frelon:
I got a bit carried away and wanted some feedback on #3023 

I added the add-github, add-git, add-gist and add-http stubs in the CLI and started working on getting the add-github all the way to DependenciesFile.

In the test-scenario i002923 should the parsed file return the repository in DepFile.GetGroup().Packages ? or it's own .Repositories with matching RepositoryName? (which could also be used by add-git later)?

PR #3125, opened 2018-03-20
Add git source

frelon:
Work for paket-CLI and PublicAPI to be able to add git-repositories to paket.lock and paket.references files (#3014)


frelon:
now it should be possible to add git repos from the CLI using the following syntax:

`paket git add C:\some\test\repo <commit/branch/tag>`
or
`paket git add https://github.com/fsprojects/FsUnit`

forki:
thanks

PR #3145, opened 2018-04-08
One-liner paket installation script

jwosty:
I've added a handy script and documentation changes that replaces some manual steps for adding the paket bootstrapper to a project. Now, instead of using the pesky mouse & GUI to download and rename `paket.boostrapper.exe` to `.paket/paket.exe`, just copy-paste this into the terminal: `curl -fsSL https://raw.githubusercontent.com/fsprojects/Paket/web-install/install/web-install.fsx | fsharpi --quiet --readline-`.

This is such a common task for me that I've already been using a version of this script for my own use,  and I think a lot of people could benefit from it.

### DIFF
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -25,12 +25,8 @@ This guide will show you
 
 ### Downloading Paket's Bootstrapper
 
-1. Create a `.paket` directory in the root of your solution.
-1. Download the latest
-   [`paket.bootstrapper.exe`](https://github.com/fsprojects/Paket/releases/latest)
-   into that directory.
-1. Rename `.paket/paket.bootstrapper.exe` to `.paket/paket.exe`.
-   [Read more about "magic mode"](bootstrapper.html#Magic-mode).
+1. In the root directory of your solution, execute `curl -fsSL https://raw.githubusercontent.com/fsprojects/Paket/web-install/install/web-install.fsx | fsharpi --quiet --readline-`
+   * This will execute a script that downloads the latest `paket.bootstrapper.exe` and save it to `.paket/paket.exe`. [Read more about "magic mode"](bootstrapper.html#magic-mode).
 1. Commit `.paket/paket.exe` to your repository.
 1. After the first `.paket/paket.exe` invocation Paket will create a couple of
    files in `.paket` — commit those as well.
--- a/docs/content/getting-started.md
+++ b/docs/content/getting-started.md
@@ -25,7 +25,7 @@ This guide will show you
 
 ### Downloading Paket's Bootstrapper
 
-1. In the root directory of your solution, execute `curl -fsSL https://raw.githubusercontent.com/fsprojects/Paket/web-install/install/web-install.fsx | fsharpi --quiet --readline-`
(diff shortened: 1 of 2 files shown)


docs/content/getting-started.md
```diff
@@ -25,12 +25,8 @@ This guide will show you
 
 ### Downloading Paket's Bootstrapper
 
-1. Create a `.paket` directory in the root of your solution.
-1. Download the latest
-   [`paket.bootstrapper.exe`](https://github.com/fsprojects/Paket/releases/latest)
-   into that directory.
-1. Rename `.paket/paket.bootstrapper.exe` to `.paket/paket.exe`.
-   [Read more about "magic mode"](bootstrapper.html#Magic-mode).
+1. In the root directory of your solution, execute `curl -fsSL https://raw.githubusercontent.com/fsprojects/Paket/master/install/web-install.fsx | fsharpi --quiet --readline-`
```
jwosty: This URL of course assumes this PR is merged and the script exists, so if you want to test it out, use this instead:

    curl -fsSL https://raw.githubusercontent.com/jwosty/Paket/web-install/install/web-install.fsx | fsharpi --quiet --readline-

jwosty:
@forki what's your opinion on this?

docs/content/getting-started.md
forki: image

^^ that's why we never did that

matthid:
also fsharpi might not be available at all

docs/content/getting-started.md
jwosty: Would a PowerShell script suffice for Windows?

jwosty:
@matthid would it not be safe to assume that in most use cases for this (F# developers?), fsi would be available? Maybe then for those cases where it's not available, the instructions would include the more manual way of downloading it yourself.

forki:
Most users are C# users and no Powershell is not enabled on all windows
systems. It's unfortunately really a hard problem that we could not solve
for years

John Wostenberg <notifications@github.com> schrieb am Sa., 14. Apr. 2018,
08:33:

matthid:
> would it not be safe to assume that in most use cases for this (F# developers?), fsi would be available? 

sadly not. I'd not depend on fsharpi/fsi at all.

> Most users are C# users and no Powershell is not enabled 

@forki Well chocolatey just enable what they need in their "one-line" install, see https://chocolatey.org/install

forki:
Chocolatey is a even more niche requirement.

Matthias Dittrich <notifications@github.com> schrieb am Sa., 14. Apr. 2018,
11:14:

matthid:
@forki I'm saying you should look at how they bootstrap ;) not at chocolatey itself :)

jwosty:
Well would it work to have the quick-install script available for those that have powershell? It's not something that the user would _need_; they could still do it manually, and it would be a nice addition for those that have it. Or is it possible to write this in native Windows shell commands (cmd.exe)? As for non-Windows users, a bash script could be written as well. So the instructions would be 1 of 3 options:

1. On Windows, if you have powershell - copy-paste: `powershell one-liner here`
2. On OSX/Linux - copy-paste: `bash one-liner here`
3. Manual download: etc etc etc

Of course it'd be impossible to cater to all users with this kind of thing, but I think covering some or most of them would be a nice thing here :)

Thoughts?

PR #3146, opened 2018-04-10
support monoandroid version 8.1

kfrie:


kfrie:
I was also deliberately triggering the CI pipeline out of curiosity just to see how it works with Travis and AppVeyor and the different enviroments. So I guess the failure means I will also have to fix the integration tests 😄

kfrie:
It's official. I am too stupid to contribute 🤣 
Although I really do not know how those remaining failures are related to monoandroid81

PR #3159, opened 2018-04-18
Keeping direct dependencies distinct from external locks

viktor-svub:
just cosmetic and debuggability improvements, as we move to use the external_lock feature :+1: 
+ introduced new PackageRequirementSource to track external lock files referenced from dependency files
+ updated code relying on the union to behave correctly with the new type -- currently equal to normal dependency file, except for logging


forki:
May I ask for what purpose you use external_lock? Azure functions?

Viktor Svub <notifications@github.com> schrieb am Mi., 18. Apr. 2018, 09:57:

viktor-svub:
Not yet -- currently I plan to use this internally to enforce consistent external dependencies across our builds. We create lower hundreds of both interim and production components, and they use some external dependencies which we want to update in controlled way, and still keep them part of the resolve graph, to enforce compatibility with the unfixed ones.

forki:
thanks!

PR #3161, opened 2018-04-19
Fix license_download docs

### LINKED ISSUE #3148: paket.dependencies download_license does not work
### Description
https://fsprojects.github.io/Paket/dependencies-file.html#License-download describes how to make paket download the licenses for nuget packages.
However, paket does not seem to understand that configuration setting.

### Repro steps
This is in my paket.dependencies file:
`source https://api.nuget.org/v3/index.json
download_license: true
nuget Automapper`

when I run '.paket\paket.exe install', paket returns the following error:
`Performance:
 - Runtime: 893 milliseconds
Paket failed with
-> Error in paket.dependencies line 7
     Unrecognized token: download_license: true`

### Expected behavior
I would expect that the licenses of the packages are downloaded.

### Actual behavior
Paket returns an error and does not download anything.

### Known workarounds
paket runs normally when not using that configuration setting, but then I do not get license files.

inosik:
The wrong flag was also used in some tests.

Fixes #3148.

### DIFF
--- a/docs/content/dependencies-file.md
+++ b/docs/content/dependencies-file.md
@@ -314,13 +314,13 @@ The global option may be
 
 ### License download
 
-If you want paket to download licenses automatically you can use the `download_license` modifier. It is disabled by default.
+If you want paket to download licenses automatically you can use the `license_download` modifier. It is disabled by default.
 
 ```paket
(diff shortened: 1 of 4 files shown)


forki:
thx

PR #3169, opened 2018-04-26
Make shasum work when the path has spaces

dgchurchill:


forki:
thanks

dgchurchill:
No worries, thanks for all the work you do maintaining Paket!

PR #3184, opened 2018-05-04
extended semver2 support in paket.template files

viktor-svub:
Attempt at supporting half-open dependency constraints with minimum from current or locked version, but open to semver-declared compatibility. 

Example of intended usage would be `My.Package ~> LOCKED:Minor >= LOCKEDVERSION`. 

- Numeric syntax is "number of segments used", negative counts non-zero segments from end. 
- Unit-testing is done by extending existing scenarios via TestCase, further may be added. 
- Combining with group-binding syntax is expected and supported, like: 

>`LOCKED-NetCore:Patch` -or- `LOCKED:Patch-NetCore`
>`LOCKED-NetCore:[3]` -or- `LOCKED:[3]-NetCore`
>`LOCKED-NetCore:[-1]` -or- `LOCKED:[-1]-NetCore`

As side effect, `paket.template` dependencies can now support inline and full-line comments.
Locked-version dependencies without group bind are assumed to be "Main" and print warnings when resolved from different group.

### DIFF
--- a/docs/content/template-files.md
+++ b/docs/content/template-files.md
@@ -238,6 +238,36 @@ dependencies
   FSharp.Core >= LOCKEDVERSION-NetStandard
 ```
 
+`LOCKEDVERSION` and `CURRENTVERSION` support using partial constraints,
+to allow transitive dependencies on specific SemVer compatibility level.
+This allows to create permissive pessimistic constraints automatically.
+
+Use only `LOCKED` or `CURRENT` and semicolon-delimited name of the latest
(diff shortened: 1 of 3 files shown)


forki:
do we need to update docs?

viktor-svub:
I can add docs change, if there are no expected changes in the proposed syntax

forki:
Yes please

Viktor Svub <notifications@github.com> schrieb am Do., 24. Mai 2018, 18:19:

viktor-svub:
Docs updated, feel free to do any fixups - English not being my first language results in some creative contraptions at times :)

PR #3192, opened 2018-05-08
PAKET_DETAILED_ERRORS to print inner stack traces

ninjarobot:
Currently doesn't give the inner stack traces when asking for detailed errors.


forki:
Thx

PR #3208, opened 2018-05-17
Support pack of global tools (`PackAsTool`)

enricosada:
require .net sdk >= 2.1.300

using `PackAsTool` with `Pack` target, allow to create tools (for .net core are FDD)

this PR:

- remove the deps from generated nuspec
- set the deps assemblies so are added to nupkg

if is just a console app (no p2p, deps are ok).

```
dotnet new console -lang f#
dotnet pack /p:PackAsTool=true
```

if you want, just put the nupkg in a reachable feed (local is ok), and do `dotnet tool install mytool -g` (global) or `dotnet tool install mytool --tool-path "%CD%\dir_with_tools"` (to put tools in `dir_with_tools`, so `dir_with_tools\mytool` will work)

if is a console app with p2p, put the `PackAsTool` property inside the fsproj, otherwise the p2p deps are not added in the nupkg.

NOTE if we modify the condition to check the type (lib vs exe), will be possibile to allow also `/p:PackAsTool=true` with p2p, but i dont care atm

```
dotnet new console -lang f#
# put the PackAsTool property in fsproj
dotnet pack
```

Tests are disabled atm (but works), i'll upgrade everything to 2.1.300-rc1 with #3183 so can be run

### DIFF
--- a/integrationtests/Paket.IntegrationTests/PackSpecs.fs
+++ b/integrationtests/Paket.IntegrationTests/PackSpecs.fs
@@ -616,7 +616,7 @@ let ``#4003 dotnet pack of a global tool with p2p``() =
     directPaket ("restore") scenario
     |> ignore
 
-    directDotnet true (sprintf "pack tool1 -o \"%s\" /p:PackAsTool=true /bl" outPath) rootPath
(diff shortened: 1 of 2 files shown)


enricosada:
@forki @matthid ready to review!

PR #3213, opened 2018-05-18
test ci

enricosada:


enricosada:
the convert test failure are caused by https://github.com/fsprojects/Paket/commit/833b36c13a37d13b5503ff3c36ebf9de4f526a22

i think the `packages.config` has a regression. use `>=` instead of `==`
also i need to repro, but i seen usage of `[1.2.3]` instead on `1.2.3` syntax (is not always supported by older nugets)

enricosada:
something like

> `Error occurred while restoring NuGet packages: An error occurred while reading file 'C:\Users\XXX\Documents\Projects\XXX.ComponentTests\packages.config': Invalid package version for package id 'xunit.runner.visualstudio': '[2.3.1]'`

instead of 

> it should be   `<package id="xunit.runner.visualstudio" version="2.3.1" />`

forki:
Is this ready?

PR #3214, opened 2018-05-19
No longer search for CredentialProviders on an empty path.

ninjarobot:
In some unusual scenarios like a paket mkbundle, `Path.GetDirectoryName` on an `Assembly.Location` returns an empty string.  When that happens, `Directory.EnumerateFiles` will eventually call `Path.InsecureGetFullPath` on mono, resulting in this error:

```
System.ArgumentException: The specified path is not of a legal form (empty).
```

coming from [here](https://github.com/mono/mono/blob/master/mcs/class/corlib/System.IO/Path.cs#L367).

Switching to `String.IsNullOrEmpty` addresses that issue.


forki:
thx

PR #3244, opened 2018-06-08
Fixes for dotnet SDK fallback folder support

### LINKED ISSUE #3243: Writing to dotnet folder?
### Description

Paket 5.172.0 is trying to write to my dotnet SDK folder on OSX:
>Error: System.UnauthorizedAccessException: Access to the path "/usr/local/share/dotnet/sdk/NuGetFallbackFolder/system.io.filesystem/4.3.0/paket-installmodel.cache" is denied.

Why would paket try to write to my root-only folder?

### Repro steps

`paket update dotnet-fable`

### Expected behavior

Shouldn't try/complain.

### Actual behavior

Stack trace for some, but not all packages.

### Known workarounds

It doesn't seem to be a critical failure, it's just complaining on every attempt at package manipulation.

csmager:
There are two small issues so far.

First, it seems in `NuGet.GetContent` we store a `paket-installmodel.cache` file that has the folder structure for performance reasons. When we try to do this in the fallback folder (which is normally read-only as it's in program files or equivalent), we get an exception. 

All I've come up with is ignoring access exceptions here. Any better ideas, I'm all ears. This fixes #3243.

Second, as part of the install process we copy package 'content' folders. This does [its own search for the package folders](https://github.com/fsprojects/Paket/blob/3d9e4b5110d4baafa2e716b402e6216a2729688f/src/Paket.Core/Installation/InstallProcess.fs#L36) and didn't handle the new fallback location in the case we have `storage:none`. 

The simplest fix was to add a check in that location.. It does occur to me that we could/should know all the package locations when they're extracted earlier in the install process, so it may be better to ensure the path is returned and to use that rather than searching again. I can investigate that.

### DIFF
--- a/src/Paket.Core/Dependencies/NuGet.fs
+++ b/src/Paket.Core/Dependencies/NuGet.fs
@@ -760,14 +760,12 @@ let private downloadAndExtractPackage(alternativeProjectRoot, root, isLocalOverr
     let nupkgName = packageName.ToString() + "." + version.ToString() + ".nupkg"
     let normalizedNupkgName = NuGetCache.GetPackageFileName packageName version
     let configResolved = config.Resolve root groupName packageName version includeVersionInPath
-    let targetFileName, isTargetInFallbackFolder =
+    let targetFileName =
         if not isLocalOverride then
-            match NuGetCache.TryGetFallbackNupkg None packageName version with
-            | Some fileName -> fileName, true
-            | None -> NuGetCache.GetTargetUserNupkg packageName version, false
+            NuGetCache.GetTargetUserNupkg packageName version
         else
             match configResolved.Path with
-            | Some p -> Path.Combine(p, nupkgName), false
+            | Some p -> Path.Combine(p, nupkgName)
             | None -> failwithf "paket.local in combination with storage:none is not supported"
 
     if isLocalOverride && not force then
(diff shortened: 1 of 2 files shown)


matthid:
> First, it seems in NuGet.GetContent we store a paket-installmodel.cache file that has the folder structure for performance reasons. When we try to do this in the fallback folder (which is normally read-only as it's in program files or equivalent), we get an exception.

Shouldn't we just re-route those requests to the regular package directory? This way we still get the performance and can read/write.

> It does occur to me that we could/should know all the package locations when they're extracted earlier in the install process, so it may be better to ensure the path is returned and to use that rather than searching again. I can investigate that.

Yes if possible we should track that through the code-base. In the future we might search for other locations as well. The current `assumption` in the code-base is that `ResolvedPackagesFolder` is all you need to find the package. This is no longer true.


Also, it would be nice to have tests for these two error conditions.

csmager:
> Shouldn't we just re-route those requests to the regular package directory? This way we still get the performance and can read/write.
The original change skipped copying the packages from the fallback folder and linked directly to them, which gets us better performance on restore - I'd have to measure how much. But yes, we could revert to do that copying and then everything is always in the user's NuGet cache folder.

> Also, it would be nice to have tests for these two error conditions.
Agreed. I'd like to work a bit more on this, just wanted to open something that would fix the problems in case you need to make another release. Though you could revert the previous changes, I guess!

matthid:
> The original change skipped copying the packages from the fallback folder and linked directly to them, which gets us better performance on restore with an empty cache - I'd have to measure how much. But yes, we could revert to do that copying and then everything is always in the user's NuGet cache folder.

Actually, I was talking about only the caching file. But thinking about it - if it doesn't hurt perf too much (and I would assume it wont) - your suggestion would fix both issues...

csmager:
Yep, it would. I'll check it out and see what the performance diff is and we'll go from there.

csmager:
It adds a little time for Disk IO, but not too significant. Probably best to go with that for now and consider any further performance optimisation down the line.

I've now reverted most of the earlier changes and added something to copy packages from the fallback folder if they exist there, similar to what we do for caches.

forki:
@matthid please release it when you think it's good

Charles Mager <notifications@github.com> schrieb am Fr., 8. Juni 2018,
23:47:

matthid:
Ok lets see if 5.172.1 works :)

PR #3248, opened 2018-06-12
Correct mapping between .NET Framework and .NET Standard versions

TeaDrivenDev:
Paket's mapping of .NET Framework to .NET Standard versions is currently incorrect according to https://docs.microsoft.com/en-us/dotnet/standard/net-standard, resulting in the inability to use .NET Standard 2.0 packages in some projects that should be able to use them.


matthid:
Sadly the table you linked is a lie. we have a huge discussion around that somewhere. Can you try to find and link it (I'm on mobile right now)

0x53A:
The (simplified) issue is that if you are still on VS2015, then netstandard 2 <-> netfx4.6.1  does not work.

The best solution would be a per-group setting which enables this, but that would probably be a lot of effort to route the flag correctly through all layers.

TeaDrivenDev:
I hate .NET Core and everything that came with it.

The problem I'm trying to get around here is that as far as I can see, it is currently impossible to use packages that are .NET Standard 2.0 only (such as EF Core) from a .NET 4.6.1 project using Paket (which works fine with NuGet), because it will not generate the appropriate condition in the project file. Is there any other way around that?

0x53A:
At the moment? No. You could use nuget *shrough*.

forki:
Well, we would probably accept a group setting. But this is really dangerous area, there are so many corner cases. And as @matthid said, that table is a lie and should be deleted. I guess the only real solution is to use net471

TeaDrivenDev:
Sorry, that is not a "solution". We don't have that choice. It would work technically, of course, but it's not a solution.

forki:
Well, the other option is that group setting. Because implementation is
breaking. Unfortunately

TeaDrivenDev <notifications@github.com> schrieb am Di., 12. Juni 2018,
15:19:

matthid:
We would definitely accept a well tested implementation ;)...

0x53A:
Just dropping https://github.com/dotnet/docs/issues/5915 here.

One important part:

> 4.7.1 still needs some facades. Only 4.7.2 has all the assemblies in-box that are needed [...]

0x53A:
And I'd love a solution, because I'll shortly (hopefully in one or two weeks) be in the same boat that I need to target NET Standard 2 from NET 4.6.1.

TeaDrivenDev:
I have at least started looking into (or rather at) this. Gonna be a few years.

0x53A:
I also took a look, and routing it through really sucks.

Can we define that this flag is global per dependencies file (NOT per group), then stash it in a global mutable variable somewhere?

TeaDrivenDev:
Yeah, routing it through was what made me stop and look for something easier to practice on first, because this essentially requires understanding the whole path from parsing the dependencies file deep into the core logic, and I need to build an understanding of the code base first.

0x53A:
So I'm a bit in a bind. 

I was originally the one to remove the net461<>ns2 mapping (https://github.com/fsprojects/Paket/pull/2380), then tried to re-add it (https://github.com/fsprojects/Paket/pull/2391), but gave up because I personally didn't need it then, and the routing was a lot of effort.

Now I need to use ASP.NET Core from a .NET 4.6.1 project (hopefully moving to .NET Core soon, but not soon enough), so am forced to fall back to plain NuGet PackageReference.

The proposals above (Environment variable and/or app.config) were partially in jest, but also partially serious.

Routing it through everything will touch almost everything, and break every public API because of added fields and parameters. This all for a small gain, which will be of less value with time as more and more moves to .NET 4.7.2+ and .NET Core.

----------

My current preference would be a switch in the app.config of paket.exe. This app.config would be checked in, so the state is stable between all developers and CI.

It should write a flag into the lockfile, erroring out if the lock was resolved with a different state than the current one, forcing you to reinstall.

It should not be possible (because of the effort involved) to specify this per-group, this is a per-repository opt-in.

I have looked at it a little bit, and it would be possible to surgically insert this into only a handfull of files - the framework definition and the places where the lockfile is parsed/written out.

@matthid  @forki What do you think?

matthid:
If it works with magic mode I’m fine with this in principle. The only thing is that app.config is no longer a thing in netcore so a different format might make sense...

0x53A:
> If it works with magic mode I’m fine with this in principle. The only thing is that app.config is no longer a thing in netcore so a different format might make sense...

Ah, good catch about the app.config.
Yes, in magic mode it would need to read the whatever-file from the stub, not the full exe.

> I personally think it should be a group setting

I also think it should be a group setting, but a group setting would be a lot of effort for a debatable gain.

forki:
tha gain is consistency. I think that's important

0x53A:
Yes, and the cause of the issue is a (hopefully) one-time massive hack by the dotnet sdk.

If there were a tooling-level 3 and 4 in the foreseeable future, then it would be different, but if everything goes well, then this switch could be added now, and forgotten and buried in 5 years.

0x53A:
The current mapping is also wrong:

https://github.com/Microsoft/dotnet/blob/master/releases/net471/KnownIssues/514195-Targeting%20.NET%20Framework%204.7.1%20copies%20extra%20files%20to%20your%20bin%20directory.md

> .NET Framework 4.7.1 intended to have full support for running .NET Standard 2.0 assets, but we found some bugs on 12 assemblies that would cause some types to not get forwarded correctly. Patching those 12 assemblies is not viable since then the applications would require the patch to be installed on end-user machines in order to function correctly.

https://github.com/fsprojects/Paket/blob/38b50c108db386cbd64bd6d84bf54508b560869f/src/Paket.Core/Versioning/FrameworkHandling.fs#L598-L600

So this should probably be fixed so only netfx47**2**+ supports netstandard2.0 natively.

matthid:
@0x53A Ok we can change that as well, even Microsoft seems to roll back: https://twitter.com/terrajobst/status/1032025428242849793 so I guess we can close all the PRs and issues floating around?

0x53A:
Burn it all down

baronfel:
hahahahahahahahahahaahahahahahahaaha

*deep breath*

hahahahahahahahahahaahahahahaha

*cries*

TeaDrivenDev:
That would mean in cases like ours, where it works with NuGet, even if unofficially from now on, we have no upgrade path to Paket unless we can move to whatever 4.7.x version works with .NET Standard 2.0 (which is beyond our control and likely to be years away); is that correct?

matthid:
Correct is that from our side nothing has changed. We probably would still accept a well tested PR as long as it is explicit opt-in according to the discussion, but it is unlikely that we are doing it. Considering that no PR has been sent until now it becomes more and more unlikely that it happens at all

Issue #3270, opened 2018-06-27
Does Paket have any culpability in VS intellisense failing issue?


jackfoxy:
Before I blast this message publicly, I would like some feedback from people more familiar with the intricacies of paket/dotnet. I have only every used this combination for F# projects, and never pure nuget/dotnet. Thanks.

The message:

I had high hopes for F# developer tools in VS 17.8 Preview 2, and at first I thought all was well. I was wrong. Preview 2 and now Preview 3 is an absolute shit show.

You never know if intellisense is going to work or not, but you can be certain of one thing. If it is working, it will soon break. F12 works everywhere, then doesn't work. 

I don't even know how to begin a bug report on this. If you can get intellisense and F12 to work, sooner or later you will type or paste in a code snippet that does not build and that is the end. Closing and reopening the file does not fix it. Closing and reopening VS does not fix it. Rebooting does not fix it.

The only thing that might work (and this doesn't even always work) is:

close vs
git clean -xfd
paket update -f (I think install might work better, but I have not experimented)
dotnet restore

Is the problem at the juncture of Roslyn tools and dotnet SDK?

Migrating F# development to Roslyn tools was probably a mistake in retrospect. What did F# developers get out of it? Answer: light bulbs.

Oh, and guess what? When everything else about the development tools is broken, the light bulbs don't work either.

What I'm talking about is real world solutions in production breaking the tools.

Everyone at Microsoft involved with F# needs to stop whatever they are working on and fix the developer tools. 

I will stop now before I write something I regret.

cartermp:
@jackfoxy Paket is not supported with implicit restore at design-time, so you must restore before you open in Visual Studio. Otherwise, tooling will simply not light up because references cannot be resolved. This is due to design constraints both in the project system and in Paket, which are very difficult to reconcile on either side. This general problem isn't new to 15.8, though if you've first adopted paket for larger solutions with 15.8 previews it's understandable that it would surface these problems for you.

> Is the problem at the juncture of Roslyn tools and dotnet SDK?

Doesn't sound like it. We also dogfood our own codebase with a mix of .NET SDK and non-SDK and do not encounter what you're describing. We do not use Paket, though.

> You never know if intellisense is going to work or not, but you can be certain of one thing. If it is working, it will soon break. F12 works everywhere, then doesn't work.
>
[...]

PR #3273, opened 2018-06-29
[WIP] Move the dependencies over to props during VS Background build

forki:
references #3270 

We already create a *.csproj.paket.props file in /obj in order to put CliReferences there. Here we dump the PackageReferences there as well (only when we are in background build). VS is watching the file and triggers restore/resolution.

    Condition="$(DesignTimeBuild) != true 

For rider and ionide we would need such properties as well.


forki:
@matthid there is one known issue. for the very first build after paket updated itself we have a caching problem. the heuristic will think everything is up-to-date and won't call paket and therefor the props file won't change. but it will also no longer add the deps in memory. any ideas how to invalidate the restore cache?

src/Paket.Core/Installation/InstallProcess.fs
```diff
@@ -338,25 +338,13 @@ let private applyBindingRedirects isFirstGroup createNewBindingFiles cleanBindin
...
     applyBindingRedirectsToFolder isFirstGroup createNewBindingFiles cleanBindingRedirects root allKnownLibNames bindingRedirects
-
-let invalidateRestoreCachesForDotnetSdk (projectFileInfo:FileInfo) =
-    let paketPropsFile = ProjectFile.getPaketPropsFileInfo projectFileInfo
-    if paketPropsFile.Exists then
-        let old = File.ReadAllText paketPropsFile.FullName
-        let newContent = old.Replace("<!-- <RestoreSuccess>False</RestoreSuccess> -->","<RestoreSuccess>False</RestoreSuccess>")
-        File.WriteAllText(paketPropsFile.FullName, newContent)
-    let assetsFile = ProjectFile.getAssetsFileInfo projectFileInfo
-    if assetsFile.Exists then
```
forki: @Krzysztof-Cieslak @enricosada  I'm not sure if this will break the ionide integration. FSAuto was watching the assets file, right? does it watch the props files?

matthid:
>  for the very first build after paket updated itself we have a caching problem.

paket updated itself? what? You mean after `paket update`?

> the heuristic will think everything is up-to-date 

It should be up-to-date after paket update?

forki:
no I mean after magic mode updates paket.exe and the targets file

matthid:
> updates paket.exe and the targets file

And why does that matter for the cache? Are we dynamically rewriting the targets file?

forki:
Yes we do, but maybe not in that case. Need to tests it.

Matthias Dittrich <notifications@github.com> schrieb am Fr., 29. Juni 2018,
16:49:

matthid:
Ok then maybe we should stop doing that ;)

src/Paket.Core/Installation/RestoreProcess.fs
```diff
@@ -264,22 +265,49 @@ let createPaketPropsFile (cliTools:ResolvedPackage seq) restoreSuccess (fileInfo
...
+                         sprintf """            <Version>%O</Version>""" p.Version
+                        //"""            <PrivateAssets Condition=" ('%(PaketReferencesFileLinesInfo.AllPrivateAssets)' == 'true') Or ('$(PackAsTool)' == 'true') ">All</PrivateAssets>""" +
+                        //"""            <ExcludeAssets Condition="%(PaketReferencesFileLinesInfo.AllPrivateAssets) == 'exclude'">runtime</ExcludeAssets>""" +
+                        //"""            <Publish Condition=" '$(PackAsTool)' == 'true' ">true</Publish>""" +
+                         """        </PackageReference>"""])
+
+                [yield sprintf "    <ItemGroup Condition=\"($(DesignTimeBuild) == true)%s\">" condition
+                 yield! packageReferences
+                 yield "    </ItemGroup>"])
+            |> fun xs -> String.Join(Environment.NewLine,xs)
```
teo-tsirpanis: Do you know that there is a [`String.concat`](https://msdn.microsoft.com/en-us/visualfsharpdocs/conceptual/string.concat-function-%5Bfsharp%5D?f=255&MSPPError=-2147217396) library routine?

forki: Yes, but what would that improve?

PR #3276, opened 2018-07-01
fix #3275 (pack localized assemblies with new .csproj)

### LINKED ISSUE #3275: Localized assemblies pack does not work with new csproj file
### Description
I using `paket pack` to generate package.
Paket trying to find localized language by examining `EmbeddedResource.Include` Item in proj file (`ProjectFile.FindLocalizedLanguageNames`).
In new csproj it can be omitted, so Paket can't find any languages and do not pack any localized assemblies.

[PackTest.zip](https://github.com/fsprojects/Paket/files/2152688/PackTest.zip)

### Known workarounds

Write something like 
```
<EmbeddedResource Remove="**/*.resx" />
<EmbeddedResource Include="test.en.resx" />
<EmbeddedResource Include="test.ru.resx" />
```

or use `dotnet pack`.

### Fix
I see two ways to correct this:
1. Drop `FindLocalizedLanguageNames` and pack all `$(assemblyName).resources.dll` in `$(outputDirectory)/lang`s folder. 
But in that case Paket will not be able to throw error `Did not find satellite assembly for (%s) try building and running pack again.` (I'm not sure it is matter)

2. Try to be clever about EmbeddedResource and for new csproj evaluate all EmbeddedResource Remove/Include/Update including wildcards values.

lexarchik:
Try to fix #3275.

Pack all `.resources.dll` in `outputDir/<cultureName>` subfolders.


forki:
released in latest beta. thx

PR #3280, opened 2018-07-03
Make the group parameter for outdated work like everywhere else

inosik:
Prior to this change, the `--group` parameter was case-sensitive, and probably some other minor issues.


forki:
released in latest alpha. thanks

PR #3298, opened 2018-07-13
[symlinks] fix: ln "-T" switch removed when isMacOS. NB. -T is not valid on OSX.

sgtz:
…aket.exe install" to fail when "storage: symlink" is set in paket.dependencies. The fix merely removes -T when isMacOS is true.


forki:
thx

Issue #3008, opened 2018-01-23
If paket detects out of date packages, it should list them


mavnn:
### Description

There are situations where Paket will check for new versions of dependencies during `paket install` even though paket.dependencies and paket.lock are unchanged. A common example is when you have a dependency with only prerelease versions available (but not marked as prerelease) - paket will check every time to see if a release version has become available.

This is fine, but currently paket does not tell you *why* it is checking; this can be frustrating with large projects which suddenly become very slow to work with.

### Repro steps

Install a package with only prerelease versions available. Do not mark it prerelease in paket.dependencies.

Run `paket install` again - paket will go and hit nuget.org with no apparent reason. 

### Expected behavior

Paket should display a message explaining what has happened, e.g.:

> Running resolver for group Main since package MyPackage is a prerelease version; 
> specify an exact version or "prerelease" in paket.dependencies if you do not which to check for updates. 

### Actual behavior

The resolver runs, every time

### Known workarounds

@forki tells you about this behaviour on Twitter 😛

mavnn:
Sorry, yeah. The title text is meaningless, I have changed it!

I wrote this in a bit of a hurry to capture what had happened and my frustration at it - and to see if it would be possible to spare other people the frustration!

In my personal opinion, I'd be very happy if paket always displayed the reason why it was going off to check packages. The only ones I'm aware of are:

- packages aren't there
- dependency/lock file changed
- we locked a prerelease version of a package that isn't marked prerelease in paket.dependencies, so we go to check if there's now a release

I wasn't aware of that last one, and it caused me some pain :). There might be others, but if there are I don't know what they are - and I doubt most other users do either!

PR #3302, opened 2018-07-14
[WIP] Give reasons for resolving group (#3008)

TeaDrivenDev:
This is a work in progress to discuss whether this is detailed enough, and a few questions I have about the code.

### DIFF
--- a/src/Paket.Core/Installation/UpdateProcess.fs
+++ b/src/Paket.Core/Installation/UpdateProcess.fs
@@ -44,6 +44,42 @@ let selectiveUpdate force getSha1 getVersionsF getPackageDetailsF getRuntimeGrap
 
     let getPreferredVersionsF,getPackageDetailsF,groupsToUpdate =
         let changes,groups =
+            let install groupName =
+                let hasAnyChanges,nuGetChanges,remoteFileChanges,getChanges = DependencyChangeDetection.GetChanges(dependenciesFile,lockFile,true)
+
+                let hasChanges groupName x = 
+                    match getChanges groupName x with
+                    | [] ->
+                        tracefn "Skipping resolver for group %O since it is already up-to-date" groupName
+                        false
+                    | changes ->
+                        tracefn "Resolving group %O because of changes:" groupName
+
+                        changes
+                        |> List.map (function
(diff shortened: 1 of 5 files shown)


src/Paket.Core/Installation/UpdateProcess.fs
```diff
@@ -10,6 +10,8 @@ open Chessie.ErrorHandling
 open Paket.Logging
 open InstallProcess
 
+type RTC = DependencyChangeDetection.ResolutionTriggeringChange
```
TeaDrivenDev: This is only to bring in the type to avoid having to fully qualify 

The alternatives I currently see are
- opening the DependencyChangeDetection module; not sure if we'd want to do that here
- specifying the type and module every time in the pattern match, which seems ugly
- moving the original type somewhere else, like to `Paket.Domain`; not sure if that would be correct

TeaDrivenDev:
This addresses #3008 opened by @mavnn.

I have "surfaced" the information about specific changes that cause the resolver to be run for a group, and when that decision is made, every single individual reason for resolving is listed.

One question is what the desired amount of detail for the different cases is. Currently
- all the different NuGet changes are listed individually; if there are multiple changes for a single package, all of them will be listed
- if the framework restriction for a group changes, every single package will also be listed as having a framework restriction change; that probably doesn't make much sense
- remote file changes will say "Remote file x changed in project y"
- settings changes will only say "Group has settings changes"

Is there anything that should have more or less detail?

Ignore the individual commits and changes to files other than `DependencyChangeDetection.fs` and `UpdateProcess.fs`; I will rebase and clean those up later.

src/Paket/Program.fs
```diff
@@ -918,3 +918,5 @@ let main() =
         else printError exn
 
 main()
+
+ignore ()
```
forki: ?

TeaDrivenDev: @forki That's one of the things to.... ignore. ;-) I used that to set a final breakpoint for debugging and didn't want to sort out what to commit while this is still unfinished. As I said, I'll clean everything up when I finish the PR.

PR #3304, opened 2018-07-16
Report NuGet download time correctly

### LINKED ISSUE #3303: Paket reports incorrect download speed after a "Still downloading" message has been printed
### Description

If Paket is taking long enough to download a NuGet package that it has to print a "Still downloading" message, then the download speed reported in the final "Download done in N seconds" message will be incorrect.

It's hard to give repro steps that will consistently reproduce the problem since it's dependent on your Internet speed, but here's a snippet of log messages that demonstrate the problem:

```text
[2018-07-14 11:20:04] rmunn@HOST:~/PROJECT_DIR (pr-110)$ time ./build.sh 
Paket version 5.174.2
Starting full restore process.
Downloading System.Diagnostics.DiagnosticSource 4.5 (Build)
Downloading FAKE 5.2 (Build)
Downloading System.Security.Cryptography.Cng 4.5 (Build)
Downloading System.Reflection.TypeExtensions 4.5 (Build)
Download of System.Security.Cryptography.Cng 4.5 (Build) done in 3 seconds. (1547 kbit/s, 0 MB)
Download of System.Reflection.TypeExtensions 4.5 (Build) done in 4 seconds. (490 kbit/s, 0 MB)
Download of System.Diagnostics.DiagnosticSource 4.5 (Build) done in 4 seconds. (266 kbit/s, 0 MB)
Still downloading from https://www.nuget.org/api/v2/package/FAKE/5.2.0 to /home/rmunn/.nuget/packages/fake/5.2.0/fake.5.2.0.nupkg (15531 kbit/s, 89 %)
Download of FAKE 5.2 (Build) done in 13 seconds. (294689 kbit/s, 20 MB)
```

Note how the other packages had relatively low download speeds (on the order of 500-1500 kbit/s). The first 90% of FAKE downloaded at a speed of about 15Mbit/s. But the final download-speed report said that FAKE downloaded with an overall speed of about 300Mbit/s. The computer this happened on is connected to the building's network via an old 100-megabit router, so it's not possible for the reported 300Mbit/s speed to be correct. :-)

A look at the `download` function in [NuGet.fs](https://github.com/fsprojects/Paket/blob/e7c939ecbf0fcc4b6c18dfa608b167442dce9f0a/src/Paket.Core/Dependencies/NuGet.fs#L907-L951) reveals that the speed calculation in the final report is using the wrong value. Here's the relevant code, with the irrelevant lines trimmed and replaced by comments:

```fsharp
let lastSpeedMeasure = Stopwatch.StartNew()
let mutable readSinceLastMeasure = 0L
// Set up HTTP request and start writing to a file stream
let bytesRead = ref -1
let mutable pos = 0L
while !bytesRead <> 0 do
    if printProgress && lastSpeedMeasure.Elapsed > TimeSpan.FromSeconds(10.) then
        let speed = int (float readSinceLastMeasure * 8. / float lastSpeedMeasure.ElapsedMilliseconds)
        let percent =
            match length with
            | Some l -> sprintf "%d" (int (pos * 100L / l))
            | None -> "Unknown"
        tracefn "Still downloading from %O to %s (%d kbit/s, %s %%)" !downloadUrl targetFileName speed percent
        readSinceLastMeasure <- 0L
        lastSpeedMeasure.Restart()
    // Continue reading, and let bytes = (bytes read in next chunk of request)
    bytesRead := bytes
    readSinceLastMeasure <- readSinceLastMeasure + int64 bytes
    pos <- pos + int64 bytes

let speed = int (float pos * 8. / float lastSpeedMeasure.ElapsedMilliseconds)
let size = pos / (1024L * 1024L)
tracefn "Download of %O %O%s done in %s. (%d kbit/s, %d MB)" packageName version groupString (Utils.TimeSpanToReadableString sw.Elapsed) speed size
```

So the final reported speed is calculated by using the total size of the file, and dividing by the time elapsed since the last speed measurement. In my case, that took 100% of the size of the FAKE package, and divided it by the time it took to download the last 10%, resulting in a speed measurement that was about ten times the reality.

Two possible solutions:
* Keep two stopwatches, one for the total elapsed time and a second for the last speed measurement.
* Change the final `let speed` line to be ` int (float readSinceLastMeasure * 8. / float lastSpeedMeasure.ElapsedMilliseconds)`, i.e. use `readSinceLastMeasure` for the size calculation instead of `pos.`

The latter is the simplest change, but means that the total download speed reported will actually be the speed of the last segment. If the user's Internet connection fluctuates in speed, this may not be quite correct. The former approach, with two stopwatches running at the same time, is slightly more complicated but will produce accurate speed measurements of the total download time, not just the last downloaded chunk.

rmunn:
Use the correct stopwatch to report total download speed (the one that's been running continuously, not the one that gets reset every ten seconds).

Fixes #3303.

### DIFF
--- a/src/Paket.Core/Dependencies/NuGet.fs
+++ b/src/Paket.Core/Dependencies/NuGet.fs
@@ -946,7 +946,7 @@ let private downloadAndExtractPackage(alternativeProjectRoot, root, isLocalOverr
                     | HttpStatusCode.OK -> ()
                     | statusCode -> failwithf "HTTP status code was %d - %O" (int statusCode) statusCode
 
-                    let speed = int (float pos * 8. / float lastSpeedMeasure.ElapsedMilliseconds)
+                    let speed = int (float pos * 8. / float sw.ElapsedMilliseconds)
                     let size = pos / (1024L * 1024L)
                     tracefn "Download of %O %O%s done in %s. (%d kbit/s, %d MB)" packageName version groupString (Utils.TimeSpanToReadableString sw.Elapsed) speed size
 


forki:
thx

Issue #910, opened 2015-07-01
paket install (+update?) should not overwrite EmbedInteropTypes flag


agross:
We have a nuget-packaged dependency that requires to be referenced with the `EmbedInteropTypes` set to true. I know that nuget's install.ps1 could be used to set the flag and that paket doesn't support PowerShell (which is a good thing):

``` xml
<Reference Include="Interop.Foobar">
  <HintPath>..\packages\Interop.Foobar.dll</HintPath>
  <EmbedInteropTypes>True</EmbedInteropTypes>
</Reference>
```

While we can certainly set the flag ourselves, `paket install` (even without `--hard`) will remove the flag.

It would be great if `EmbedInteropTypes` was kept.

forki:
does this thingy have `<paket>true</paket>`?

agross:
Yes, I just copied our pre-Paket reference.

forki:
ok. then try to set the falg to false.

that said: I think we should add that to our settings, so that paket can generate that itself.

agross:
I think it'll probably be the best if paket migrates all custom flags and just manages the `HintPath` element. What do you think?

forki:
Maybe, but that's not really easy with the current design.

Currently we do on every install:
1) delete existing paket = true nodes
2) add nodes for all packages where we can't find an existing node

agross:
So if we just manually reference the DLL paket won't re-add the node with paket=true?

(The dep doesn't need TargetFramework-specific nodes.)

forki:
Yes. At least as long as you don't use --hard. But I'm not completely sure
about the other libs. I assume they will be ignored as well.

matthid:
Closing as the interest doesn't seem to be high and new sdk things are different anyway.

PR #3314, opened 2018-07-27
Allow addition of <EmbedInteropTypes> for Nuget packages

ruhullahshah:
- Problem
NuGet package reference with a COM interop DLL, `<EmbedInteropTypes>True</EmbedInteropTypes>` needs to be managed manually and `paket update` removes the flag.

- Solution
Provide a way to express this through existing `paket.references` syntax
`PkgName embed_interop_types: true`

- Related Issues
https://github.com/fsprojects/Paket/issues/910

There might be some failing testcases but adding the PR to open up the discussion.

### DIFF
--- a/docs/content/references-files.md
+++ b/docs/content/references-files.md
@@ -79,6 +79,11 @@ DotNetZip framework: >= net45
 FSharp.Core redirects: on
 ```
 
+## Adding support for COM interop DLL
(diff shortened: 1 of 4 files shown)


forki:
thanks!

ruhullahshah:
You are welcome. Thanks for taking care of the failing tests.

PR #3357, opened 2018-09-12
Add nuget package output details to paket pack output

ruhullahshah:
It turns out to be useful for certain scenarios in our case, to know the details of the nuget package generated by the `paket pack` in a CI setting. So we propose to add the details to the output.


forki:
thx

ruhullahshah:
You are welcome

PR #3363, opened 2018-09-17
Paket pack with `--template` fails trying to load the dependencies of templates who should be ignored instead

enricosada:
When specified the `--template` argument, the `paket pack` search also for project references (finding all the `paket.template` files in same directory of project files)

Doing so, paket try to load **and validate** the found `paket.template` files.
But these template files may be not needed (like when `--template` is specified), and some may fail to load (for example a `files` required doesnt exists, because was not built)

This PR refactor the code to parse the found paket.template files, and just load these when are needed (if are not needed, are not validated, just considered for metadata)


enricosada:
@forki @matthid ready to review.

test failure are unrelated (master already fails, ref [appveyor](https://ci.appveyor.com/project/paket/paket/build/0.0.1.766/job/x7tvsfxswnxarsx4/tests) )

forki:
thx

PR #3364, opened 2018-09-17
Referencing a zip file

mgnslndh:
Document how Paket behaves when using a zip archive as a http dependency


enricosada:
👍 nice @mgnslndh !

i added a fix to the markdown style, so is not all in one line

enricosada:
Merging. Thank you @mgnslndh

mgnslndh:
Thank you!

PR #3367, opened 2018-09-19
Add NetcoreApp3.0 support

forki:


enricosada:
I checked in code and should inherit from netcoreapp2.2 and netstandard 2.0

maybe adding an unit test like https://github.com/fsprojects/Paket/blob/8fc56c9c57a772fe493321db392bc6e7ab2c9a7c/tests/Paket.Tests/Versioning/FrameworkCompatibilitySpecs.fs#L86 can help maintainance

PR #3371, opened 2018-09-21
[DONT-MERGE] use paket .net core as tool in this repo

enricosada:
just a check that things works.

probably will be a good idea to add a build matrix to check that, like:

- add a `.paket_dotnet` dir with the new files needed
- in appveyor/travis, at beginning, replace `.paket` with `.paket_dotnet` in 
  a build matrix job


enricosada:
Closing.

not needed anymore, as .net core tools is already in `master`

PR #3372, opened 2018-09-21
Switch order of arguments for ln

### LINKED ISSUE #3368: storage: symlink fails on Linux
### Description

`storage: symlink` doesn't work under Linux (Ubuntu 18.04.1). Paket fails to create a symbolic link and throws an error about already existing files.

### Repro steps

* Have an empty folder and setup a basic paket setup.
* Have the following `paket.dependencies`:  
  ```
  storage: symlink
  framework: netstandard2.0
  nuget Newtonsoft.Json
  ```
* Execute `mono .paket/paket.exe install`
* Be greeted with:  
  ```
  ...snip...
  Paket failed with
  -> symlink /home/<me>/.nuget/packages/newtonsoft.json/11.0.2 -> 
  /home/<me>/<somewhere>/packages/Newtonsoft.Json failed with error : [1] with output : 
      and error : /bin/ln: failed to create symbolic link 
  '/home/<me>/.nuget/packages/newtonsoft.json/11.0.2': File exists
  ...snip...
  ```

### Expected behavior

It should create the symlinks.

### Actual behavior

It doesn't create the symlinks.

---

Looking at the code I think you have the `source` and `target` wrong: https://github.com/fsprojects/Paket/blob/86a790131e3e8060e374620b6e3626735c3d51f1/src/Paket.Core/Common/SymlinkUtils.fs#L21

https://linux.die.net/man/1/ln

Contrary to `mklink` with `ln` you have the target and then the link name. In your parameter naming that would be `source` and then `target`.

Not sure if this affects MacOS as well since they use `ln` too.

inosik:
Fixes #3368.

### DIFF
--- a/src/Paket.Core/Common/SymlinkUtils.fs
+++ b/src/Paket.Core/Common/SymlinkUtils.fs
@@ -18,11 +18,11 @@ let makeDirectoryLink target source =
 
     let ln (p:ProcessStartInfo) =
         p.FileName <- "ln"
-        p.Arguments <- sprintf @"-sT ""%s"" ""%s""" target source
+        p.Arguments <- sprintf @"-sT ""%s"" ""%s""" source target
 
     let ln_onMacOS (p:ProcessStartInfo) =
         p.FileName <- "ln"
-        p.Arguments <- sprintf @"-s ""%s"" ""%s""" target source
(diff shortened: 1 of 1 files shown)


forki:

inosik:
Done.

Issue #2902, opened 2017-11-16
Getting started and FAQ disagree on whether to commit .paket/paket.exe


Danl2620:
### Description

In the documents, the "Getting Started with Paket" [section](https://fsprojects.github.io/Paket/getting-started.html#Manual-setup) instructs you to add `.paket/paket.exe` to your source control repository. Meanwhile in the [FAQ](https://fsprojects.github.io/Paket/faq.html#What-files-should-I-commit) it tells you _not_ to add that file to the repository.

forki:
Oh it's difference between magic mode and real paket.exe. Yes we need to
make docs consistent.

Currently we recommend: download latest paket bootstrapper.exe and rename
it to paket.exe and commit it.

Would love to see PRs that fix docs accordingly. Thanks so much

Am 16.11.2017 20:20 schrieb "Dan Liebgold" <notifications@github.com>:

PR #3379, opened 2018-10-02
Defect #2902, removing paket.exe from list of files not added to version control.

morgankenyon:
My first commit to this project. Based on discussion [here](https://github.com/fsprojects/Paket/issues/2902) it seemed like I should remove the reference to paket.exe because that is a file that is needed into source control. 

I'm not sure what your current branching strategy is, if you want me to change which branch to merge into let me know.

Let me know if you have any questions, comments or suggestions.

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -159,10 +159,6 @@ The following files can be committed, but are not essential:
 The following files should *not* be committed to your version control system,
 and should be added to any ignore files:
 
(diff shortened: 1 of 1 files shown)


forki:
HI, I think it would be better to keep that paragraph, change it a bit and link to https://fsprojects.github.io/Paket/bootstrapper.html#Magic-mode

forki (changes requested):
please try to rewrite this to reflect the magic mode

PR #3424, opened 2018-11-14
Update NuGet.fs

b0urb4k1:
Timeouts are only set when the stream supports timeouts.


forki:
thx. can you please check out FAQ if the env variables are documented properly?

b0urb4k1:
Yes, i forgot about that, do i have to make a new pull request for that?

forki:
I copied the FAQ from the old pr. Please double check.

Issue #3435, opened 2018-11-21
using paket as a global tool without paket.bootstrapper


vilinski:
### Description

Maybe I'm doing it wrong?  I have installed paket as a global dotnet tool and used it in terminal with a new dotnet project.
I supposed I don't need a bootstrapper anymore.
Getting error by `dotnet build` or `restore`

### Repro steps

```
dotnet tool install paket -g
mkdir -p classlib/src/classlib
cd classlib/src/classlib
dotnet new classlib -lang f#
cd ../..
paket init
paket add FSharp.Core --project src/classlib/classlib.fsproj
cd src/classlib
dotnet build
```

If possible then please create a git repository with a repro sample or attach a zip to the issue.

### Expected behavior

Project built without any errors.

### Actual behavior

`paket update` works, but neither `dotnet build` nor `dotnet restore`. Getting following error:

>$ dotnet build
  shasum: /Users/username/Projects/fsharp/classlib/.paket/../paket-files/paket.restore.cached:
  Cannot open assembly '/Users/username/Projects/fsharp/classlib/.paket/paket.exe': No such file or directory.
/Users/username/Projects/fsharp/classlib/.paket/Paket.Restore.targets(91,5): error MSB3073: The command "/Library/Frameworks/Mono.framework/Commands/mono --runtime=v4.0.30319 "/Users/username/Projects/fsharp/classlib/.paket/paket.exe" restore" exited with code 2. [/Users/username/Projects/fsharp/classlib/src/classlib/classlib.fsproj]

# Environment

MacOS with latest mono, dotnet SDK and paket

[classlib.tar.gz](https://github.com/fsprojects/Paket/files/2605871/classlib.tar.gz)

### Known workarounds

Install `paket.bootstrapper` manually into `.paket` folder or use plain nuget instead.

inosik:
From https://github.com/fsprojects/Paket/issues/3436#issuecomment-440981451:

``` shell
dotnet tool install --tool-path .paket Paket
```

I believe the `dotnet` CLI creates the correct shim (`paket.exe` on Windows, `paket` on Unix) and the latest `Paket.Restore.targets` should already support that.

baronfel:
Note that with .net sdk 3 preview 7 you can do this in a more centrally-managed way: https://gist.github.com/baronfel/80f49ecb93ebacb84ca840fca7a12fc2

PR #3445, opened 2018-12-01
Making Paket.Restore.targets work with Paket as a global tool.

tomzo:
Adds support for using paket as global tool #3435 #3398

As in  #3398 - currently there is little benefit of installing paket with `dotnet tool install -g Paket --version *` because soon after using paket restore, it will regenerate `Paket.Restore.targets` and next restore will attempt to run with `mono`.

This PR is a rebase of @NatElkins work.

It changes the fallback to actually prefer a globally available `paket` command rather than falling back to mono.

### DIFF
--- a/src/Paket.Core/embedded/Paket.Restore.targets
+++ b/src/Paket.Core/embedded/Paket.Restore.targets
@@ -21,16 +21,16 @@
     <PaketExePath Condition=" '$(PaketExePath)' == '' AND Exists('$(PaketToolsPath)paket')">$(PaketToolsPath)paket</PaketExePath>
     <PaketExePath Condition=" '$(PaketExePath)' == '' AND Exists('$(PaketRootPath)paket.exe')">$(PaketRootPath)paket.exe</PaketExePath>
 
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' == 'Windows_NT' ">$(PaketToolsPath)paket.exe</PaketExePath>
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'assembly' ">$(PaketToolsPath)paket.exe</PaketExePath>
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'native' ">$(PaketToolsPath)paket</PaketExePath>
+    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' == 'Windows_NT' AND Exists('$(PaketToolsPath)paket.exe')">$(PaketToolsPath)paket.exe</PaketExePath>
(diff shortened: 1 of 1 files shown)


src/Paket.Core/embedded/Paket.Restore.targets
```diff
@@ -21,16 +21,16 @@
...
     <PaketExePath Condition=" '$(PaketExePath)' == '' AND Exists('$(PaketRootPath)paket.exe')">$(PaketRootPath)paket.exe</PaketExePath>
 
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' == 'Windows_NT' ">$(PaketToolsPath)paket.exe</PaketExePath>
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'assembly' ">$(PaketToolsPath)paket.exe</PaketExePath>
-    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'native' ">$(PaketToolsPath)paket</PaketExePath>
+    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' == 'Windows_NT' AND Exists('$(PaketToolsPath)paket.exe')">$(PaketToolsPath)paket.exe</PaketExePath>
+    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'assembly' AND Exists('$(PaketToolsPath)paket.exe') ">$(PaketToolsPath)paket.exe</PaketExePath>
+    <PaketExePath Condition=" '$(PaketExePath)' == '' AND '$(OS)' != 'Windows_NT' AND '$(PaketExeImage)' == 'native' AND Exists('$(PaketToolsPath)paket')  ">$(PaketToolsPath)paket</PaketExePath>
```
lust4life: seems changes here will break the `PaketBootstrapperStyle == proj` style, since if we will use `.paket/paket.bootstrapper.props` [mentioned here ](https://github.com/enricosada/paket-netcore-testing-as-tool#how-to-migrate-from-paket-net ) , then this `Condition` will never matched (if `Exists('$(PaketToolsPath)paket')` then `PaketExePath` will be assigned before). Since `paket` will be installed from target `PaketBootstrapping` after this check.

The old version will cover this situation. If we want to support global paket installed like this `dotnet tool install -g Paket`, seems we should check `Exists('$(PaketToolsPath)paket.bootstrapper.proj')` not check `Exists('$(PaketToolsPath)paket')`

forki:
Any chance you can send a pull request?

Am Do., 6. Dez. 2018, 09:34 hat Jun <notifications@github.com> geschrieben:

TWith2Sugars:
I've just done a PR after testing locally. Before this change paket as a global tool failed to restore. After it works fine.

lust4life:
@TWith2Sugars  😂  and one more pr to fix the same thing #3451

PR #3451, opened 2018-12-06
fix conflict between native local-path style with global install style

lust4life:
issues here:

If we use `paket.bootstrapper.proj` style try to install paket in `PaketToolsPath` in the first time (no more paket or paket.exe ), targets here will use `paket` as the default `PaketCommand`, but `paket.bootstrapper.proj` will install paket in `PaketToolsPath` , so we should check `PaketToolsPath` then fallback to global native style. Otherwise we will get `paket: not found` error.

mentioned in https://github.com/fsprojects/Paket/pull/3445#discussion_r239364069

strategy here:
```
 <!-- windows, root => tool => proj style => bootstrapper => global -->
 <!-- no windows,  root => tool => proj style => mono paket => bootstrpper => global -->
```


lust4life:
https://github.com/fsprojects/Paket/pull/3445#issuecomment-444791354
@forki  ping.

PR #3466, opened 2019-01-02
Use the correct request header

### LINKED ISSUE #3458: Paket (.net core) push fails
### Description

When using the .net core version of paket, pushing nuget packages fails.

Example push to myget:
```
Could not push /xxx/xxx.nupkg: Request to 'https://www.myget.org/F/xxx/api/v2/package' failed with: 400 BadRequest — Invalid or corrupt NuGet package detected. Number of entries expected in End Of Central Directory does not correspond to number of entries in Central Directory.
```

Example push to Octopus Deploy:
```
Could not push /xxx/xxx.nupkg: Request to 'https://xxx/nuget/packages' failed with: 400 BadRequest — {
  "ErrorMessage": "There was a problem with your request.",
  "Errors": [
    "A package file must be provided"
  ],
  "ParsedHelpLinks": []
}
```

The .nupkg file is valid and swapping back to using `mono paket.exe push ...` is able to push the file successfully.

Possibly related:
When using .net core, `System.Net.WebClient` actually uses `System.Net.Http.HttpClient` under the hood.

### Known workarounds

Use the full .NET version of paket to push .nupkg files.

inosik:
Fixes #3458.

### DIFF
--- a/src/Paket.Core/Common/NetUtils.fs
+++ b/src/Paket.Core/Common/NetUtils.fs
@@ -322,7 +322,6 @@ type HttpClient with
             |> Encoding.UTF8.GetBytes
         let newlineBytes = Environment.NewLine |> Encoding.UTF8.GetBytes
         let trailerbytes = String.Format(System.Globalization.CultureInfo.InvariantCulture, "--{0}--", boundary) |> Encoding.UTF8.GetBytes
-        x.DefaultRequestHeaders.Add("Content-Type", "multipart/form-data; boundary=" + boundary)
         use stream = new MemoryStream() // x.OpenWrite(url, "PUT")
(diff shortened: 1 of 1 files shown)


inosik:
Hm, crashes at runtime. I'm looking into that.

inosik:
This is ready now. I followed [an answer on SO](https://stackoverflow.com/a/10679340) to fix the runtime error. I also tried to use the `Headers.ContentType` property to be safe, but the constructor of `System.Net.Http.Headers.MediaTypeHeaderValue` doesn't accept the value we pass in.

haf:
This would not have been an issue had you been using HTTP.fs BTW.

haf:
Thank you @inosik

Issue #2804, opened 2017-09-30
pack --include-referenced-projects ignores recursive inter-project dependencies with paket.template


mbett7:
### Description
`pack --include-referenced-projects` misses recursive inter-project dependencies that have a `paket.template` file.

i.e. in a solution with dependencies:
ProjectA -> ProjectB -> ProjectC

If ProjectA and ProjectC were to have `paket.template` files, The packed ProjectA nuspec should have a dependency on ProjectC

### Repro steps
1. Extract [ReferencePackTest.zip](https://github.com/fsprojects/Paket/files/1346184/ReferencePackTest.zip)
2. `paket restore`
3. Build solution
4. Run `pack.cmd`

### Expected behavior
ProjectC is included in ProjectA's nuspec dependencies.

### Actual behavior
It's not.

### Known workarounds
Rollback to paket 4.8.8

Gonnagle:
Was just hitting the exact same issue with 5.156.1. Updating to the latest (5.196.2) seemed to fix the issue.

Based on release notes was probably fixed in [5.195.7](https://github.com/fsprojects/Paket/releases/tag/5.195.7).

mbett7:
Indeed it is.  Thanks!

PR #3469, opened 2019-01-07
fix #2776 pack transitive dependencies with --include-referenced-projects

### LINKED ISSUE #2776: [Question/possibly bug] How `--include-referenced-projects` should work?
For now `--include-referenced-projects` switch does not include transitive dependencies in package.
Setup: 3 projects: [test-transitive.zip](https://github.com/fsprojects/Paket/files/1310282/test-transitive.)
projectA (only project with template file) has reference to projectB
projectB has reference to projectC. 

If i pack projectA with `include-referenced-projects` result package contains projectA.dll and projectB.dll, but not projectC.dll.
Same for nuget dependencies. Result package have only dependencies referenced by projectB, but not projectC.

I thought idea of `include-referenced-projects` is to get package that contains all it needed.

Of course i can add all i need to template file in dependencies/files section, but it makes `include-referenced-projects` less helpful.

lexarchik:
#2776
Example to illustrate fix:
```
ProjectA (with templatefile)
|-> ProjectB
    |-> ProjectC (with templatefile, nuget nlog)
        |-> ProjectD
    |-> ProjectE
        |-> ProjectF (nuget moq)

(template files is "empty" - only description/authors, not adding any files to package)
```
When we pack ProjectA with `--include-referenced-projects` flag resulting package must contains: 
`ProjectB.dll`, `ProjectE.dll`, `ProjectF.dll`, nuget dependency to `ProjectC` and nuget dependency to `moq`

With current version of Paket only `ProjectB.dll` get into package


lexarchik:
ping

Gonnagle:
This fixed #2804 (at least seems like it and the description matches really well to the described issue).

PR #3471, opened 2019-01-08
Do not change the AutoGenerateBindingRedirects for exe output type

BlythMeister:
This has some weird behavior when you set to true on exe output types.

It's safer to leave them alone and let the user decide to enable if they wish.

Currently, i am having to revert the paket change so as to keep my actual redirects rather than those auto generated.

Stupidly in MSBuild, this switch is ONLY used in exe output types to generate the redirects.
But when set on Library output types, it prevents warnings...slow clap for msbuild using the same property for 2 tasks...

### DIFF
--- a/src/Paket.Core/Installation/BindingRedirects.fs
+++ b/src/Paket.Core/Installation/BindingRedirects.fs
@@ -158,9 +158,10 @@ let private applyBindingRedirects isFirstGroup cleanBindingRedirects (allKnownLi
         use f = File.Open(configFilePath, FileMode.Create)
         config.Save(f, SaveOptions.DisableFormatting)
 
-    match projectFile.GetAutoGenerateBindingRedirects() with
-    | Some x when x.ToLower() = "true" -> ignore()
-    | _ -> projectFile.SetOrCreateAutoGenerateBindingRedirects()
+    match projectFile.OutputType, projectFile.GetAutoGenerateBindingRedirects() with
+    | ProjectOutputType.Exe, _ -> ignore()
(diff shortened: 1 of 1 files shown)


BlythMeister:
Are these failures related to my change?  I can't see how they are...i ran the build locally and they go green.

BlythMeister:
@forki any chance fmof getting this (or something similar) in?

Causing me issues when I forget to revert

src/Paket.Core/Installation/BindingRedirects.fs
```diff
@@ -158,9 +158,10 @@ let private applyBindingRedirects isFirstGroup cleanBindingRedirects (allKnownLi
         use f = File.Open(configFilePath, FileMode.Create)
         config.Save(f, SaveOptions.DisableFormatting)
 
-    match projectFile.GetAutoGenerateBindingRedirects() with
-    | Some x when x.ToLower() = "true" -> ignore()
-    | _ -> projectFile.SetOrCreateAutoGenerateBindingRedirects()
+    match projectFile.OutputType, projectFile.GetAutoGenerateBindingRedirects() with
+    | ProjectOutputType.Exe, _ -> ignore()
+    | _, Some x when x.ToLower() = "true" -> ignore()
+    | _, _ -> projectFile.SetOrCreateAutoGenerateBindingRedirects()
```
blumu: Should not the last branch also check that `redirects` is not set to off?

BlythMeister:
Originally this was put to suppress build warnings. I'm starting to think actually we should remove since msbuild appears to have been fixed

PR #3479, opened 2019-01-14
Add PaketRestoreDisabled when NCrunch enabled in targets

BlythMeister:
As suggested here:
https://forum.ncrunch.net/yaf_postst2467_F---net-core-projects-doesn-t-build-in-after-upgrade-to-VS-2017-15-7-0.aspx
When running new style project files & using NCrunch
The paket restore will fail (as its not got everything it needs)
So prevent the restore from running.
NCrunch will still work as expected as it will use the restored files

The Caveat is that a user will have to build manually in order to
1st create the restore files.  But at least this way NCrunch will work!


BlythMeister:
@forki is this the right file to change?

BlythMeister:
It's quite confusing that there are failing tests which seem unlinked to changes made...

BlythMeister:
this one is ready for review please

PR #3483, opened 2019-01-17
Binding redirect on library only

BlythMeister:
Only set binding redirects on libraries.
On all other project types, remove the node from project files if it's present

### DIFF
--- a/src/Paket.Core/Installation/BindingRedirects.fs
+++ b/src/Paket.Core/Installation/BindingRedirects.fs
@@ -159,9 +159,9 @@ let private applyBindingRedirects isFirstGroup cleanBindingRedirects (allKnownLi
         config.Save(f, SaveOptions.DisableFormatting)
 
     match projectFile.OutputType, projectFile.GetAutoGenerateBindingRedirects() with
-    | ProjectOutputType.Exe, _ -> ignore()
-    | _, Some x when x.ToLower() = "true" -> ignore()
-    | _, _ -> projectFile.SetOrCreateAutoGenerateBindingRedirects()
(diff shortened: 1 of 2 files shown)


BlythMeister:
this one is ready for review please

forki:
can you explain in words what that does?

BlythMeister:
If we are adding binding redirects & the project type is a library we check if the "AutoGenerateBindingRedirects" node exists and is true.  If it is, then we do nothing.
If it doesn't exist, or is not true, we will create the node and set it's value to true.

For any other project type when we are adding binding redirects, we will remove the "AutoGenerateBindingRedirects" as this has unintended behaviour and msbuild will then ignore our redirects and create it's own.

forki:
Cc @isaac_abraham

Am Mo., 21. Jan. 2019, 16:13 hat Chris Blyth <notifications@github.com>
geschrieben:

matthid:
Honestly I have no idea and I'd prefer to delegate this mess to msbuild instead of doing it ouself (which would be the reverse of this pr)

BlythMeister:
@matthid at the moment, the situation with paket can cause issues.
This fixes the issues, but i completely agree with you re using msbuild.

My workaround to this was to set redirects off, and then in my projects, add the AutoGenerateBindingRedirects to true so that MsBuild does it for me.

That msbuild option kinda makes the paket binding redirects pointless...

BlythMeister:
@matthid in all honesty, the paket binding re-direct stuff can be gutted in favour of setting `AutoGenerateBindingRedirects` to `true` and let msbuild deal with it.

I typically now set my projects up to have this as true in the proj file & then disable paket redirections.

BlythMeister:
@matthid what do you suggest here

Issue #3449, opened 2018-12-06
Paket prints too many restore information with `--no-restore` in `dotnet build --no-restore`


nightroman:
### Description

Paket prints too many restore information with `--no-restore` in `dotnet build --no-restore`
This is the case with the latest 5.193.0
The older 5.163.2 works as expected/desired.

### Repro steps

Use attached project with latest Paket.5.193.0

Invoke commands:

```cmd
.paket\paket.exe restore
dotnet restore
dotnet build --no-restore
```

### Expected behavior

Output of the last command `dotnet build --no-restore` should be:

```
Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Test1 -> C:\Users\rjq\Desktop\_181206_105136_pk\Paket.5.193.0\bin\Debug\netstandard2.0\Test1.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)
```

### Actual behavior

Output of the last command `dotnet build --no-restore` contains too many information about restoring:

```
Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

  Paket version 5.193.0
  The last restore is still up to date. Nothing left to do.
  Performance:
   - Runtime: 128 milliseconds
  Test1 -> C:\Users\rjq\Desktop\_181206_105136_pk\Paket.5.193.0\bin\Debug\netstandard2.0\Test1.dll

Build succeeded.
    0 Warning(s)
    0 Error(s)
```

The picture is even worse if we build a solution with many projects (30+ in our case).
For every project we get 1 useful log line about DLL and 4 noise lines about restoring.

### Known workarounds

I do not know any workarounds. For the moment I prefer to stay with older
Paket.5.163.2 which works as expected. The project is attached, too.

***
- [Paket.5.193.0.zip](https://github.com/fsprojects/Paket/files/2652932/Paket.5.193.0.zip)
- [Paket.5.163.2.zip](https://github.com/fsprojects/Paket/files/2652935/Paket.5.163.2.zip)

forki:
@matthid any ideas about that?

Am Do., 6. Dez. 2018, 12:49 hat Roman Kuzmin <notifications@github.com>
geschrieben:

nightroman:
pinging...

forki:
[...]

PR #3486, opened 2019-01-17
Fixes for --no-restore still doing a restore

### LINKED ISSUE #3485: Paket restore runs on dotnet build --norestore
### Description

When calling dotnet build with the --norestore option a paket restore is still run.

### Repro steps

On any repo, call donet build --norestore.

### Expected behavior

Paket is not called

### Actual behavior

Paket is called

### Known workarounds

None

BlythMeister:
Fixes #3485 and #3449

The version number used in the PaketPropsVersion in https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Installation/RestoreProcess.fs#L328 is `5.185.3` which does not match the targets.
This means that `PaketRestoreRequired` is always true, even if you have literally just built.


baronfel:
Is there a way to template this value so that it could be populated on release builds? I'm just worried about reintroducing this failure transparently down the line.

Either that or a more low tech fix: a test that checks if the version in the embedded targets matches the current assembly version?

BlythMeister:
@baronfel have added comments above the 2 values which need to be in sync.

These appear to be values for the format of the data since the value in the props does not match that of paket itself.

Have also made an additional change to the `beforetarget` to make paket mirror the order of the standard nuget restore target.

BlythMeister:
Scrap that before target stuff...that's just plain wrong...will rebase and fix up this PR :D

src/Paket.Core/embedded/Paket.Restore.targets
```diff
@@ -102,7 +102,11 @@
       <PaketRestoreRequired Condition=" '$(PaketRestoreLockFileHash)' == '' ">true</PaketRestoreRequired>
     </PropertyGroup>
 
-    <PropertyGroup Condition="'$(PaketPropsVersion)' != '5.174.2' ">
+	<!-- 
+		This value should match the version in the props generated by paket 
+		If they differ, this means we need to do a restore in order to ensure correct dependencies
+	-->
+    <PropertyGroup Condition="'$(PaketPropsVersion)' != '5.185.3' ">
```
forki: I don't think this is correct. This is the props file format version  - and it did not change recently.

forki: maybe we need to check if it's not older than the version in the targets file. dunno

BlythMeister: this commit is when the props version was changed. https://github.com/fsprojects/Paket/commit/828f9e4553d4d242be603e72a08cdec0c5fb454f
This file was not updated at that time.

forki:
Ouch. Good catch

Am Fr., 18. Jan. 2019, 10:16 hat Chris Blyth <notifications@github.com>
geschrieben:

BlythMeister:
i'm an experienced archaeology driven developer ;)

PR #3512, opened 2019-02-21
improved restore speedup

viktor-svub:
hi everyone and sorry for the long silence...
- this change improves performance of `paket restore`, specifically where way too many inter-connected dependencies are involved 
- we have projects depending on the majority of the AspNetCore.App bundle, multi-targeting net462, netstandard2.0, and netcoreapp2.1
- paket restore and/or dotnet restore take tens of seconds, easily over minute, per project/platform combination
- with ANTS profiler, I identified the result-generating loop of LockFile's GetOrderedPackageHull, as it was re-creating a lot of Set instances
- after rewriting the loop to re/use mutable state, the performance improved at least 20 times, taking approx 2 seconds for the whole solution


isaacabraham:
Mutability for the win!

baronfel:
Better yet, profiling + judicious use of localized mutability 👍  Nice work @viktor-svub!

forki:
Cool. What abut the other 2 sets?

Am Do., 21. Feb. 2019, 18:17 hat Chet Husk <notifications@github.com>
geschrieben:

viktor-svub:
Well, I could change them as well, but their operations did not light up under the profiler and I generally avoid touching working code -- using F# Set there does not hurt, so from my PoV it's the preferred way. 

I suspect the inherently expensive slow operation here is Set.map (which seems to always enforce re/creation of new set), and rest of the refactoring was only done to keep the code readable, as an attempt to replace just the map/filter and keep the rest semi-immutable did not end well.

forki:
ok thanks

PR #3523, opened 2019-03-18
Prevent Simplifier removing item if it has settings or version requirements

BlythMeister:
I noticed that when i had any settings or version requirements in references/dependencies files and run a simplify it would remove these items and therefore remove the settings.

Now, as well as the item being implied as a transitive dependency, it also needs to have no settings or version requirements in order to be removed.


BlythMeister:
@forki are these build failures mine, or pre-existing.  It is quite confusing 😕

BlythMeister:
This one is ready

forki:

BlythMeister:
@forki done :)

PR #3524, opened 2019-03-18
Always try and extract Paket.Restore.targets even if up to date

BlythMeister:
Also add a log line to state that Paket.Restore.targets was saved

This means that even if your restore is up to date, it will get the latest Paket.Restore.targets to use


BlythMeister:
@forki are these build failures mine, or pre-existing.  It is quite confusing 😕

BlythMeister:
This is ready for review please.

forki:

BlythMeister:
@forki done :)

forki:
why is this needed?

BlythMeister:
If paket is already up to date, it will not get you the latest restore targets.

This means when targets are updated, they only get downloaded (released) when paket has a need to download a dependency.

PR #3526, opened 2019-03-20
Fix DebuggerDisplay-Attribute of GroupName

inosik:
This makes the debugger display the `compareString` as intended, instead of this error message:

``` text
error CS0103: The name 'Item2' does not exist in the current context
```


forki:
thx

PR #3536, opened 2019-03-31
Use nuget.org's V3 service index

loic-sharma:
The NuGet team advocates the V3 API over the V2 API. I realize Paket can map between the two, but we should prefer the V3 service index if possible.


forki:

forki:
never mind

PR #3538, opened 2019-04-03
Simplify Fix Extra Settings

BlythMeister:
Some settings were missed on simplify improvements.

Update to take these into account & add test cases for these to ensure they are not removed when present.
Also added a test case to ensure "simplify:never" does work (was missed from last PR)


BlythMeister:
@forki this one should be ready.
Also, nice work on getting the builds green, so much easier to check PRS are ok!

BlythMeister:
Thanks @forki

PR #3542, opened 2019-04-07
Adapt PackTask to breaking changes in MsBuild 16

cdrnet:
Attempting to fix #3540

For MsBuils 16 and newer:

Changes:
- BuildOutputFolder -> BuildOutputFolder**s**

While at it, also adds the following (preparation for #3475). I can also remove them from the PR if you prefer:
- PackageLicenseFile
- PackageLicenseExpression
- PackageLicenseExpressionVersion

PR #3543, opened 2019-04-08
Adapt PackTask to breaking changes in MsBUild 16 - missing condition

### LINKED ISSUE #3540: Running dotnet pack with Paket.Restore.Targets referenced in a new csproj file with msbuild 16 fails.
### Description

Running dotnet pack with Paket.Restore.Targets referenced in a new csproj file with msbuild 16 fails.

```
C:\code\xyz\src\.paket\Paket.Restore.targets(298,15): error MSB4064: The "BuildOutputFolder" parameter is not supported by the "PackTask" task. Verify the parameter exists on the task, and it is a settable public instance property. [C:\code\.....csproj]
C:\code\xyz\src\.paket\Paket.Restore.targets(261,5): error MSB4063: The "PackTask" task could not be initialized with its input parameters.  [C:\code\.....csproj]
```

### Known workarounds

Dont install msbuild 16 on the system (luckily I had not gotten around to installing it on the build servers yet) :D 
I have no idea if it is possible to tell dotnet to use a specific version of msbuild, but I havent figured it out so far...but if that is possible that is probably an option as well.

cdrnet:
Unfortunately somehow a condition got lost while preparing the PR #3542. This is needed to properly fix #3540

Without this line, in cases where MsBuild 16 is detected, it actually gets handled by NuGet4.
Sorry!


cdrnet:
//cc @forki

forki:
thanks for quick fix!

Issue #3550, opened 2019-04-10
paket restore failing in 5.201.2


2sComplement:
### Description

Running paket restore on an existing solution results in an error on 5.201.2:

`Resource stream 'Paket.Restore.targets' could not be found in the Paket.Core assembly`

Running paket restore on the same project with 5.201.1 works.

### Repro steps

Run a paket restore using version 5.201.2 on an existing solution.

### Expected behavior

paket restore should run successfully.

### Actual behavior

```
Paket failed with
-> Resource stream 'Paket.Restore.targets' could not be found in the Paket.Core assembly
```

### Known workarounds

Use 5.201.1

WramblinWreck:
I'm seeing the same behavior.  Any chance this release could be rolled back?

jfrank-summit:
Me too.

sandoichi:
Same here...causing a lot of issues on our build server that uses the bootstrapper to get the latest.

mandrepont:
I agree with @WramblinWreck that this release should be rolled back or a new release should be made with a revert commit. If anyone is using paket in "Magic Mode" it breaks quite mysteriously on the build server. That is not that magic I expected 😢

baronfel:
This has to do with the names of resources: The name of the resource has changed somehow from the simple version to the fully-qualified version. For example in my work application I had to make a function to prepend the expected namespace of an embedded resource to the 'simple name', like so:

```fsharp
let resourceName simpleName = sprintf "%s.%s" "My.Assembly.Fully.Qualified.Name" simpleName
```

BlythMeister:
I made a change recently to always get the targets file...but that has been in a while (since 5.201.0), so don't think that it's that.
Maybe we should make it check simple and fully qualified names.

trumpi:
Does anyone know how to pin the paket.exe version so that we can work around this?

blair55:
@trumpi https://fsprojects.github.io/Paket/dependencies-file.html#Global-options

Zagrophyte:
@blair55 You can force paket version in the [paket.depencencies file](https://fsprojects.github.io/Paket/dependencies-file.html#Required-Paket-version)

Additionally you can run the bootstrapper.exe with the version you want as the first argument.

BlythMeister:
The "extractElement" function in https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Installation/RestoreProcess.fs is where this error is coming from.

[...]

PR #3552, opened 2019-04-10
add types to ambiguous parameters to help overload resolution

baronfel:
While working through #3550 I had a devil of a time building with more recent Mono versions than are used in the CI pipeline. This PR fixes the type-based mono failures which have arisen due to mono incorporating span overloads for certain members.


forki:
can you please send a PR that activates this mono version in CI?

PR #3553, opened 2019-04-11
use newer versions of mono (5.10 and 5.12 were from more than a year ago)

baronfel:


forki:
should we keep the old ones as well?

baronfel:
I don't think so. This is only for building, and any contributors will likely have recent mono versions.

forki:
kk

Issue #3500, opened 2019-02-11
Reference .NET Standard 2.0 assembly for Xamarin.Mac project


sergey-tihon:
### Description

Paket (5.196.2) generate incorrect reference Condition to `.NET Standard 2.0` assemblies in `fsproj` that target `Xamarin.Mac` framework - related issues https://github.com/fsprojects/Fabulous/issues/328

### Repro steps

Please provide the steps required to reproduce the problem

1. Install Fabulous template `dotnet new -i Fabulous.Templates`

2. Create new project `dotnet new fabulous-app -n SqueakyApp --macOS --Android=false --iOS=false`

3. Convert to Paket

This is what paket generate - https://github.com/sergey-tihon/MinMe/commit/c39d88f13844fe9c7e978832a1f73b439eabfe5a (repro branch)

### Expected behavior

I think that Paket should explicitly mention framework in Condition

```xml
<When Condition="($(TargetFrameworkIdentifier) == '.NETStandard' And $(TargetFrameworkVersion) == 'v2.0') 
Or ($(TargetFrameworkIdentifier) == 'Xamarin.Mac')">
```

### Actual behavior

Paket does not mention `Xamarin.Mac` in condition

```xml
<When Condition="$(TargetFrameworkIdentifier) == '.NETStandard' And $(TargetFrameworkVersion) == 'v2.0'">
```  

### Known workarounds

Manually update `fsproj` after Paket.

sergey-tihon:
Close it,  because it is works, at least for me

PR #3555, opened 2019-04-14
Xamarin.Mac supports .Net Standard 2.0

sergey-tihon:
fix for #3500

### DIFF
--- a/src/Paket.Core/Versioning/FrameworkHandling.fs
+++ b/src/Paket.Core/Versioning/FrameworkHandling.fs
@@ -601,7 +601,7 @@ type FrameworkIdentifier =
         | MonoMac -> [ DotNetStandard DotNetStandardVersion.V1_6 ]
         | Native(_) -> [ ]
         | XamariniOS -> [ DotNetStandard DotNetStandardVersion.V1_6; DotNetStandard DotNetStandardVersion.V2_0 ]
-        | XamarinMac -> [ DotNetStandard DotNetStandardVersion.V1_6 ]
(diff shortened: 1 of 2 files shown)


forki:
What about 2.1?

Sergey Tihon <notifications@github.com> schrieb am So., 14. Apr. 2019,
21:11:

matthid:
How does that fit into the table at the end of [this document](https://docs.microsoft.com/en-us/nuget/reference/target-frameworks)?

sergey-tihon:
@forki  I cannot find info about 2.1 [except this](https://devblogs.microsoft.com/dotnet/announcing-net-standard-2-1/):

> .NET Core 3.0 as well as upcoming versions of Xamarin, Mono, and Unity will be updated to implement .NET Standard 2.1.

I assume that it is not released yet.

@matthid at least based on [this table](https://docs.microsoft.com/en-us/dotnet/standard/net-standard), it is supported 
_NET_Standard___Microsoft_Docs


BTH the difference between `Modern` & `Full` version of `Xamarin.Mac` still not clear for me
Fullscreen_4_15_19__10_13_AM
Fullscreen_4_15_19__10_14_AM

matthid:
Question is how do they differ in the NuGet package. Maybe they have different monikers or the moniker is versioned now?
What about people using older versions of Xamarin and a newer version of Paket (including this PR). That will lead to issues, correct? Or are all Xamarin users forced to use latest somehow?

forki:
yeah it's weird that they only have one moniker. that's nuts

sergey-tihon:
I have no idea, tbh, I think depends on the IDE version that you have (latest available at the install time). I just know that latest Paket generate incorrect assembly references nowadays.

Do you know Xamarin expert that can consult us ? // cc @TimLariviere

PR #3558, opened 2019-04-19
consider dependencies target framework restrictions in paket pack

enricosada:
Scenario: there are framework restrictions (like `framework: net461`) in paket.dependencies or paket.references

The `paket pack` should consider these and filter dependencies for the nuspec target framework groups

Before, all packages where added in all target framework groups


enricosada:
@forki the travis failure is unrelated, the build job doesnt start because cannot install the .net core sdk with `apt`

forki:
replaced by https://github.com/fsprojects/Paket/pull/3573

PR #3559, opened 2019-04-19
[WIP] trying to tame the slow tests

enricosada:
### step 1: identify slow tests

set test timeouts

- tests: 2 sec
- integration tests: 20 sec

PR #3573, opened 2019-05-08
Paket pack by tfm

forki:
subsumes https://github.com/fsprojects/Paket/pull/3558


forki:
@enricosada I merged with master. It's red!?

enricosada:
it's red because appveyor job timeout after 1h.
I added one test, but tests anyway where pretty near 1h

that's why i opened https://github.com/fsprojects/Paket/pull/3559 to check what are the really slow ones

You can rerun the appveyor job if happen, sometime will go faster

PR #3585, opened 2019-05-24
fix multiple dotnet pack invocations with different Version property

### LINKED ISSUE #3584: Paket fails when running `dotnet pack` on a library after its version was changed.
### Description

When the version of a library changes and a `dotnet pack` is invoked again, Paket gets confused, because there are many `*.nuspec` files in `<project-directory>/obj/<configuration>`: one for the old version, and another for the new one.

### Repro steps

1. Create a library powered by Paket.

2. Set it a version in the project file, say `1.0.0`.

3. `dotnet pack`.

4. Set it to another version, say `2.0.0`.

5. `dotnet pack` once again.

If possible then please create a git repository with a repro sample or attach a zip to the issue.

### Expected behavior

A `nupkg` file with the new version.

### Actual behavior

Paket fails with an incorrect path error.

A screenshot from my project exhibiting the bug

### Known workarounds

Delete the `nuspec` files before the packing. However, .NET Core SDK is smart enough to keep both of them and pick the correct one.

### Additional info

* The path Paket complains about might give us some insight. It is a semicolon-delimited list of the `nuspec` of a package version, and _the `obj/<configuration>` directory_.

* ~~I ran an experiment: I temporarily set my library's version to 2.0.0, without deleting the `obj` folder, and tried to pack again. The faulty path had the `nuspec` of version _4.0.0_; the _chronologically_ earliest, not the lexicographically one.~~ Actually, it was built for version 4.0.0, so never mind.

enricosada:
fix #3584


integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -739,6 +739,36 @@ let ``#4003 dotnet pack of a global tool with p2p``() =
...
+    |> ignore
+
+    let pack outDir additionalArgs expectedVersion =
+        let outPath = Path.Combine(rootPath, outDir)
+
+        directDotnet true (sprintf "pack -o \"%s\" %s" outPath additionalArgs) rootPath
+        |> ignore
+
+        let nupkgPath = Path.Combine(outPath, sprintf "%s.%s.nupkg" project expectedVersion)
+        if File.Exists nupkgPath |> not then Assert.Fail(sprintf "Expected '%s' to exist" nupkgPath)
```
teo-tsirpanis: You could have used `Assert.True` here.

integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -739,6 +739,36 @@ let ``#4003 dotnet pack of a global tool with p2p``() =
...
+
+    pack "out1" "" "1.0.0"
+    pack "out1" "/p:Version=1.2.3" "1.2.3"
+    pack "out2" "/p:Version=2.3.4" "2.3.4"
+
+
```
teo-tsirpanis: These lines seem a bit off.

integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -739,6 +739,36 @@ let ``#4003 dotnet pack of a global tool with p2p``() =
...
+[<Test>]
+let ``#4004 dotnet pack using different versions``() = 
+    let project = "lib1"
+    let scenario = "i004004-pack-version"
+    prepareSdk scenario
+
+    let rootPath = scenarioTempPath scenario
+
+    directPaket ("restore") scenario
```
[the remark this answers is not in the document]
enricosada: done

integrationtests/Paket.IntegrationTests/PackSpecs.fs
enricosada: done

integrationtests/Paket.IntegrationTests/PackSpecs.fs
enricosada: done

enricosada:
@teo-tsirpanis thx for review, good feedback.

Issue #3535, opened 2019-03-28
Could not acquire lock to ...\paket.locked


ivanbenko:
### Description

I've experienced very strange `paket restore` behaviour today. I was constantly getting this error message on restoring packages: `Could not acquire lock to C:\Gitlab-Runner\builds\cc2bad82\0\***\obj\paket.locked`. Finally I found out it was caused by blocking the creation of `*.locked` files by my antivirus (McAfee) because of some virus which encrypts files and creates *.locked copy of it I guess. 

Since our organization policy doesn't allow us to change any security settings it becomes unreal to run my build pipeline.

Please advice. Is it possibly to make `paket.locked` constant configurable or rename it to smth else? Thanks

### Known workarounds

I made a custom build with changed to `paket.locked-whatever` constant (located in `Constants.fs`) instead of `paket.locked`.

forki:
I think the correct fix is to tell McAfee about it. It's a bug in their
scanner.

Ivan Bianko <notifications@github.com> schrieb am Do., 28. März 2019, 22:32:

Lordinaire:
We've got the same behaviour in my company and it's pretty annoying. It's not _directly_ related to Paket but we planned to remove Paket because of that :(

forki:
all I can say: tell your antivirus vendor. They usually whitelist pretty fast

mjirous:
@forki Files ending with *.locked extension are treated by McAfee as ransomware. It may be stupid, but that's how McAfee works. I issued a ticket to whitelist this file in our company and after a week we only got temporary approval. I agree that it McAfee problem, but many companies use it and security experts are just afraid. We are basically forced to come with some alternative because our company does not want to allow it permanently. The alternative may be to not use Paket or have a custom fork. None of these options are good for us. It would be much better to just change the single line of code in Paket to change the filename. Is there some specific reason the file must have the name `paket.locket`?

forki:
I'm fine with changing the name

Michal Jirouš <notifications@github.com> schrieb am Fr., 24. Mai 2019,
16:15:

ivanbenko:
@forki would you mine to name it `paket.lockfile` or smth else? I can create a PR for it.

matthid:
I'm fine with renaming the file as long as the change is backwards compatible on restore.

forki:
I think it should be something like .processlock

daz10000:
[...]

PR #3586, opened 2019-05-28
Rename paket.locked file to .processlock to avoid McAfee scanner to evaluate the file as threat

mjirous:
Files ending with *.locked extension are treated by McAfee as ransomware. It's not always easy to get the file to Whitelist (especially in big companies) and therefore it is much easier to simply rename the file causing troubles.

I renamed the file paket.locked to .processlock as suggested by @forki 

The original discussion is here: #3535


src/Paket.Core/Common/Constants.fs
```diff
@@ -24,7 +24,7 @@ let [<Literal>] BootstrapperFileName      = "paket.bootstrapper.exe"
 let [<Literal>] PaketFileName             = "paket.exe"
 let [<Literal>] TargetsFileName           = "paket.targets"
 let [<Literal>] ReferencesFile            = "paket.references"
-let [<Literal>] AccessLockFileName        = "paket.locked"
+let [<Literal>] AccessLockFileName        = ".processlock"
```
forki: I think it should be `paket.processlock` - just in case people have gitignore rules like `paket.*` (even if this file is not really important for git)

forki: don't worry. will do that right now

PR #3588, opened 2019-05-28
Allow for nuget v3 api to have an url that does not end with /index.json

keroiber:
This change also helps to circumvent #3575 by using v3 api

### DIFF
--- a/src/Paket.Core/Versioning/PackageSources.fs
+++ b/src/Paket.Core/Versioning/PackageSources.fs
@@ -150,7 +150,7 @@ type PackageSource =
 #endif
                     LocalNuGet(source,None)
                 else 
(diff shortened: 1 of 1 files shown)


forki:
hopefully this does not match too much

keroiber:
A better solution would be to just try and read the nuget resources json from the provided url (without any regex). On success use that api and otherwise fall back to v2. But I cannot provide a solution for that because I'm a complete beginner with F#.

forki:
does this still match things that end in index.json?

keroiber:
Your'e right, it does not. I'll prepare a new pull request.

keroiber:
https://github.com/fsprojects/Paket/pull/3590

PR #3590, opened 2019-05-28
Allow for nuget v3 api to have an url that does not end with /index.json - third try

keroiber:
Reduce the check for v3 api to searching the string "/v3/" in the url

This change also helps to circumvent #3575 by using v3 api


forki:
thansk!

PR #3636, opened 2019-08-10
Support creating snupkg symbol packages

teo-tsirpanis:
This PR would allow specifying a custom SymbolPackageFormat value to PackTask.

I also fixed a typo in line 10 of Paket.Restore.targets. I presume it was by accident; in MSBuild, quotes are needed for equality checks against the empty string.


baronfel:
Cool! This needs to be applied to the targets in src/Paket.Core/embedded though, because that's the version that gets distributed to everyone.

teo-tsirpanis:
@baronfel, done.

Issue #3660, opened 2019-09-23
Add support for using paket as a .NET Core 3 local tool


tboby:
### Description

.NET Core 3 adds support for repository specific tools that are restored with the correct version for your platform. These are defined in a manifest called `dotnet-tools.json`, a lockfile for dotnet "local" tools. 

This removes the need to commit paket.exe, removes the need for mono, and generally makes it really easy to install paket.

Paket can be defined as a dependency in this way, correctly restores, and can be manually run using `dotnet paket` or `dotnet tool run paket`. 

However, `Paket.Restore.targets` does not seem to know about local tools.

### Repro steps
1. clone `https://github.com/tboby/paket-local-tool-sample`

2. `dotnet tool restore`

3. `cd src`

4. `dotnet build`

### Expected behavior

The project should trigger a paket restore, then build.

### Actual behavior

```
  '"paket.exe"' is not recognized as an internal or external command,
  operable program or batch file.
C:\Users\Thomas\Git\paket-local-tool-sample\.paket\Paket.Restore.targets(138,5): error MSB3073: The command ""paket.exe" restore" exited with code 9009. [C:\Users\Thomas\Git\paket-local-tool-sample\src\paket-local-tool-sample.fsproj]
```
### Known workarounds

Add `<PaketCommand>dotnet paket</PaketCommand>` to the project file.

forki:
I'd be happy to get PRs

Thomas Boby <notifications@github.com> schrieb am Mo., 23. Sep. 2019, 10:24:

tboby:
As far as I can tell msbuild doesn't know about local tools, and the path to the local tool probably isn't consistent (e.g. `C:\Users\tboby\.dotnet\toolResolverCache\1\paket`.
Would it be a reasonable approach to try to use an `Exec` task to run the local tool, and use the exit code to determine if it's available?

kjohnphillip:
It appears that PaketCommand in Paket.Restore.targets is missing the case for Paket as a local tool i.e. when PaketExePath does not have an extension. Inserting the following line temporarily fixes the issue, until that file is overwritten:
<PaketCommand Condition=" '$(PaketCommand)' == '' AND '$(_PaketExeExtension)' == '' ">dotnet "$(PaketExePath)"</PaketCommand>

I'm happy to create a PR for this if anyone can point me towards where best to extend any tests that cover the state of that file.

atlemann:
[...]

PR #3668, opened 2019-09-27
3660: Add PaketCommand contidion for Paket installed as .NET Core 3.0 local tool

atlemann:
Fixes: #3660


forki:
thanks

atlemann:
Np, I feel I just stole the other guys fix though... 

Thanks for merging fast!

Issue #3669, opened 2019-09-27
paket dependency order importance


BlythMeister:
### Description

The order of dependencies in a paket.dependency file is important when using `strategy:min`
Even though the docs state:
`Note, however, that all direct dependencies will still get their latest matching versions, no matter the value of the strategy option. If you want to influence the resolution of direct dependencies then read about the lowest_matching option.` this is NOT always the case.

### Repro steps

Create a repo with dependency of

```
source https://api.nuget.org/v3/index.json
framework: auto-detect
storage: none
strategy: min

nuget Castle.Core
nuget Castle.Windsor
```

Run a paket install, everything looks good, you have castle.core with the latest version.

Flip the 2 dependencies round.

Castle.Core is downgraded to the miniumum required by Castle.Windsor

### Expected behavior

Order *is not* important in a dependencies file

### Actual behavior

Order *is* important in a dependencies file

### Known workarounds

Get things in the right order...whatever that may be ;)

forki:
Where do we say it's not important? It definitely is

Chris Blyth <notifications@github.com> schrieb am Fr., 27. Sep. 2019, 11:43:

BlythMeister:
i guess when using max, it's not as important...i guess i've always assumed it's not important :(

but the docs on min do say that direct dependencies will get `their latest matching versions, no matter the value of the strategy option` 
But this is not the case if you have something with the dependency as transitive above it...

BlythMeister:
i _think_ it's because as soon as a package is explored, even if it's a top level dependency, we never get latest.
Therefore meaning this statement is actually not correct.

matthid:
This is definitely unexpected...

BlythMeister:
I have a fix...I think 🤣

PR #3670, opened 2019-09-27
Update getResolverStrategy for rootRequirements

BlythMeister:
When root requirement, us the resolver strategy for direct dependencies.
Even if it also has other requirements

Attempted fix for #3669 

And meaning that the statement `that all direct dependencies will still get their latest matching versions, no matter the value of the strategy option. If you want to influence the resolution of direct dependencies then read about the lowest_matching option.` in the docs is actually correct

### DIFF
--- a/tests/Paket.Tests/Resolver/GlobalOptimisticStrategySpecs.fs
+++ b/tests/Paket.Tests/Resolver/GlobalOptimisticStrategySpecs.fs
@@ -140,7 +140,7 @@ let ``should favor strategy from parent when it overrides``() =
     let resolved =
         DependenciesFile.FromSource(config)
         |> resolve graph2 UpdateMode.UpdateAll
-    getVersion resolved.[PackageName "Castle.Windsor"] |> shouldEqual "3.2.1"
+    getVersion resolved.[PackageName "Castle.Windsor"] |> shouldEqual "3.3.0"
     getVersion resolved.[PackageName "Castle.Windsor-NLog"] |> shouldEqual "3.3.0"
     getVersion resolved.[PackageName "Castle.Core-NLog"] |> shouldEqual "3.3.0"
     getVersion resolved.[PackageName "Castle.Core"] |> shouldEqual "3.3.1"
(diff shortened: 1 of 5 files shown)


BlythMeister:
@forki this change does update tests...but i think that the tests contradict the documented behaviour.

In these tests, the dependencies are specified in the paket.dependencies file
However, we do not expect to get the latest version

BlythMeister:
Is there anything more needed from me to get this merged?
Looks like the build went green 🙂

tests/Paket.Tests/Resolver/GlobalOptimisticStrategySpecs.fs
```diff
@@ -159,7 +159,7 @@ let ``should favor strategy from parent that overrides strategy``() =
     let resolved =
         DependenciesFile.FromSource(config)
         |> resolve graph2 UpdateMode.UpdateAll
-    getVersion resolved.[PackageName "Castle.Windsor"] |> shouldEqual "3.2.1"
+    getVersion resolved.[PackageName "Castle.Windsor"] |> shouldEqual "3.3.0"
```
forki: why is this increased? I thought we use min strategy?

BlythMeister: this test:
```
[<Test>]
let ``should favor strategy from parent that overrides strategy``() = 
    let config = """
    strategy max
    source http://www.nuget.org/api/v2
    nuget Nancy.Bootstrappers.Windsor !~> 0.23
    nuget Castle.Windsor
    nuget Castle.Windsor-NLog !> 0
    """
    let resolved =
        DependenciesFile.FromSource(config)
        |> resolve graph2 UpdateMode.UpdateAll
    getVersion resolved.[PackageName "Castle.Windsor"] |> shouldEqual "3.3.0"
    getVersion resolved.[PackageName "Castle.Windsor-NLog"] |> shouldEqual "3.3.0"
    getVersion resolved.[PackageName "Castle.Core-NLog"] |> shouldEqual "3.3.0"
    getVersion resolved.[PackageName "Castle.Core"] |> shouldEqual "3.3.0"
    getVersion resolved.[PackageName "Nancy.Bootstrappers.Windsor"] |> shouldEqual "0.23"
```
is set to max but more importantly has Castle.Windsor as a top level dependency.  So should take the latest version

forki:
thank you

PR #3672, opened 2019-09-28
3671: Create a "SetPaketCommand" target to be able to try Exec the local cli tool

atlemann:
Suggested by @NatElkins.

Fixes: #3671

PR #3706, opened 2019-11-09
Change cli target detection issue 3705

### LINKED ISSUE #3705: "dotnet paket --version" hangs until forcefully closed
#### Short version

Some threads hang at `dotnet paket --version` and never moves on. This results in builds hanging because some projects never gets build; Visual Studio won't unload projects or ReSharper fails to load; possibly also causing other faulty behaviors.

#### Longer version

We've found that having the CPU max out sometimes causes `dotnet paket --version` commands to hang a process. It seems to be some kind of timing issue, but we are uncertain if the issue is in dotnet faulting on the output from paket, og if paket simply never exists correctly.

The problem seems to only happen when using the new SDK project format (which is the one leveraging dotnet to begin with), as we do not have the same issue with our large Full Framework projects that have not yet been upgraded.

Also, having anti virus monitoring file changes, as well as using ReSharper, seems to increase the likelyhood dramatically of it happening.

We started noticing dotnet processes not closing down again after calling `dotnet paket --version`, and they seem to be what is blocking all the processes. Even more odd is, that when inspecting the process with [Process Explorer](https://docs.microsoft.com/en-us/sysinternals/downloads/process-explorer) and going to the Threads tab, the process finally closes down. What exactly causes the process to close down from that inspection we have still not uncovered, as listing the threads of the process via PowerShell does not yield the same result.

### Repro steps

1. Have a large solution (in our case 230 projects) in Full Framework, but in the new SDK format.
2. Have ReSharper active in Visual Studio.
3. Work for a bit in Visual Studio, building and navigating around for 4-5 minutes.
4. Attempt to unload the solution

### Expected behavior

Visual Studio should simply unload the solution.

### Actual behavior

Unloads hangs at "Unloading project 0 of 230 ..." and never quits (have tried leaving it for multiple hours). In Process Explorer a process will be running with the command `dotnet paket --version`.

### Known workarounds

Either:
1. Shut down the hanging process by going to the threads tab, or forcing the process to exit.
Or:
2. Removing the part in the `paket.targets` file that actually calls `dotnet paket --version`, and replacing it with calling the local paket.exe always instead. This get overridden however when `paket install/update` is called, and must be corrected again, and there are no switches to prevent it from happening.

### Final thoughts

This issue has been downright cripling for our development department, as everything slowed down to an almost standstill. Different developers saw the bug more or less frequently, but some had an unresponsive Visual Studio for 10-15 minutes at a time.

We do not know, however, if the error is with dotnet instead of directly with paket, but it is only in combination of the two we have the error.

The best solution, for us right now, I believe is to have some sort of flag, that would brevet paket to get asked all the time.

I hope the issue makes somewhat sense, and that a more permanent solution can be found, rather than we manually fix the targets file, time and time again.. :)

JohnyWS:
This is only a workaround, that will solve most usecases, but still leave the problem intact for installations dependent upon the CLI. So this only ensures that all other installations that does not rely on the CLI approach experiences the bug that supporting it introduces.

I've changed 2 things to work around the issue:

1. Introduced "PaketDisableCliTest" which, when set to "True", will disable the CLI test completely. This can then be set in the projects or in system variables.
   Is that an acceptable approach? If yes, where should I document the feature?
2. Change the priority in which paket resolves. Previously it was "CLI => root => tool => proj style => bootstrapper => global" but that has now been changed to "root => tool => CLI => proj style => bootstrapper => global"
   Is that an improved priority? If not, why? If yes, should CLI be moved even further back? If yes, how long - and can that be moved to its' own issue if that is the case?

Also, when changing the file, I've added a few more lines of comments, and tried to align things a bit more.

In the main commit (7e68928) I have also removed a few instances where the Windows parts used ".exe" which does not seem to be needed when testing, resulting in fewer cases to test for.

In the secondary commit (d66711e) I have removed a case that seemed to be obsolete by now - non-CLI approach where paket is a dll. It appears that no path through target file will hit that logic. That allowed me to isolate the extension testing to the only case that needed it, so the test only runs in that case. Lastly I renamed "Internal" to "_" to match how "_PaketExeExtension" is named.

Fixes #3705

### DIFF
--- a/.paket/Paket.Restore.targets
+++ b/.paket/Paket.Restore.targets
@@ -20,7 +20,7 @@
     <PaketBootstrapperStyle Condition="Exists('$(PaketToolsPath)paket.bootstrapper.proj')">proj</PaketBootstrapperStyle>
     <PaketExeImage>assembly</PaketExeImage>
     <PaketExeImage Condition=" '$(PaketBootstrapperStyle)' == 'proj' ">native</PaketExeImage>
-    <MonoPath Condition="'$(MonoPath)' == '' And Exists('/Library/Frameworks/Mono.framework/Commands/mono')">/Library/Frameworks/Mono.framework/Commands/mono</MonoPath>
+    <MonoPath Condition="'$(MonoPath)' == '' AND Exists('/Library/Frameworks/Mono.framework/Commands/mono')">/Library/Frameworks/Mono.framework/Commands/mono</MonoPath>
     <MonoPath Condition="'$(MonoPath)' == ''">mono</MonoPath>
 
     <!-- PaketBootStrapper  -->
@@ -28,7 +28,7 @@
     <PaketBootStrapperExePath Condition=" '$(PaketBootStrapperExePath)' == '' ">$(PaketToolsPath)paket.bootstrapper.exe</PaketBootStrapperExePath>
     <PaketBootStrapperExeDir Condition=" Exists('$(PaketBootStrapperExePath)') " >$([System.IO.Path]::GetDirectoryName("$(PaketBootStrapperExePath)"))\</PaketBootStrapperExeDir>
     
-    <PaketBootStrapperCommand Condition=" '$(OS)' == 'Windows_NT'">"$(PaketBootStrapperExePath)"</PaketBootStrapperCommand>
+    <PaketBootStrapperCommand Condition=" '$(OS)' == 'Windows_NT' ">"$(PaketBootStrapperExePath)"</PaketBootStrapperCommand>
     <PaketBootStrapperCommand Condition=" '$(OS)' != 'Windows_NT' ">$(MonoPath) --runtime=v4.0.30319 "$(PaketBootStrapperExePath)"</PaketBootStrapperCommand>
 
     <!-- Disable automagic references for F# dotnet sdk -->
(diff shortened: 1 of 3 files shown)


JohnyWS:
My deduction in the second commit was apparantly faulty. I'll look into it.

atlemann:
Another way to check for local clitool is like this:
https://github.com/fsprojects/Paket/pull/3672#issuecomment-536210446

Basically reading the tool-manifest.json file. That would avoid invoking the tool. It's not as bullet proof though.

atlemann:
This has become a bit of a mess due to all the different ways paket has been bootstrapped over time. It would have been nice to just have local or global clitool and ditch the rest, but that will probably break a lot of projects out there.

When adding support for local clitool, we just added all the previous steps as they were after checking for local to avoid breaking things. My first attempt broke global tool support...

atlemann:
I guess it's fine to check for local clitool after all the searching inside .paket folder, but we have to prioritize local clitool before global tool.

JohnyWS:
@atlemann what are the downsides of that approach? Because if it only supplies false negatives (saying the CLI tool isn't there when it is) we can just add that as an extra step before attempting with `dotnet paket --version`.

I've removed my second commit, and trying again. Debugging is hard when I can't get it to run locally.

I'm trying to use VS Code on a Windows machine, but I can't seem to run the tests, and I've got all sorts of things that won't load correctly. Any thorough guide on getting a dev environment to run?

atlemann:
When I tried to fix it the last time I was just doing manual testing by changing the paket.Restore.targets file in one of my own projects and setting `export PAKET_SKIP_RESTORE_TARGETS=true` and installing paket as global and then as local tool etc. and seeing if they all worked.

atlemann:
Don't think there are any downsides by checking the tools-manifest except false negatives as you said. Then running `dotnet paket --version` could maybe be the last check if none of the others find anything.

forki:
Is this ready?

Atle Rudshaug <notifications@github.com> schrieb am So., 10. Nov. 2019,
07:01:

atlemann:
I don't think so. I haven't had time to try it, only looked at it on my phone. And I think maybe we should add that parsing of tools.json first to avoid invoking dotnet all the time.

forki:
Can someone remind me. What are we doing with the Paket version in the targets file?

atlemann:
Nothing, it's just used as a means to see if the process fails or not. Like, what's the cheapest paket command we can call to see if local tool invocation succeeds.

forki:
I see. So we don't really need to parse the file right? Just check if it
exists and look for a substring. Might be much faster and easier to do.

Atle Rudshaug <notifications@github.com> schrieb am So., 10. Nov. 2019,
09:39:

atlemann:
Yes, IMHO we could replace invoking dotnet paket with checking the manifest file e.g. as shown in that comment I linked to. There might be better ways to check a file for some substring I msbuild though.

atlemann:
And we should update paket.targets to keep them in sync.

JohnyWS:
Ok, I'll try to sum up status for now:

**Primary concerns:**

1. Searching for the CLI can be improved by looking at `dotnet-tools.json` as @atlemann suggested, and I believe that will reduce the times it errors even more - I'll look into it right away.
1. `paket.targets` should be updated too - I can't find the source for that file though? Can anyone point me towards that?

**Secondary concerns:**

1. The search order this PR introduces is: "root => tool => CLI => proj style => bootstrapper => global", this keeps CLI before global, but lets local versions in root and tool folder take precedence. That only leaves "proj style / bootstrapper" - if they too should have precedence over CLI, I think a new issue should be created to tackle that. I think the same arguments for changing the order for root/tool could be made for bootstrapper as well. But for now, this PR changes nothing for bootstrapper projecets.
1. Removing the option that changed .dll to force the use of dotnet made the build the fail, so it is obviously being used. I'll try to see where that is, but for now, that can be ignored.

JohnyWS:
Oh, and forgot another point - is adding "PaketDisableCliTest" something that should be introduced or not? I haven't tested it out yet (I do not have access to the code here that uses paket), and I'm wondering if it is, at all, useful anymore, with the other fixes in place?

JohnyWS:
Hmm, there's a lot of failures under the "Allowed failures" section. Is that because they are flaky, or?

forki:
Yes flaky tests that are on the backlog

Johny Woller Skovdal <notifications@github.com> schrieb am So., 10. Nov.
2019, 16:07:

JohnyWS:
Well, then I actually think this is a wrap for the main concern.

That leaves:

1. docs - where in the docs does this fit? I think the priority would be important to document? I'll gladly do it, if you have a good suggestion as to where to integrate it? :)
1. "PaketDisableCliTest" - should that be kept or removed? If kept, it should be tested and documented - docs I'll gladly add in the above docs, but I've yet to have testing work for me. :-/
1. `paket.targets` - where do I find that? If it's long behind the other, I think that should be moved to a separate issue.and the

Am I missing something? :)

forki:
paket.targets is only found in:

image

@atlemann @matthid are we ok with merging this in?

atlemann:
Nope, I'm getting this error

```
error MSB3073: The command ""paket" restore" exited with code 127.
```

when setting

```
<PaketDisableCliTest Condition=" '$(PaketDisableCliTest)' == '' ">True</PaketDisableCliTest>
```

which means that the `dotnet-tools.json` check doesn't work.

JohnyWS:
@forki ah, my bad, I thought that was for setting up .paket for the paket repo itself. Is that target file what is distributed to Full Framework installations using the old proj format then?

@atlemann well that proves the switch works at least. ;) So in your environment, you do have the `dotnet-tools.json` placed here relative to the target file: `..\.config\.dotnet-tools.json` - plus it contains the word "paket" somewhere in it? Because that's what the test is supposed to do at the moment. :)

atlemann:
Yes, there's where I have it and it contains paket. Probably just a silly little thing.

JohnyWS:
@atlemann it was indeed... Can you try it out again with the latest change? :)

atlemann:
@JohnyWS Seems to work now.

As for paket.targets, we just copied over the `SetPaketCommand` target the last time, so I guess it should just be updated with the same changes you made here.

JohnyWS:
@atlemann I pushed the changes to both target files in the .paket folder now. Can you double check it still works for you? :)

@forki and @atlemann : what about docs - or can we postpone that part? Would really like this change, because everytime someone updates a package, it overrides our local hack... :)

forki:
I'm fine with release it as soon as we get that second sign off.

Johny Woller Skovdal <notifications@github.com> schrieb am Di., 12. Nov.
2019, 11:25:

atlemann:
I'm getting this for some reason:
```
.paket\paket.targets(45,11): error MSB4100: Expected "!$(PaketDisableCliTest)" to evaluate to a boolean instead of "!", in condition " '$(PaketExePath)' == '' AND !$(PaketDisableCliTest) AND !$(_ConfigContainsPaket)".
```

atlemann:
Hmm...seems to be an issue when building from VisualStudio, but doing `dotnet build` from the command line doesn't give that error.

JohnyWS:
Good catch! I'll look into it right away!

tir. 12. nov. 2019 12.45 skrev Atle Rudshaug <notifications@github.com>:

atlemann:
A solution with only new SDK proj files which uses `paket.Restore.targets` works both with CLI `dotnet build` and from VisualStudio build, so seems to be something weird with old SDK format.

JohnyWS:
@atlemann Hmm, I seem to be needed a lot more details to reproduce it exactly.

The case I've tried:
1) .NET 4.7.2 project with auto-restore on and old proj format
2) built from commandline with msbuild with or without full `git clean -xfd`
3) build from Visual Studio with or without full `git clean -xfd`

I could not make it fail like you did. Any details I'm missing?

atlemann:
Hmmm... I just copy-pasted the raw paket.targets file from this PR into my solution and built in VS2019. I can try to do a proper clean and try again tomorrow at work.

atlemann:
It seems to work. Not sure what happened yesterday, but I just tried copying it over again today and now it builds.

forki:
So ready to go live?

Atle Rudshaug <notifications@github.com> schrieb am Mi., 13. Nov. 2019,
08:52:

JohnyWS:
I believe so yes. Any objections @atlemann ? :)

atlemann:
No, it should be fine.

matthid:
It would be nice to sync such changes with a fake release, but I have no idea how that would look like in practice...

forki:
thanks! it's released

atlemann:
Thanks! I just realized that our team working on this massive solution we have, had this same issue with the build hanging randomly. They just didn't tell me...and it didn't happen for me the few times I have built it lately. Works better now!

forki:
hehe. awesome!

PR #3713, opened 2019-11-11
Remove FSharp.Compiler.Tools and move to SDK based projects

forki:
replaces #3569


paket.lock
```diff
@@ -24,7 +24,6 @@ NUGET
       FSharp.Core (>= 4.1) - restriction: < netstandard1.6
       FSharp.Core (>= 4.1.17) - restriction: >= netstandard1.6
       NETStandard.Library (>= 1.6.1) - restriction: >= netstandard1.6
-    FSharp.Compiler.Tools (4.1.23)
```
forki: /cc @cartermp :P

integrationtests/Paket.IntegrationTests/PackSpecs.fs
```diff
@@ -586,11 +586,6 @@ let ``#2694 paket fixnuspec should not remove project references``() =
     | None -> Assert.Fail("Expected package to still contain the FSharp.Core reference!")
     | Some s -> ignore s
 
-    // Should we remove Microsoft.NETCore.App?
```
forki: @matthid sorry to bother you mate. This is the only integration test that is red. I tried to port everything to the SDK and somewhere here in this PR there is something that changes nuspec.Dependencies.Value.Length from 3 to 2.
The comment above the assertion is already very suspicious, but still I'm wondering what changed. Any ideas?

matthid: Can you you check and post what was in the list before and what is there now to see what has disappeared?

Github seems to make it a bit harder to see the code when force pushes are used apparently...

forki: Dependencies: 

    [(PackageName ("library","library"), VersionRequirement (Minimum 1.0.0,No),
      ExplicitRestriction >= netcoreapp2.0);
     (PackageName ("FSharp.Core","fsharp.core"),
      VersionRequirement (Minimum 4.2.3,No), ExplicitRestriction >= netcoreapp2.0);
     (PackageName ("Microsoft.NETCore.App","microsoft.netcore.app"),
      VersionRequirement (Minimum 2.0.0,No), ExplicitRestriction >= netcoreapp2.0)]

so it's removing Microsoft.NETCore.App

matthid: I think Microsoft.NETCore.App is indeed special in a way. 
Question is if we have added some kind of special casing for it in the code-base?
Seems like it was magically added before this PR, because in the lockfile of this test it is not even referenced, see https://github.com/fsprojects/Paket/blob/fullsdk/integrationtests/scenarios/i002694/before/paket.lock
So I guess it is "safe" that it is no longer there. I'm more wondering why it was there before this PR, which only really changes project setup?

Maybe some DEFINES are not like they where before?

forki:
Yes. I tried convert the compiler switches but maybe I lost one.

Matthias Dittrich <notifications@github.com> schrieb am Do., 14. Nov. 2019,
15:56:

forki:
I also updated the sdk to 3.0.100. Maybe that's relevant?

Steffen Forkmann <sforkmann@gmail.com> schrieb am Do., 14. Nov. 2019, 16:05:

integrationtests/Paket.IntegrationTests/PackSpecs.fs
forki: Yes. I tried convert the compiler switches but maybe I lost one.
I also updated the sdk to 3.0.100. Maybe that's relevant?

forki: ok confirmed. I updated the sdk in the test PR (see https://github.com/fsprojects/Paket/pull/3718) and the dependency is gone as well.

matthid: Ah because we use the SDK nuspec output as input for ourself. It seems like they have decided to filter this one. So at least that confirms that we should filter as well if needed?

enricosada:
@forki pushed some commit to use `dotnet-mergenupkg` v3

checking the repo, things who can be done:
- use `--no-build` after the first `dotnet build paket.sln` to speedup, because pack/test is rebuilding (i tried to do it, dunno why failed because build was needed, something touched the files)
- use dotnet.sourcelink, instead of sourcelink
- use `dotnet test` for quicktest, instead of nunit-console
- make `paket` a normal .net tool (`PackAsTool = true`), and include the `paket.exe` as content. This will deprecate `dotnet-mergenupkg`

i dont know if i can contribute these, but i listed it anyway

isaacabraham:
@enricosada can you create issues for these?

enricosada:
@isaacabraham i fixed these in PR #3734

Issue #3710, opened 2019-11-11
paket.template conformance


cartermp:
Currently, `paket.template` is not conforming with NuGet server changes. Here's a checklist to get into conformance:

License stuff:
- [ ] Add support for `license` (#3475)
- [ ] Add support for [`licenseFile`](https://docs.microsoft.com/nl-nl/dotnet/core/tools/csproj#packagelicensefile) (probably a path to the license file?)
- [ ]  Deprecate `licenseUrl` ([deprecated in NuGet](https://docs.microsoft.com/nl-nl/dotnet/core/tools/csproj#packagelicenseurl))

Icon stuff:
- [ ] Add support for [`icon`](https://docs.microsoft.com/en-us/nuget/reference/nuspec#icon) (probably a path to the icon file?)
- [ ] Deprecate `iconUrl` ([deprecated in NuGet](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packageiconurl))

Shmew:
Pitching in what I've noticed: paket.template also doesn't appear to follow the same behavior when generating nuspecs, as newest nuget always adds exclude="Build,Analyzers".

PR #3715, opened 2019-11-11
[WIP] Add license expression and icon to template spec

cartermp:
My attempt at #3710, but minus the deprecation parts


forki:
what's the status?

cartermp:
I need to figure out how to actually pack files with paket.template. Can't seem to figure it out :/

CaptnCodr:
Can we go on here? These nuget deprecation hints are annoying.

baronfel:
@CaptnCodr you can also put the license expression in your project file and pack with `dotnet pack` these days, you don't need to use paket.template.

CaptnCodr:
@baronfel How about the deprecated "iconUrl"?

baronfel:
@CaptnCodr https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#packageicon this help doc describes what properties you'd use to set an image, along with a number of others. note that nuget now wants you to embed the icon instead of pointing to an external resource.

cartermp:
This one should be easy enough, I just kind of gave up on some of the integration tests largely because...I couldn't write a proper paket.template file :/

CaptnCodr:
Yes, I see. But we only use the paket.template to generate our nuget packages.
But why isn't it implemented like "license" which is the reaction on the deprecation of "licenseUrl" so how about to go on here and you have one PR from the list and an issue from the other? 😄

cartermp:
I think we'd all be happy to take a PR that only does each one at a time.

cartermp:
`license` is also added as per here: https://github.com/fsprojects/Paket/pull/3824

CaptnCodr:
Ah yes.

PR #3735, opened 2019-11-28
Add back icon

enricosada:
add icon to paket.exe

to add icons with `fsc` it's possible to use native resources like 

```xml
<Win32Resource>paket.res</Win32Resource>
```

the `ApplicationIcon` property is not supported either in `fsproj`/`fsc` ( ref https://github.com/dotnet/sdk/issues/2818 ), so cannot do

```xml
<ApplicationIcon>paket.res</ApplicationIcon>
```

The issue is that the `fsc` running on `coreclr` doesnt support yet native resources (ref https://github.com/dotnet/fsharp/issues/1172 ), so native resources doesnt work in .NET Core Sdk

As workaround, a free tool (http://angusj.com/resourcehacker/) is used to add the icon after the exe is generated (windows only)

Other fix, unrelated, fix `net461` `boostrapper.exe` inside `Paket.Bootstrapper` package


enricosada:
💄 

image

For me is done @forki , obv minus CI/feedback/review

forki:
hugs

PR #3737, opened 2019-11-30
`WIP` Paket global and local tools

enricosada:
modify `Paket` too so can be installed global (`-g`) and local (`--local`) with different behaviour.

as global tool, installed like `dotnet tool install -g Paket`, behaviour depends:
- if the working directory is a repo with paket (`.paket/paket.exe` OR `dotnet paket`), it execute the repo paket exe, passing the same arguments (sort of magic mode). so `paket --help` is like `.paket/paket.exe --help`
- if the working directory is not a repo with paket, just a subset of commands exists. like `paket init`.

as local tool, installed like `dotnet tool install --local Paket` (the `--local` i think is optional), is:
- pinned in the tool manifest
- can be run as `dotnet paket`
- `paket --help` -> `dotnet paket --help`, so will forward.

### Current status

works now as ☝️ in windows, printing the command who will run

### TODO

- [ ] fix the `TODO` comments
- [ ] use `PRocessStartInfo.ArgumentList` ? or continue to use Paket.Bootstrapper quoting code.
- [ ] message errors (if tool manifest exists but doesnt contains paket)
- [ ] add another init, to use tool manifest
- [ ] fix the sdk integration (should be fast, should run `dotnet paket` ALWAYS)

ref #3623

### DIFF
--- a/src/Paket.Bootstrapper/WindowsProcessArguments.cs
+++ b/src/Paket.Bootstrapper/WindowsProcessArguments.cs
@@ -3,7 +3,7 @@ using System.Text;
 
 namespace Paket.Bootstrapper
 {
-    static class WindowsProcessArguments
+    public static class WindowsProcessArguments
     {
         static void AddBackslashes(StringBuilder builder, int backslashes, bool beforeQuote)
         {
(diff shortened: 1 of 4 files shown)


enricosada:
@forki i'll finish it, but that's the idea

forki:
@enricosada currently master is broken: https://travis-ci.org/fsprojects/Paket/jobs/619580524#L1407

any ideas?

enricosada:
> @enricosada currently master is broken: https://travis-ci.org/fsprojects/Paket/jobs/619580524#L1407
> any ideas?

@forki i'll check, seems a bug in the packaging, i opened https://github.com/fsprojects/Paket/issues/3738

Issue #3763, opened 2019-12-21
NuGets with duplicate files get only partially extracted


viktor-svub:
### Description

NuGets with duplicate files get only partially extracted, resulting in Paket steps passing, but dotnet/msbuild restore failing as if they were not present. 

### Repro steps

1. Using paket.dependencies with: 
```
version 5.241.2 --prefer-nuget
storage: none
framework: netcoreapp3.1
source https://api.nuget.org/v3/index.json
nuget Microsoft.Extensions.DependencyInjection.Abstractions ~> 3.1
```

2. Run ```paket update --no-install && paket install && dotnet restore --force && dotnet build```
Paket steps seem to complete with suspicious log messages (will elaborate) but without explicit errors or warnings. DotNet restore & build fail as if the affected packages were missing completely, e.g. 
```My.csproj : error NU1100: Unable to resolve 'Microsoft.Extensions.DependencyInjection.Abstractions (>= 3.1.0)' for '.NETCoreApp,Version=v3.1'. [My.sln]```

### Expected behavior

Packages in global storage should be fully usable or not present at all. We should make best effort to make even badly authored packages usable, especially in scenarios where NuGet itself works. 

### Actual behavior

Packages with duplicate file entries are sometimes unpacked only partially, with some files and directories missing, without explicit errors or yellow warnings. 

interesting log fragments: ```
Paket version 5.241.2
[...update seems to run fine...]
Download of Microsoft.Extensions.DependencyInjection.Abstractions 3.1 done in 141 milliseconds. (3134 kbit/s, 0 MB)
[...other files are downloaded...]
Something went wrong while downloading Microsoft.Extensions.DependencyInjection.Abstractions 3.1
Message: The file '[user-home]\.nuget\packages\microsoft.extensions.dependencyinjection.abstractions\3.1.0\packageIcon.png' already exists.
  ==> Trying again
Downloading Microsoft.Extensions.DependencyInjection.Abstractions 3.1
[...other files are downloaded...]
Not moving as '[user-home]\.nuget\packages\microsoft.extensions.dependencyinjection.abstractions\3.1.0\microsoft.extensions.dependencyinjection.abstractions.3.1.0.nupkg' already exists (maybe some other instance downloaded it as well)
Download of Microsoft.Extensions.DependencyInjection.Abstractions 3.1 done in 562 milliseconds. (786 kbit/s, 0 MB) ```

resulting ```paket-installmodel.cache``` which mathes the actual unpacked directory: ```
[...]

PR #3764, opened 2019-12-21
bugfix/issue #3763 partially extracted nugets

viktor-svub:
fixing partial nuget extraction #3763 by extending fixes for borked and leaky zips to catch more cases


forki:
This is released in v6 alpha channel. Can you please test it out and then I'll push it to v5 as well. Thanks for your help!

viktor-svub:
Verified 6.0.0-alpha009 with previously failing dependencies -- the expected yellow warning "Trying to extract files individually" shows up, and the packages are fully extracted and work correctly in dotnet restore/build/run. Thank you!

viktor-svub:
It seems version `5.241.3` has some new merge/build issue -- `paket install` ends with:
```
Paket failed with
-> Resource stream 'Paket.Restore.targets' could not be found in the Paket.Core assembly
```
It happens on all install attempts, regardless of `.paket\Paket.Restore.targets` presence, and the file indeed does not re/appear with this version when not present. 

It does _not_ happen when 6.0.0-alpha009 is used for install, regardless if fresh or using 5.241.3 bootstrapper with version in paket.dependencies to use the alpha.

forki:
Damn. Unlisted that one.

Can you please try .4? If that also broken then I need to unlisted and
check that tomorrow.


Viktor Svub <notifications@github.com> schrieb am So., 22. Dez. 2019, 13:02:

forki:
Ok never mind. It's still broken. Seems I need to dig deeper tomorrow.

forki:
ok released 5.241.5 and that one seems to work. I had to install VS 2019 because it used a wrong compiler version. ffs

viktor-svub:
thank you! 5.241.5 works perfectly :)

PR #3783, opened 2020-01-29
FSharp.DependencyManager.Paket FSI extension for #r "paket: ..."

smoothdeveloper:
Bringing @forki's old friend.

I've tested with dotnet/fsharp and https://github.com/fsharp/FSharp.Compiler.Service/ master.

I'm looking for ideas/guidance about how to integrate the testing: I brought the fsharpqa tests we've had in the experimental branch but not worked yet on running tests in the solution.

The first contender is getting fsi.exe that understands this, I don't know how to pull it from dotnet/fsharp CI (@cartermp?), I can either build a version that will be used in the test project until it ships, or I can rely on FSharp.Compiler.Service to use in the test infrastructure that will be defined to run the checks.

@forki, do you know if there was things in the way it works that you weren't happy with but we didn't address back then?


baronfel:
As far as testing so concerned, when the next version of FCS releases you should be able to use that to that typechecking and compilation of scripts that would use #r paket

forki:
what's the plan here? how are we bootstrapping this thing? @cartermp suggest to use "dotnet tool install". would that be possible?

/cc @isaacabraham

smoothdeveloper:
@forki, I can pull an fsi.exe built from master and use that in the meantime if that sounds ok.

I haven't heard from the question of how to pull dotnet/fsharp master binary artifacts from CI.

If we want to test FCS as well, we can use the next release once it is built / published.

For the tests, I'll review those that were originally added to fsharpqa, I think there are positive and negative tests, and they may need a bit of different infrastrcture.

I'll go over recent change in master (https://github.com/dotnet/fsharp/pull/8415), and start at sketching the tests.

KevinRansom:
I hadn’t considered global tools, however that may well be a good approach.  Let me mull it over.

cartermp:
Bootstrapping via a local tool should probably work.

forki:
hey @smoothdeveloper  what's the status here?

smoothdeveloper:
@forki last time I was working on it, there was issue with inconsistent requirement between FSI shipped in VS and the one shipped in dotnet core sdk.

I'll check what is the status wrt this issue and bring the branch up to date.

forki:
Cool. Thanks a lot!

Gauthier Segay <notifications@github.com> schrieb am Sa., 25. Juli 2020,
09:09:

smoothdeveloper:
@forki this is ready for a round of review, two added tests are passing, involving `dotnet fsi` (running script and checking output) and `FCS` (type checking).

smoothdeveloper:
Note: I still need to revive the test set that was done in fsharp branch, it covers many advanced usecase relying on ambient `paket.dependencies` which are not exercised yet.

forki:
would really love to get it into the alpha

smoothdeveloper:
@forki, I'm almost there, there is only one of the original tests I need to have a custom unit test for due to output being non deterministic (full of paths and stack trace stuff).

I'm going to update FCS to most recent and fix that last test early this week, and the PR will be good for deep review.

There will still be some rough things in terms of the more elaborate integration (how `ResolveDependenciesResult` is used in the eco-system is still a bit opaque) but so far so good in FCS and FSI.

@KevinRansom could you explain in `ResolveDependenciesResult` the difference between `Roots` and `Resolutions` folder? I had to tweak that a bit and I'm currently passing the same "working area" folder to both, because it is needed to `#I` that folder, it seems `Roots` is used for that but I'm unsure what the other is for.

smoothdeveloper:
@forki, this is green, with all the original tests, plus a simple test using FCS, please let me know if there is anything we want adjusted.

forki:
I think what we need is a bit of documentation on how to use it. 
Currently I can't release since the certificate is outdated. But as soon as I get a replacement I will merge and release this

smoothdeveloper:
@forki there is now an initial take for the documentation.

Also, it seems to work in dotnet-interactive after copying the binary:

image

There are rough edges, it somehow keeps adding stuff to the deps file from dotnet-interactive, all of this is still as experimental as it was when it was PR to fsharp compiler repository...

forki:
this is released in alpha channel. hurray

smoothdeveloper:
@forki is it possible to have the dll part of the github release artifacts? I think that's what the people will download.

Otherwise they still need to compile it themselves.

The other thing to consider is to package this either in the paket or a new nuget package.

forki:
Sure. Could you please send a pr against build.fsx - that's where push to github releases

Issue #3788, opened 2020-02-06
Generated load scripts not working with dotnet fsi


eiriktsarpalis:
### Description

It seems that generated load scripts do not work well with `dotnet fsi`.

### Repro steps

1. Check out [this branch from Argu](https://github.com/fsprojects/Argu/tree/load-script-repro).
2. Run `dotnet tool restore && dotnet paket restore`.
3. Open `dotnet fsi`
4. Type `#load ".paket/load/netcoreapp3.1/DocGeneration/docgeneration.group.fsx" ;;`

### Expected behavior

Should load dependencies in context.

### Actual behavior

Complains about missing framework assemblies:
```
.paket\load\netcoreapp3.1\DocGeneration\docgeneration.group.fsx(42,1): error FS0078: Unable to find the file 'System.ComponentModel.Composition' in any of
 C:\Program Files\dotnet\shared\Microsoft.NETCore.App\3.1.1
 C:\Program Files\dotnet\sdk\3.1.101\FSharp\../../../packs/Microsoft.NETCore.App.Ref\3.1.0\ref\netcoreapp3.1
 C:\Users\eitsarpa\devel\public\Argu
 C:\Program Files\dotnet\sdk\3.1.101\FSharp\
```

### Known workarounds

Manually delete framework assemblies loaded from script.

baronfel:
System.ComponentModel.Composition isn't part of the sdk ref pack for 3.1.x, how does paket determine which assemblies to load?  FSI itself just auto-includes every .dll in the refs path (https://github.com/dotnet/fsharp/blob/master/src/fsharp/DotNetFrameworkDependencies.fs#L315) so paket shouldn't be getting in the business here of listing some platonic ideal list of refs.

matthid:
@baronfel I agree, but as I understand the issue it is the reverse, the script contains references the OP thinks it shouldn't contain.
I'd assume (without checking) that these are declared as framework dependencies in one of the packages. 

The question remains: How should paket decide what to generate and what not?

baronfel:
That's a good point, should paket probe the potential folders and cross-check the presence/absence of these framework refs then?  

Side-question: @eiriktsarpalis do you know which reference in particular is causing this framework reference to be added? looking at its Nuspec in more detail may shed some light on the discussion.

matthid:
>

inosik:
The filtering of framework assembly references doesn't seem to work correctly:

``` shell
[...]

PR #3793, opened 2020-02-10
Don't generate refs to full framework assemblies for netstandard scripts

inosik:
See #3788.


forki:
Ok no test was hit. Any chance you could add one?

inosik:
Do you have some test cases in mind?

PR #3804, opened 2020-03-01
Added type annotations to fix netstandard2.1 build

xavierzwirtz:
These type annotations were required for a netstandard2.1 build.


forki:
thanks

Issue #3690, opened 2019-10-19
package resolving blows stack


jdodds:
### Description

running `paket update` or other commands that need to resolve dependencies is blowing the stack

+ Paket 5.215.0, 5.226.0
+ macOS 10.13.6
+ dotnet 3.1.100-preview1-014459

### Repro steps

```sh
git clone https://github.com/jdodds/paket-blowing-stack-repro.git && cd paket-blowing-stack-repro
dotnet paket update
```

####  paket.dependencies:
```
source https://www.nuget.org/api/v2
source https://api.nuget.org/v3/index.json
storage: none

framework: netstandard2.0, netcoreapp3.0

nuget YoloDev.Expecto.TestSdk
nuget Microsoft.Extensions.Logging ~> 2.2
nuget Microsoft.AspNetCore.Identity.EntityFrameworkCore ~> 2.2
nuget Microsoft.AspNetCore.Proxy

```

### Expected behavior

Probably a failure message due to not being able to resolve packages

### Actual behavior

```sh
Process is terminating due to StackOverflowException.
Abort trap: 6
```

jdodds:
full output for `5.226.0`:

```shell
Paket version 5.226.0
Resolving packages for group Main:
 - Microsoft.Extensions.Logging 2.2.0
 - Microsoft.AspNetCore.Identity.EntityFrameworkCore 2.2.0
 - YoloDev.Expecto.TestSdk 0.8.0
 - Microsoft.AspNetCore.Proxy 0.2.0
 - Expecto 8.12.0
 - FSharp.Core 4.7.0
 - Microsoft.AspNetCore.Identity 2.2.0
 - Microsoft.EntityFrameworkCore.Relational 3.0.0
 - Microsoft.Extensions.Identity.Stores 3.0.0
 - Microsoft.Extensions.Identity.Stores 2.2.0
 - Microsoft.Extensions.Configuration.Binder 3.0.0
 - Microsoft.Extensions.DependencyInjection.Abstractions 3.0.0
 - Microsoft.Extensions.Logging.Abstractions 3.0.0
 - Microsoft.Extensions.Options 3.0.0
 - NETStandard.Library 2.0.3
 - System.Collections.Immutable 1.6.0
 - Microsoft.AspNetCore.Http.Abstractions 2.2.0
 - System.ComponentModel.Annotations 4.6.0
 - System.Text.Encodings.Web 4.6.0
 - Microsoft.EntityFrameworkCore 3.0.0
[...]

PR #3805, opened 2020-03-02
these lines checks seem to be triggering stack overflows

jdodds:
see #3690

Removing the `(x,y).Depth` parts of the `compare` calls, or swapping their order will cause paket to run without causing a StackOverflow on my repro. The behavior is the same whether or not the checks for `IsGlobalOverride` are separated.:

```fsharp
// this
let c = compare
    (not y.VersionRequirement.Range.IsGlobalOverride,y.Depth)
    (not x.VersionRequirement.Range.IsGlobalOverride,x.Depth)

// and this both "succeed"
let c = compare (x.VersionRequirement.Range.IsGlobalOverride, y.VersionRequirement.Range.IsGlobalOverride)
if c <> 0 then c else
let c = compare (y.Depth, x.Depth)

// this
let c = compare
     (not x.VersionRequirement.Range.IsGlobalOverride,x.Depth)
     (not y.VersionRequirement.Range.IsGlobalOverride,y.Depth)

// and this both "fail"
let c = compare (x.VersionRequirement.Range.IsGlobalOverride, y.VersionRequirement.Range.IsGlobalOverride)
if c<> 0 then c else
let c = compare (x.Depth, y.Depth)
```

Commenting the lines out as per https://github.com/fsprojects/Paket/commit/89d91444268535944dba2ea3fb95200388d586e8 also "succeeds". Seems like there might be something funky going on / a missed infinite recursion path around the logic for updating the requirements graph  of a `PackageRequirement`?

### DIFF
--- a/src/Paket.Core/Versioning/Requirements.fs
+++ b/src/Paket.Core/Versioning/Requirements.fs
@@ -1226,10 +1226,10 @@ type PackageRequirement =
 
     static member Compare(x,y,startWithPackage:PackageFilter option,boostX,boostY) =
         if obj.ReferenceEquals(x, y) then 0 else
-        let c = compare
-                  (not x.VersionRequirement.Range.IsGlobalOverride,x.Depth)
-                  (not y.VersionRequirement.Range.IsGlobalOverride,y.Depth)
-        if c <> 0 then c else
+//        let c = compare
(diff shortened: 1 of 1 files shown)


forki:
but that changes the order, right?

PR #3812, opened 2020-03-18
Allow to restore netcoreapp5.0   

### LINKED ISSUE #3811: Restore fails for netcoreapp5.0 projects
Paket does not like the .NET 5 preview released the other day, in particular it does not recognize netcoreapp5.0 references in paket.dependencies and fsproj files.

forki:
fixes #3811

/cc @kerams


src/Paket.Core/Versioning/FrameworkHandling.fs
```diff
@@ -650,6 +656,7 @@ type FrameworkIdentifier =
         | DotNetCoreApp DotNetCoreAppVersion.V2_2 -> [ DotNetCoreApp DotNetCoreAppVersion.V2_1 ]
         | DotNetCoreApp DotNetCoreAppVersion.V3_0 -> [ DotNetCoreApp DotNetCoreAppVersion.V2_2; DotNetStandard DotNetStandardVersion.V2_1 ]
         | DotNetCoreApp DotNetCoreAppVersion.V3_1 -> [ DotNetCoreApp DotNetCoreAppVersion.V3_0 ]
+        | DotNetCoreApp DotNetCoreAppVersion.V5_0 -> [ DotNetCoreApp DotNetCoreAppVersion.V3_1; DotNetFramework FrameworkVersion.V4_8 ]
```
sergey-tihon: Are you sure that `DotNetFramework FrameworkVersion.V4_8` should be here? (Just asking)

forki:
Isn't that what you sent me in the youtube link?

Sergey Tihon <notifications@github.com> schrieb am Mi., 18. März 2020,
08:56:

sergey-tihon:
I read this line as "There will be backward compatibility between DotNetCoreAppVersion.V5_0 and FrameworkVersion.V4_8" and that is probably not true

forki:
Yes the line reads as ".net 4.8 libs can be used in .net 5 apps". I thought
that's the whole idea.

Sergey Tihon <notifications@github.com> schrieb am Mi., 18. März 2020,
09:38:

khellang:
From [the design document](https://github.com/terrajobst/designs/blob/net5/accepted/2020/net5/net5.md):

> `net5.0`. This TFM is for code that runs everywhere. It combines and replaces the `netcoreapp` and `netstandard` names.

Also, under [design](https://github.com/terrajobst/designs/blob/net5/accepted/2020/net5/net5.md#design), it mentions that `net5.0` will be compatible with `net1..4` with a **NU1701** warning.

> `net4x` and earlier will continue to use `.NETFramework` as the TFI. This means that `net4x` and `net5.0` aren't considered compatible by default, but the compatibility will continue to be provided by .NET Framework compatibility mode we introduced in .NET Standard 2.0. It's handled via `AssetTargetFallback` in NuGet restore which also means consumers from `net5.0` will continue to get a proper warning.

> Please note that `net5.0`+ will map the TFI to `.NETCoreApp`. We need to announce this change so that package authors with custom .props and .targets are prepared.

> **Is .NET 5 a superset of .NET Framework 4.x?**
> No, .NET 5 isn't a superset of .NET Framework. However, .NET 5 is the successor of both .NET Core 3.x as well as .NET Framework 4.x.

khellang:
Basically, `netcoreapp5.0` is a temporary TFM that will go away (the code to change it to `net5.0` has already been merged and it's being discussed whether it should be backported to preview2 (from preview3)).

forki:
thx.

PR #3814, opened 2020-03-26
Add support for MonoAndroid10.0

TimLariviere:
Added support for the TFM `monoandroid10.0` for packages targeting Android 10.


forki:
this is targettting the master branch. are you using the prerelease versions? If not then we also need a PR against `bugfix` branch. Thanks

PR #3823, opened 2020-04-05
Update LiteGuard.Source.nuspec

adamralph:
FWIW 🤷‍♂


forki:
any specific reason why we need to update it? It's just old test data right?

adamralph:
@forki I've no idea. I just saw an old version of something and updated it to the current version.

This PR is yours to do with what you will. 🙂

forki:
Well it's test data. But Tests stay green so it's not important. Thx anyway

Issue #2883, opened 2017-11-02
Losing restrictions when 'dotnet pack'


MiloszKrajewski:
### Description

Yet again, it might not be paket's problem, but maybe there is something you can do about it.

Guys from 'NLog' team made a mess and rolled back version making making 4.5 newer than 5.0. 

My library, let's call it `libA` requires version 4.5 so `paket.dependencies` contains:

```nuget NLog ~> 4 beta```

and it is getting 4.5. So far, so good.

Now, my other project, let's call it appB, which uses `libA` does not know or care about NLog. The only dependency it has is `libA`:

```nuget libA```

But... when I do `paket install` it resolves `NLog` version `5.0`.

Apparently, `dotnet pack` for `libA` puts the restriction `>= 4.5` instead of `>= 4.5 && < 5`.

Long story short, my very precise restrictions about versions are lost when crossing `dotnet pack` boundary.

### Repro steps

[nlfu.zip](https://github.com/fsprojects/Paket/files/1438693/nlfu.zip)

Run `setup.cmd`. Notice `NLog 4` is referenced in `libA` (correctly) but `appB` references `NLog 5` (wrongly).

### Expected behavior

Both `libA` and `appB` reference version 4 of `NLog`.

### Actual behavior

Restrictions are "forgotten", and `appB` references "newest" version of "NLog"

### Known workarounds

Explicitly put restrictions in appB again (even if assembly is not used directly).

matthid:
We can make this work by modifying `fix-nuspec` which is called in the `dotnet pack` process.
We have to not only remove all transitives but also fix the range accoding to `paket.dependencies` (and ideally `paket.template`). Help is welcome.

matthid:
Related:
- https://github.com/fsprojects/Paket/issues/2678 - performance
- https://github.com/fsprojects/Paket/issues/2370 - Docs missing?
- https://github.com/fsprojects/Paket/issues/3011 - clarify the recommended variant
- https://github.com/fsprojects/Paket/issues/3047 - missing paket.template features

baronfel:
the code in question to be changed should be https://github.com/fsprojects/Paket/blob/1ab743a5a928146059842ad7f35cc8501769ff80/src/Paket.Core/PublicAPI.fs#L878-L889, from what I'm seeing

forki:
PRs always welcome

Chet Husk <notifications@github.com> schrieb am Fr., 24. Apr. 2020, 16:41:

baronfel:
[...]

PR #3835, opened 2020-04-24
Apply version ranges to nuspecs during fix-up command

baronfel:
Per #2883, version ranges specified in paket.dependencies weren't patched into auto-generated nuspecs as part of the `fix-nuspecs` call.  This PR attempts to add them in, in addition to a slight cleanup of the existing logic to make it easier to follow in my view.

The approach is to first determine what the nuspec file ranges should be for all known dependencies. This is done by taking the ranges specified in the dependencies file, which are optimistic, and modifying them with the 'resolved' information from the lock file. What this means depends on the type of range specified:

* for greater than, minimums, specific versions and overrides, this means changing the semver used for that bound to be the locked version. This seems safe for every case.
* for maximums, this means changing the range to a Between, because we want to ensure that consumers use _at least_ the version that this package was built against, while not going higher than this package's own build requirements
* similarly for less than, we want to ensure that the minimum bound is changed to the version that _this package_ was built against
* finally for Ranges, we change the lower bound to an inclusive lower bound on the locked version for similar reasons.

Once we have these ranges derived, we iterate through the dependencies in the nuspec as before, only taking action if we encounter a dependency that _we know_ is a direct dependency.

Caveats:
* this only tracks the main group of dependencies
  * we have the references file for this project, we could use that information to look for the range of the group that was actually referenced.
* this does not take into account targetframework restrictions.

I don't have tests for this yet, but from manual testing on F# Formatting (after adding an upper bound myself to the dependencies file and manually calling `fix-nuspecs` on the generated nuspecs), this seems to work:

```
nuget FSharp.Compiler.Service >= 35.0 < 36.0 lowest_matching:true
```
results in a nuspec with the range
```
<dependency id="FSharp.Compiler.Service" version="[35.0.0,36.0.0)" exclude="Build,Analyzers" />
```

### DIFF
--- a/src/Paket.Core/PublicAPI.fs
+++ b/src/Paket.Core/PublicAPI.fs
@@ -908,6 +908,23 @@ type Dependencies(dependenciesFileName: string) =
                 | None -> versionRequirement
             )
 
+        let (|IndirectDependency|DirectDependency|UnknownDependency|) (p: PackageName) =
+            if not (known.Contains p)
+            then UnknownDependency
+            else
+                if directDeps.Contains p
+                then DirectDependency
+                else IndirectDependency
+
+        let (|MatchesRange|NeedsRangeUpdate|LockRangeNotFound|) (packageName: PackageName, nuspecVersionRequirement: VersionRequirement) =
+
+            match lockedPackageVersionRequirements.TryFind packageName with
+            | Some lockedFileRange ->
+                if lockedFileRange = nuspecVersionRequirement
+                then MatchesRange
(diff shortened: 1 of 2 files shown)


baronfel:
There's one failure on appveyor that's in an old-framework test, I'm not sure it's relevant here

baronfel:
@forki do you have opinions/pointers on where/how I can test this?

forki:
is this something that should be released in 5.x? In that case can you please retarget it against `bugfix`?

baronfel:
Retargeted as requested 👍

PR #3864, opened 2020-06-11
Remove (Argu) Quotations.

ChrSteinert:
Argu removed the need for Quotations – I just stumbled across this and though it could not hurt :)
(If it does, by all means, just reject this :D )


forki:
Nice!

PR #3865, opened 2020-06-11
Changes the close behaviour of `paket find-packages`.

ChrSteinert:
`paket find-packages` without a search term gives a REPL experience to search for packages.
While I approve of the vim-style `:q` to quite I believe that pressing Ctrl+C is more CLI-like. 

This PR adds the graceful shutdown of Paket when pressing Ctrl+C (while removing the `;q` which could count as a breaking change…)


forki:
thanks!

PR #3868, opened 2020-06-15
Correct flaky tests in bootstrapper

JohnyWS:
I was trying to set up my local dev environment and ran into issues with flaky tests for the boostrapper, as well as having my dev environment overriding the code style in the repo.

Last thing first, was the simple thing: I added an `.editorconfig` file with the correct whitespace settings for cs files. I didn't dive deeper into how it should be for the rest of the repo, but now there is a place to add that config and have Visual Studio respect is (as well as a ton of other IDE's supporting it).

As for fixing the flaky test, it seems it broke with .NET Core 3.0 because readonly fields can no longer be overridden with reflection, so I had to refactor a bit, so it no longer relied on a static field.

It might not be the prettiest way of doing it (it is getting very close to needing an actual IoC Container instead), but for the current scope this seems to do the trick. I added an "IProxyProvider" that contains methods to aquire the different proxies, so it can be changed in the integration tests, removing the Singleton behavior entirely.

How do I check if the unit tests works here on Github though? I can't find any buildlogs with the Boostrapper tests in them? Locally it works just fine now though, but I'm having difficulties understanding how the build.fsx works. My current guess is though, that the unit tests for the boostrapper are only run locally, and not on appveyor.

NB: The DateTimeProxy should probably also be included in the new IProxyProvider, but it served no gain at the moment, and the current implementation works.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*.cs]
+indent_style = space
+indent_size = 4
--- a/Paket.sln
+++ b/Paket.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27009.1
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30204.135
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{A6A6AF7D-D6E3-442D-9B1E-58CC91879BE1}"
 EndProject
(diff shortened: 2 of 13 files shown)


forki:
@enricosada  could you please take a look? I think this code was mostly "yours"

JohnyWS:
Important note however: It will result in a new bootstrapper release, and perhaps that should wait to v6? There's nothing gained, a part from ensuring all tests can run.

Also, the question remains: Are the bootstrapper tests run in any of the build checks and/or release builds?

In the appveyor log from this PR, I found the following in the log:

```
[00:03:49] Creating C:\projects\paket\tests_result\netcore\Paket.Bootstrapper.Tests
[00:03:49] dotnet "test" "tests/Paket.Tests/Paket.Tests.fsproj" --configuration Release --framework netcoreapp3.0 --filter TestCategory!=Flaky --logger:trx;LogFileName=C:\projects\paket\tests_result\netcore\Paket.Bootstrapper.Tests\TestResult.trx --no-build -v n 
```

It seems that Paket.Tests.fsproj is run and the result logged to Paket.Bootstrapper.Tests\TestResult.trx - or am I misunderstanding the log output?

JohnyWS:
@forki any chance you have knowledge of that part of the buildsystem or knows who does? I'm primarily versed in Azure Pipelines, and not everthing I know from there translates that well to appveyor.. :)

forki:
thx

PR #3870, opened 2020-06-19
Aligned FAQ with getting started

### LINKED ISSUE #2902: Getting started and FAQ disagree on whether to commit .paket/paket.exe

JohnyWS:
I hope that I updated the FAQ correctly to properly reflect the two "supported" paths to using Paket.

I also included a warning that having committed `paket.bootstrapper.exe` is from using an approach that is no longer recommended.

Fixes #2902

### DIFF
--- a/docs/content/faq.md
+++ b/docs/content/faq.md
@@ -115,9 +115,11 @@ we will help to fix this in the affected packages.
 
 ## What files should I commit?
 
-Paket creates a number of files in your repository, and most of them should be
-committed to source control. To be clear, these are the files that should be
-committed to source control:
+There are two main ways to incorporate paket in your repository which are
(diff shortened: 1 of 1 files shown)


forki:
Thanks. Docs fixes are very appreciated

JohnyWS:
You're welcome @forki - I like to do them, but it requires in depth knowledge that I don't have yet, but gradully getting familiar with the whole projet by poking here and there.

PR #3871, opened 2020-06-22
Skip restore warning

### LINKED ISSUE #3869: PAKET_SKIP_RESTORE_TARGETS=true vs. false.
Hi,

this is related to issue #3863 .

For someone such as myself who is still learning Paket, I have a suggestion that when a "restore" is done, and `Paket.Restore.targets` is not created (because of `PAKET_SKIP_RESTORE_TARGETS=true`) that a STDOUT message (the opposite message of `Paket.Restore.targets` being created) be indicated -maybe something like:

`"Paket.Restore.targets not extracted to: %s (Can be enabled by deleting PAKET_SKIP_RESTORE_TARGETS or set to false)"`

----
I must have seen (when I did a paket restore, etc.):

`"Extracted Paket.Restore.targets to: %s (Can be disabled with PAKET_SKIP_RESTORE_TARGETS=true)"
`
-and I thought, hmmmmm.... why do I need this Paket.Restore.targets with .Net Core...? I don't recall seeing this with Paket/.Net "classic", so I will set `PAKET_SKIP_RESTORE_TARGETS=true`, and then I just forgot about it.

...well this caused a very subtle "issue", where `Paket.Restore.targets` was not created, and I could not figure out why, until I read through the Paket source code.

----
Also, in the Paket documentation I can find no official reference to `PAKET_SKIP_RESTORE_TARGETS` -it only popped up when google searches pointed to some old Paket issues.

Thanks,
George H.
06/15/2020 5:28pm ET

JohnyWS:
There a various scenarios where restoring Paket.Restore.targets are skipped by default, which is per design in the methods.

There is also an override however that is a bit more subtle, where skipping is done due to Environment var PAKET_SKIP_RESTORE_TARGETS being set to "true". This does not alter the output in any way.

This PR attempts to increase the verbosity of that action, by printing out when this happens.

Resolves #3869

### DIFF
--- a/src/Paket.Core/Installation/RestoreProcess.fs
+++ b/src/Paket.Core/Installation/RestoreProcess.fs
@@ -238,10 +238,12 @@ let extractRestoreTargets root =
         else
             let (fileWritten, path) = extractElement root "Paket.Restore.targets"
             copiedElements := true
-            if fileWritten then tracefn "Extracted Paket.Restore.targets to: %s (Can be disabled with PAKET_SKIP_RESTORE_TARGETS=true)" path
+            if fileWritten then
+                if verbose then tracefn "Extracted Paket.Restore.targets to: %s (Can be disabled with PAKET_SKIP_RESTORE_TARGETS=true)" path
+                else tracefn "Extracted Paket.Restore.targets to: %s" path
             path
     else 
-        tracefn "Skipping extraction of Paket.Restore.targets - if it was enabled, it would have been extracted to: %s (Can be re-enabled with PAKET_SKIP_RESTORE_TARGETS=false or deleting the environment variable to revert to default behavior)" path
+        if verbose then tracefn "Skipping extraction of Paket.Restore.targets - if it was enabled, it would have been extracted to: %s (Can be re-enabled with PAKET_SKIP_RESTORE_TARGETS=false or deleting the environment variable to revert to default behavior)" path
         path
 
 let CreateInstallModel(alternativeProjectRoot, root, groupName, sources, caches, force, package) =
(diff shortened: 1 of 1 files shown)


JohnyWS:
I would have loved to add more documentation and testing to it, but ran into a few issues:

Docs: I could not find a section where it fits - restore and install section does not include anything on restoring `Paket.Restore.targets`, and I do not have enough knowledge to describe the default scenarios, which I belive is important to have in the docs before adding this rare edge case.

Tests: I do not have a functioning local test environment yet, and I could not find any tests on neither `RestoreProcess.fs` nor `InstallProcess.fs` that I hoped I could use as starting point for adding the tests.

I hope the creator of #3869 will be able to verify if it works as intended or not.

forki:
I think we should make it `if verbose then` - right?

JohnyWS:
Hmm, shouldn't it rather be the other way around. Using the flag is rare,
but very hard to notice if you do?

So I'd propose adding the verbose flag requirement to the existing message
instead?

ons. 24. jun. 2020 07.46 skrev Steffen Forkmann <notifications@github.com>:

matthid:
Another approach to solve this newcomer issue would be to remove the tip in the braces and only add that in verbose mode.

JohnyWS:
Hmm, what about this then? :) Then you only get the instruction if you use verbose, and the new info is only used in verbose. Then you'll only lack instructionst if you both run with verbose first time, enable the environment variable, and then in consecutive runs run it without verbose.

JohnyWS:
@matthid @forki ? :)

forki:
thanks

Issue #3723, opened 2019-11-14
paket pack and developmentDependency: Fail on Linux (works on Windows)


jbaehr:
### Description

tl;dr; `paket pack` handles development depencencies correctly (aka *not* placing it in the resulting dependency list) on Windows, but fails so on Linux (i.e. *does* add dependencies on development dependencies).

I have a library targeting netstandard20, developed using .NET Core SDK 2.2.
I'm using a `paket.template`, type project, to create the nupkg via `paket pack`. Using paket v5.219.0, installed as a local dotnet tool.

The library depends on some other nugets, two of them being "development dependencies" (marked with `<developmentDependency>true</developmentDependency>` in their nuspec).

Running `paket pack` on a Windows 10 workstation results in a nupkg that does not have those development dependency nugets listed in its nuspec, as expected.
Running the very same command, using the very same version of paket, on a Linux CI server, running Microsoft's official SDK-2.2 docker image, paket generates a nuspec that *does* have those dependencies listed.

### Repro steps
(paket bootstrapping as in https://github.com/fsprojects/Paket/issues/3623#issuecomment-541098823)

[paket-issue-3723.zip](https://github.com/fsprojects/Paket/files/3859054/paket-issue-3723.zip)

On Windows:
```
git clone ....
dotnet restore .paket
dotnet build
.paket\paket pack
```
results in a correct nuget.

On Linux
```
git clone ....
dotnet restore .paket
dotnet build
.paket/paket pack
```
results in a nuget falsely depending on the development dependencies, too.
(And having only `\n` as line separator, opposed to the `\r\n` in the windows-created nuget, but that's the only difference when diffing the two packages and doesn't matter at all)

### Expected behavior

Running `paket pack` on Linux should honor the development dependencies correctly, just like it's doing [already](https://github.com/fsprojects/Paket/issues/796) on Windows.

### Actual behavior

On Linux, `paket pack` adds wrongly dependencies to development-only dependencies.

### Known workarounds

Use `excludeddependencies` in the templates and list all development dependencies explicitly. -- This defeats the whole purpose of having `developmentDependency`, though.

I haven't checked `dotnet pack` on Linux, as on Windows it does *not* handle development dependencies (see #2370). In addition, `dotnet pack` reads the package version from the `PackageVersion` msbuild property while `paket pack` (with type project) infers it from the `AssemblyInformationalVersion` attribute, which fits better for us.

forki:
can you please upload a small zip which I can use as test case?

[...]

PR #3873, opened 2020-06-26
Do not add development dependencies to the nuspec of a target nupkg while paket packing

ruhullahshah:
### Motivation:
It was observed on some systems that `paket pack` adds development dependencies to the nuspec of a target nuget package.
The root cause for this has been documented [here](https://github.com/fsprojects/Paket/issues/3723#issuecomment-650228275). This PR aims to fix issue #3723

### Description of fix:
If a nupkg does not exist in the local nuget cache, try to load the nuspec from the `DefaultPackagesFolder` before returning a `Nuspec.All`


forki:
thanks a lot for taking care of this

ruhullahshah:
You are welcome

Issue #3874, opened 2020-07-01
Paket doesn't recognise UAP version 10.0.10240


BillHally:
### Description

Paket doesn't recognise UAP version 10.0.10240, and so reports a warning or error when installing a Nuget package which references it.

### Repro steps

1. Add a dependency to a Nuget package with a dependency on UAP 10.0.10240 (e.g. SkiaSharp)

2. Run `paket install`

There's a tiny repo at https://github.com/BillHally/Paket.Issue.UnsupportedUapVersion which demonstrates the issue when you run `paket install`.

### Expected behavior

No warning or error is reported.

### Actual behavior

The following warning or error is reported:

```
Could not detect any platforms from 'UAP10.0.10240' in SkiaSharp 1.68.3, please tell the package authors
```

Various related warnings are also reported near the end of the run.

If your paket.dependencies specifies v2 source for nuget.org, then the message is a warning.

If however you specify a v3 source for nuget.org, then the message is an error.

### Known workarounds

Run `paket install` for a 2nd time will avoid the error reported when using a v3 source for nuget.org, because the error is only reported when the package is first installed.

However, some similar warnings are always reported.

### Notes

I'll submit a PR to fix it.

BillHally:
Fixed in 5.257.0. Thanks @forki!

PR #3875, opened 2020-07-01
Add support for UAP v10.0.10240

BillHally:
This is a fix for #3874.

I *think* it's right, but I've not used it to build anything which uses the UAP version in question (or any other UAP version for that matter). I have however used it to build a project which uses a Nuget package which declares a dependency for this UAP version, and it prevents it from reporting incorrect errors and warnings (i.e. it fixes the bug described by the issue mentioned above).

### DIFF
--- a/src/Paket.Core/Dependencies/NuGetCache.fs
+++ b/src/Paket.Core/Dependencies/NuGetCache.fs
@@ -123,7 +123,7 @@ type NuGetPackageCache =
       Version: string
       CacheVersion: string }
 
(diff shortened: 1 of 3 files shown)


forki:
can you please take a look at the red build? thanks

BillHally:
It timed out - but I guess you worked that out given that you've closed and re-opened the PR :-)

forki:
thanks. this is released in alpha and stable channel

Issue #3768, opened 2020-01-02
add-credentials fails


haf:
### Description

```
$ .paket/paket config add-credentials https://nuget.pkg.github.com/haf/index.json
Paket version 5.241.5
Username: haf
Password: ****************************************
Authentication type (basic|ntlm, default = basic): basic
Performance:
 - Runtime: 3 seconds
Paket failed with
-> PlatformNotSupportedException: Windows Data Protection API (DPAPI) is not supported on this platform.
```

### Repro steps

See above

### Expected behavior

Not use NTLM code

### Actual behavior

Uses NTLM code

### Known workarounds

None

peterhirn:
Since the DPAPI is never going to work on anything but windows, I think paket should add support for cleartext passwords, eg.:

```
paket config --cleartext add-credentials https://pkgs.dev.azure.com/myorg/_packaging/myfeed/nuget/v3/index.json
paket config --cleartext add-credentials https://pkgs.dev.azure.com/myorg/_packaging/myfeed/nuget/v2
```

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <credentials>
    <credential source="https://pkgs.dev.azure.com/myorg/_packaging/myfeed/nuget/v3/index.json" username="PAT" cleartext="personal-access-token" authType="basic" />
    <credential source="https://pkgs.dev.azure.com/myorg/_packaging/myfeed/nuget/v2" username="PAT" cleartext="personal-access-token" authType="basic" />
  </credentials>
</configuration>
```

This would enable pushing packages from CI.

vilinski:
- it's a duplicate of #3462
- sad that it doesn't work on mac and linux
- the only workaround was for me to use credentials in the environment variables
  - documented here: http://fsprojects.github.io/Paket/nuget-dependencies.html#NuGet-feeds
  - example: `source https://fiz-baz.com/artifactory/api/nuget/our-corp-nuget username: "%REPOSITORY_USER%" password: "%REPOSITORY_USER_PASS%"`

bengobeil:
Hi, I will be trying to fix this.

bengobeil:
Would it be acceptable to use AES instead of DPAPI for password encryption? @forki

forki:
Guess so.

Benjamin Gobeil <notifications@github.com> schrieb am Di., 21. Juli 2020,
17:31:

bradcypert:
This looks like its resolved right? Can this issue be closed?

PR #3882, opened 2020-07-22
Use AES encryption when not on windows

bengobeil:
When not on windows, instead of using Data Protection API, use AES encryption.

### DIFF
--- a/src/Paket.Core/Common/Encryption.fs
+++ b/src/Paket.Core/Common/Encryption.fs
@@ -1,0 +1,161 @@
+﻿namespace Paket.Core.Common
+
+open System
+open System.IO
+open System.Net.Http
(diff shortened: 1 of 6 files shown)


bengobeil:
Fixes issue #3768

forki:
Is this backwards compatible? I assume it is, but are windows users forced to reenter their passwords?

bengobeil:
Its backwards compatible. If its on windows we will use DPApi. Else, use AES. That means if it worked before on windows, you will not have to re-enter

forki:
do you think this should go to paket 5.x as well? then pleae feel free to send a similar PR that targets  `bugfix` branch.
Thanks!

bengobeil:
OK will do

PR #3884, opened 2020-07-23
Use AES encryption when not on windows

bengobeil:
(cherry picked from commit b22361ce510ab530c8f8886f012d617893e61a8d)


bengobeil:
If we can release this today, it will be very helpful for us. Thanks!

forki:
image

bengobeil:
What are the repro steps?

bengobeil:
Oh its just doesn't build... one sec

bengobeil:
Oh, probably a different lang version on 5.XXX ?

forki:
yes it's old school ;-)

bengobeil:
OK, it was a case of implicit yields

forki:
Cool. Will release as soon as I'm back home

bengobeil:
Great thanks!

PR #3886, opened 2020-07-24
[Improvement of fix introduced in PR #3873] Search all the sub-directories under the default packages directory

ruhullahshah:
## Problem:
While paket packing if a dependency is not available in the local user nuget cache, before returning the default Nuspec record, we used to seek the package in the default packages directory, but this method is not fool proof as the dependency might be located within a sub-directory of the default pakcages directory, for instance `packages\build`

## Solution:
Perform an exhaustive search under the default packages directory instead of looking at the top level only.

### DIFF
--- a/src/Paket.Core/Dependencies/CacheExtensions.fs
+++ b/src/Paket.Core/Dependencies/CacheExtensions.fs
@@ -10,7 +10,7 @@ module CacheExtensions =
             let nuspec = Path.Combine(folder,sprintf "%O.nuspec" name)
             let fi = FileInfo(nuspec)
             // It happens in some cases that the package is not available in the cache, and in that case
-            // it would be sensible to recusrively search for the nuspec in the default package folder before returning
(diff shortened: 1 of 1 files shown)


forki:
is that something that should go to paket 5? then please target `bugfix` branch

ruhullahshah:
@forki Yes the fix is intended for Paket 5, but can be merged to master as well since the fix is applicable to Paket 6 also. Do I create a separate PR with the commits cherry picked to `bugfix`?

forki:
if you retarget to bugfix then I'll release it for both

ruhullahshah:
@forki Alright, I will try to fix this issue on Linux as well and create a new PR targeting the `bugfix`

forki:
cool thank you!

ruhullahshah:
@forki I have added a new PR https://github.com/fsprojects/Paket/pull/3888 that targets `bugfix` and fixes the issue on Linux as well. I will close this one

Issue #3816, opened 2020-03-26
Could not detect any platforms from 'XCODE10' / 'XCODE11'


TimLariviere:
### Description

Starting v4.5.0, [Xamarin.Forms](https://www.nuget.org/packages/Xamarin.Forms/4.5.0.495) added 2 new folders in their NuGet packages, under the folder `build`, named `XCODE10` and `XCODE11`.

When installing this package with Paket, the following warnings appear:
```
Xamarin.Forms 4.5.0.495 unzipped to /Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms
Could not detect any platforms from 'XCODE10' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE10/Xamarin.Forms.Platform.iOS.dll', please tell the package authors
Could not detect any platforms from 'XCODE10' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE10/Xamarin.Forms.Platform.macOS.dll', please tell the package authors
Could not detect any platforms from 'XCODE10' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE10/Xamarin.Forms.Platform.iOS.pdb', please tell the package authors
Could not detect any platforms from 'XCODE11' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE11/Xamarin.Forms.Platform.iOS.dll', please tell the package authors
Could not detect any platforms from 'XCODE11' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE11/Xamarin.Forms.Platform.macOS.dll', please tell the package authors
Could not detect any platforms from 'XCODE11' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE11/Xamarin.Forms.Platform.iOS.pdb', please tell the package authors
Could not detect any platforms from 'XCODE10' in '/Users/timothelariviere/Git/GitHub/Fabulous/packages/androidapp/Xamarin.Forms/build/XCODE10/Xamarin.Forms.Platform.iOS.dll', please tell the package authors
[...]

PR #3887, opened 2020-07-24
Add props from root build folders

bengobeil:
#3816 

Props and target files were not added to project files if they were found in the root of build, buildTransitive or buildMultiTargeting folders.

### DIFF
--- a/src/Paket.Core/PaketConfigFiles/InstallModel.fs
+++ b/src/Paket.Core/PaketConfigFiles/InstallModel.fs
@@ -54,13 +54,16 @@ type MsBuildFile = {
 }
 
 module MsBuildFile =
-    let ofFile (f:FrameworkDependentFile) =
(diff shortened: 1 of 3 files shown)


forki:
is that something that should go to paket 5? then please target `bugfix` branch

bengobeil:
Yes I will add it to paket 5. ALl tests pass on my machine but some fail on AppVeyor. Not sure how to diagnose

forki:
let's try to rerun

bengobeil:
Actually I was not running the RunIntegrationTestDotnetCore target. Trying it on my machine

bengobeil:
Yeah it also fails on my PC

bengobeil:
Should be fixed now. So Ishould merge this in paket 6 and paket 5?

forki:
thank you

PR #3891, opened 2020-07-27
Do not include buildTransitve/buildMultiTargeting by default

bengobeil:
Causes issues when building projects if multiple copies of same target file get imported to project


bengobeil:
#3816

bengobeil:
@forki would it be possible to merge this? and I have another one for 6.0.0

forki:
it's released in 5.x and 6.x. thx

bengobeil:
there was a slight f*** up. This should have been targeting bugfix. it was targeting master

bengobeil:
I will change my other PR

Issue #3897, opened 2020-08-12
SDK 3.1.401 blows up on "dotnet pack" if project uses paket


MiloszKrajewski:
### Description

As usual - new dotnet core release (3.1.401) new problems.
After installing new SDK (3.1.401 released 2020/08/11) previously working project blows up on "dotnet pack"

```
C:\dev\k4os\pkdn31401> "C:\Program Files\dotnet\dotnet.EXE"  pack "C:\dev\k4os\pkdn31401\src\pkdn31401a" --configuration Release --output "C:\dev\k4os\pkdn31401\.output" --no-build --no-restore /nodeReuse:False "/bl:C:\Users\...\AppData\Local\Temp\tmp4C4B.tmp.binlog" (In: false, Out: false, Err: false)
Microsoft (R) Build Engine version 16.7.0-preview-20360-03+188921e2f for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

C:\Program Files\dotnet\sdk\3.1.401\MSBuild.dll -distributedlogger:Microsoft.DotNet.Tools.MSBuild.MSBuildLogger,C:\Program Files\dotnet\sdk\3.1.401\dotnet.dll*Microsoft.DotNet.Tools.MSBuild.MSBuildForwardingLogger,C:\Program Files\dotnet\sdk\3.1.401\dotnet.dll -maxcpucount -property:Configuration=Release -property:PackageOutputPath=C:\dev\k4os\pkdn31401\.output -property:NoBuild=true -target:pack -verbosity:m /bl:C:\Users\...\AppData\Local\Temp\tmp4C4B.tmp.binlog /nodeReuse:False C:\dev\k4os\pkdn31401\src\pkdn31401a\pkdn31401a.csproj
C:\Program Files\dotnet\sdk\3.1.401\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(198,5): error MSB4181: The "PackTask" task returned false but did not log an error. [C:\dev\k4os\pkdn31401\src\pkdn31401a\pkdn31401a.csproj]
: C:\Program Files\dotnet\sdk\3.1.401\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting\NuGet.Build.Tasks.Pack.targets(198,5): error MSB4181: The "PackTask" task returned false but did not log an error.
```

then actual error being: `error MSB4181: The "PackTask" task returned false but did not log an error.`

### Repro steps

[...]

PR #3898, opened 2020-08-18
Add rollforward to globaljson

cartermp:
Pinning to a specific version can lead to unexpected issues, like Visual Studio removing the SDK it installed and replacing it with a new one, causing all projects to fail to load.

This will ensure that you can continue to update your tools and (usually) not get affected.

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,5 +1,6 @@
 {
     "sdk": {
-      "version": "3.1.302"
+      "version": "3.1.302",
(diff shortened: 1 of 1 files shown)


forki:
Thx

forki:
Unfortunately we need to fix https://github.com/fsprojects/Paket/issues/3897 before we can use 3.1.401

PR #3899, opened 2020-08-19
netcoreapp5.0 -> net5.0

thinkbeforecoding:
the name for the next version of .Net is net5.0

It is now part of FrameworkVersion (not DotNetCoreAppVersion).

I still defined RawSupportedPlatforms for net5.0 as being [ netcoreapp3.1; netstandard2.1 ]


forki:
wow. NICE!

thinkbeforecoding:
this is an opportunistic contribution 😄 
I have projects that are using paket and I want to check I can multitarget them to net5.0 before it's released

PR #3908, opened 2020-09-14
[Don't Merge Yet] Fix #3906: missing url escape of square backets

### LINKED ISSUE #3906: Paket pack doesn't escape square brackets
### Description

Nuget packages (*.nupkg) conform to the "Open Packaging Concentions", OPC. Thus files within a package are URL-adressed. This also means that reserved chars need to be percent-encoded. For most chars `paket pack` does the encoding correctly (e.g. a space becomes `%20`). Square brackets, however, get through unescaped (i.e. `[]` *does not* become `%5B%5D`).

This becomes an issue when processing a nuget package with .NET's OPC API (`System.IO.Packaging`). If a part contains a square bracket an `ArgumentException` is thrown, stating that the URI is not valid. And indeed, according to [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt) (cf. very end of Appendix A) square brackets are reserved within a path segment and thus need to be percent-encoded.

### Repro steps

1. Create a paket.template that harvests a directory
2. ensure there is a file within the harvested dir that contains square brackets (e.g. `file[0].txt`)
3. `paket pack` that template

### Expected behavior

The file *within* the resulting nuget should look like this: `file%5B0%5D.txt`.

### Actual behavior

The file actually looks like this: `file[0].txt`.

### Known workarounds

None I'm aware of.

jbaehr:
As System.Uri.GetComponents(...*UriFormat.SafeUnescaped*) is based on
the outdated RFC 2396 it unescapes angle brackets, which usage is only
declared "unwise" in the original URI RFC from 1998. The uptodate one,
RFC 3986, forbidds them in path segments. This causes trouble with
software based on the newer RFC, such as the OPC API in
`System.IO.Packaing`.

Instead of hooking a third layer of escape/unescape layer on top of the
existing code, I decided to rewrite the `ensureValidName` method, hoping
the intend of the code is clearer now. For more details see the comments
in #3906.

***Note***: I was only able to verify the method isolated in LinqPad. I tried hard but failed to compile paket (and thus run the tests) because of a corporate proxy requiring credentials. Sorry for that; let's hope for the CI.

### DIFF
--- a/src/Paket.Core/Packaging/NupkgWriter.fs
+++ b/src/Paket.Core/Packaging/NupkgWriter.fs
@@ -293,43 +293,42 @@ module internal NupkgWriter =
             normalizePath path = outputFolder || (exclusions |> List.exists (fun f -> f path))
 
         let ensureValidName (target: string) =
-            // Some characters that are considered reserved by RFC 2396
-            // and thus escaped by Uri.EscapeDataString, are valid in folder names.
-            // Concrete problem solved here:
+            // Some characters that are considered reserved by (obsolete) RFC 2396
+            // and thus escaped by Uri.EscapeDataString, are valid in folder names
+            // according to the current RFC 3986.
+            // In nuget packages this over-aggressive escaping does not hurt when
+            // unpacking using a nuget client. However, it makes the raw package content
+            // harder to read for humans and may confuse other tools.
+            // Concrete problem solved here (cf. #1348):
             // Creating deployable packages for javascript applications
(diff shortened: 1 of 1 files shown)


jbaehr:
Currently, it's not clear if this proposed fix here is for better or worse, cf. https://github.com/fsprojects/Paket/issues/3906#issuecomment-696666734.
So for now, I suggest to *not* merge.

PR #3916, opened 2020-10-10
Allow repository branch and commit

wastedsxeyouth:
Small PR (building on i003707) to allow RepositoryBranch and RepositoryCommit to be specified as per https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets

Also, when setting up from a clean Windows box the bin folder was not created in the root of the project. When I manually ran the publish task it was created, so I have included a lone commit add the publish step into the "all" target.


build.fsx
```diff
@@ -691,6 +695,7 @@ Target "All" DoNothing
   ==> "Restore"
   ==> "AssemblyInfo"
   ==> "Build"
+  ==> "Publish"
```
forki, line 698: I'm not sure about this one!?

forki (changes requested):
Can we also please have a bit in the docs

forki:
thanks!

PR #3919, opened 2020-10-22
tweaks to build script to publish FSharp.DependencyManager.Paket

smoothdeveloper:
Not 100% sure it is working, I expect those things:
* FSharp.DependencyManager.Paket.dll is part of github release artifacts
* FSharp.DependencyManager.Paket.{version}.nupkg is part of github release artifacts
* FSharp.DependencyManager.Paket is a nuget package published on nuget.org


forki:
I had to adjust the path, but I think it worked

PR #3920, opened 2020-10-22
Adds Testimonial of RheinSpree

brase:
It's not a real bug. More a lack of content ;)

Here is our short Rhein-Spree testimonial. We hope it gets approved and merged soon.


forki:
Thanks so much!

PR #3922, opened 2020-10-24
attempt to fix pathToExtension in FsiExtension tests.

smoothdeveloper:


smoothdeveloper:
I'm actually confused with `<OutputPath>` not being honored on a call to `build.cmd`.

forki:
yeah I'm confused as hell

smoothdeveloper:
Ok, there is a problem with the output path being inconsistent between build from IDE and build script, the other issue I suspect is that we need a binding redirect for another assembly than just FSharp.Core in the FCS test.

For the first issue, something like the `pathToExtension` hack in this branch may be good work around for now.

For the second, this is confusing error because the loading mechanism will silently move on from an extension assembly that can't be loaded (due to binding redirect issue), this is the reason https://github.com/fsprojects/Paket/blob/fc90013d093211018b8ab8ba09943f563847b37f/integrationtests/Paket.IntegrationTests/FsiExtension.fs#L21-L29 was put in place.

Does FSharp.Core 5 depends on additional thing?

luchitor (approved):

Issue #3933, opened 2020-11-10
net5.0-windows platform doesn't work


michalkovy:
### Description
When I target net5.0-windows platform and use latest preview paket I get this error:

Paket version 6.0.0-alpha051
...
Paket failed with
-> Extracting platforms from path 'net5.0-windows' failed, a newer Paket version might support it, otherwise please open an issue for a missing platform handler.

.NET 5.0 has new OS specific target frameworks which needs to be used when OS specific functionality is required, see [documentation](https://docs.microsoft.com/en-us/dotnet/standard/frameworks). E.g. one needs to use net5.0-windows to use WPF or WinForms.

OS version can be also specified (and is required to get new APIs), these target platforms are valid, too:
net5.0-windows10.0.19041.0
net5.0-windows10.0.18362.0
net5.0-windows10.0.17763.0
net5.0-windows8.0
net5.0-windows7.0

I haven't tested other OS specific platforms.

### Repro steps

Create project that targets framework net5.0-windows.

### Expected behavior

All net5.0 OS specific frameworks are supported

### Actual behavior

Paket fails

### Known workarounds

None to me

forki:
PR highly wanted

Michal Kováč <notifications@github.com> schrieb am Di., 10. Nov. 2020,
21:01:

michalkovy:
I have added to the description info about windows versions for net5.0-windows.

forki:
@michalkovy please test latest version and let us know.

michalkovy:
my change doesn't work completely, I am working on a fix

PR #3934, opened 2020-11-11
Net5 os dependent frameworks

mk185147:
.NET 5.0 comes with new target frameworks that are OS specific.

This is fix of #3933


forki:
this looks good!

forki:
this is released in alpha and stable channel. thanks a lot

Issue #3939, opened 2020-11-13
.NET 5 projects won't compile in Visual Studio when framework restriction is used


mk185147:
### Description

When you set in paket.dependencies restriction to net5.0 or net5.0-windows or similar then you will be able to compile with dotnet build but when you open such a solution in Visual Studio it won't load nuget packages for such .NET 5 based projects

### Repro steps

Use Paket 6.0.0-alpha054

1. Have a solution with projects that target netstandard2.0, netstandard2.1, net5.0 and net5.0-windows

2. Create paket.dependencies and add line:
framework: netstandard2.0, netstandard2.1, net5.0, net5.0-windows

3. Ensure the project use somethig from referenced nuget packages

4. Open the solution in Visual Studio

### Expected behavior

Visual Studio is able to work when paket.dependencies contains restriction to .NET 5 based projects

### Actual behavior

You will see that netstandard project have nuget packages loaded but net5.0 and net5.0-windows won't and if you try to compile them they will fail on the missing references. If you try to compile the same solution with dotnet build it will work.

thinkbeforecoding:
It was working with the VS 16.8 last preview version.... what did they break ?!

thinkbeforecoding:
Actually it seems to be working with v5.253.0
But not with 5.254.0 and 5.255.0

PR #3943, opened 2020-11-16
Fix how msbuild condition is created based on target frameworks

mk185147:
this is fix of #3939

The problem was that when you used in proj file e.g. target framework net5.0-windows it was not matching msbuild condition '$(TargetFramework)' == 'net5.0-windows7.0' which was created for the net5.0-windows target framework.

Generally, there are many ways how to write the same TargetFramework. Hence, I switched to the form like "($(TargetFrameworkIdentifier) == '.NETCoreApp' And ($(TargetFrameworkVersion) == 'v5.0' And '$(TargetPlatformIdentifier)' == 'Windows'))" like it was done on other place in Paket


forki:
the travis is red. unfortunately

mk185147:
@forki  I checked it and it seems like some infrastructure problem:
EXEC : The remote server returned an error : (429) too many requests. (Github - cached) [/home/travis/build/fsprojects/Paket/src/Paket/Paket.fsproj]

forki:
ok let's rerun this

PR #3946, opened 2020-11-16
Update ReferenceLoading.PaketHandler.fs

smoothdeveloper:
The framework moniker `"netcoreapp5.0"` should resolve to  `"net50"` instead of `"net5.0"` AFAICS, otherwise the path to load script is incorrect.


smoothdeveloper:
@forki this one is good, not sure what happened between preview and release version of dotnet 5 but this works with the release version.

PR #3955, opened 2020-11-23
Pick default project

isaacabraham:
@forki this is built on top of the **add-without-init** branch. It's totally the wrong way to do it, but if there's a way to identify any project in the "current" folder, then I think this approach would work. At least, testing locally it seems to do the right thing: you can go to an empty project, type `paket add Saturn` (assuming you have paket as a global tool) and everything will just work.

### DIFF
--- a/src/Paket.Core/PackageManagement/AddProcess.fs
+++ b/src/Paket.Core/PackageManagement/AddProcess.fs
@@ -25,11 +25,11 @@ let private add installToProjects addToProjectsF dependenciesFileName groupName 
         traceWarnfn "%s contains package %O in group %O already." dependenciesFileName package groupName
     else
         let lockFileName = DependenciesFile.FindLockfile dependenciesFileName
-        let lockFile = ref None
+        let mutable lockFile = None
         let lockFileHasPackage =
             if not lockFileName.Exists then false else
             let lf = LockFile.LoadFrom lockFileName.FullName
-            lockFile := Some lf
+            lockFile <- Some lf
             let lockFileGroup = lf.GetGroup(groupName)
             let vr = DependenciesFileParser.parseVersionString version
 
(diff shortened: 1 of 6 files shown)


isaacabraham:
@forki I think that this is good to go now. Behaviour is as follows on a completely clean project (no paket)

```cmd
dotnet Paket add Saturn
Paket version 6.0.0-beta2
Adding package 'Saturn'
Resolving dependency graph...
Created dependency graph (16 packages in total)
Extracted Paket.Restore.targets to: C:\Users\Isaac\Source\Repos\Samples\projectb\.paket\Paket.Restore.targets
Resolved package 'Saturn' to version 0.14.1
Total time taken: 5 seconds
```

I'm just going to move the Paket.Restore.Targets line to verbose as well, but otherwise I think that this is good to test.

src/Paket.Core/PackageManagement/AddProcess.fs
```diff
@@ -25,11 +25,11 @@ let private add installToProjects addToProjectsF dependenciesFileName groupName
         traceWarnfn "%s contains package %O in group %O already." dependenciesFileName package groupName
     else
         let lockFileName = DependenciesFile.FindLockfile dependenciesFileName
-        let lockFile = ref None
+        let mutable lockFile = None
```
forki, line 28: REF CELLS!!!

forki:
the appveyor is red

forki:
still red

isaacabraham:
Mmm, need to look into that.

isaacabraham:
Works on Travis but not AV.

isaacabraham:
@forki does paket even need to support dotnet tools any more?

isaacabraham:
Interestingly, just running the failing test command `paket add dotnet-fable --version 1.3.7 -t clitool --no-resolve` (from current live paket) also doesn't generate a lock file.

isaacabraham:
Ok I think I know what the problem is!

isaacabraham:
Green.

Issue #3950, opened 2020-11-21
Dependency manager


MecuSorin:
### Description

Attempted to play with the new features in FSharpInteractive using the FSharp.DependencyManager.Paket, but encountered some issues. Was thinking that we should provide an easier experience for newcomers.

### Repro steps
In a new folder:
```
dotnet new tool-manifest
dotnet tool install paket
dotnet paket init
dotnet paket add FSharp.DependencyManager.Paket
```
Changed the settings.json for VSCode to include
{    
    "FSharp.fsiExtraParameters": [
        "--compilertool:c:/Users/Sorin/.nuget/packages/fsharp.dependencymanager.paket/6.0.0-alpha053/lib/netstandard2.0"
    ]
}
Then created ```something.fsx``` with the content:
```
#r "paket: nuget FParsec"
open FParsec
```

### Expected behavior

The script to run in FSI without issues

### Actual behavior

exception while resolving dependencies: 
```
System.Exception: Paket was not found in 'd:\dev\fsharp\sandbox\mlnet' or a parent directory, or 'C:\Users\Sorin'. Please download the tool and place it in one of the locations.
```

### Known workarounds

To manually copy paket.exe in a reachable location for the FSI. The problem with that is that on a tool version change most probably the tooling around FSI will not work properly anymore.

MecuSorin:
For reference I copy paste here the discussion from the gitter channel:
mecusorin @mecusorin_twitter Nov 12 12:46
Wanted to play with the new fsi features in a new folder:

dotnet new tool-manifest
dotnet tool install paket
dotnet paket init
dotnet paket add FSharp.DependencyManager.Paket
Change the settings.json for VSCode to include

{    
    "FSharp.fsiExtraParameters": [
        "--compilertool:c:/Users/Sorin/.nuget/packages/fsharp.dependencymanager.paket/6.0.0-alpha053"
    ]
}
Then create something.fsx with the content:

#r "paket: FParsec"
open FParsec
And send to the FSI the content of the file
I get the error

somethig.fsx(2,1): error FS3216: Package manager key 'paket' 
was not registered in [C:\Program Files\dotnet\sdk\5.0.100\FSharp; C:\Program Files\dotnet\sdk\5.0.100\FSharp\], [c:/Users/Sorin/.nuget/packages/fsharp.dependencymanager.paket/6.0.0-alpha053]. Currently registered: nuget
What I am doing wrong?


mecusorin @mecusorin_twitter Nov 12 12:56
[...]

PR #3962, opened 2020-12-02
build extension with separate fsharp core version

smoothdeveloper:
This makes the extension work out of box in VS.

Also added some paths to scan paket.exe after suggestions in #3950


smoothdeveloper:
@forki this is green, for review.

forki:
awesome. thanks
next step would be to use fsharp.core 5 everywhere

smoothdeveloper:
somehow the release is still compiled against fsharp.core 4.7.0.0, not sure how it is happening.

image

forki:
Could you please try to update FSharp.Core everywhere?

Gauthier Segay <notifications@github.com> schrieb am Fr., 4. Dez. 2020,
19:38:

PR #3967, opened 2020-12-12
missing redirects: force for the fsi extension,

smoothdeveloper:
not sure if this address the wrong fsharp.core being referenced (still happening in the latest release)

### DIFF
--- a/paket.dependencies
+++ b/paket.dependencies
@@ -60,4 +60,4 @@ group Build
 
 group FSharpDepManagerExtension
   source https://nuget.org/api/v2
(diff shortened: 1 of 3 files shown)


forki:
Maybe it's because we use the wrong sdk?

smoothdeveloper:
This solved the issue, all solved with using paket 🙂

The extension works out of the box in VS and compiled against FSharp.Core 5.0, let see when they start using a newer version.

forki:
yay! good work!

PR #3974, opened 2021-01-03
Use dotnet tool Paket to build Paket

Krzysztof-Cieslak:
This is the first step for removing mono from the build process here, the next one would be updating FAKE to FAKE 5.


forki:
I like where this is going

Krzysztof-Cieslak:
Green - there are some failures on Travis in the `Allowed Failures` section, but I don't think it's related to this change (?). Other than that it is ready to merge.

forki:
hugs

PR #3981, opened 2021-02-04
use lazy initialization to fix UnauthorizedAccessException in Paket.C…

agalisz:
Constants.NuGetCacheFolder can't be initialized without write access to:
Environment.SpecialFolder.LocalApplicationData

Setting "NuGetCachePath" does not help, Constants.NuGetCacheFolder still tries to create LocalApplicationData/Nuget/Cache directory:
_TypeInitializationException: The type initializer for 'Paket.Constants' threw exception.
UnauthorizedAccessException: Access to the path '***/Nuget/Cache' is denied_

Solution:
use Option.orElseWith (lazy initialization) instead of Option.orElse


forki:
thanks!

PR #3983, opened 2021-02-08
don't remove nodes that correspond to packed projects from nuspecs

baronfel:
While working on https://github.com/fsprojects/FsLexYacc/pull/145 I discovered that `paket fix-nuspecs` was removing package references from the nuspec that corresponded to project references, but only when those project references correspond to a packagereference that paket is tracking.

This happens for fslex because it has a project ref on the fslexyacc.runtime project to run the generated lexers and parsers it contains, but it also needs to generate those parsers using the fslexyacc meta-package, which includes a transitive reference to fslexyacc.runtime.  This transitive is what caused the removals.

I'll add a test to this as time permits.


baronfel:
The CI issues seem to be because a roslyn feed is missing?

baronfel:
This test works for me when I manually build and do all the orchestration things the FAKE script would do, since there's a dependency problem there with the roslyn bits it seems.

baronfel:
I believe that the single test failure here is unrelated to my change, as it's checking for Microsoft.CSharp on the v3 endpoint, which I didn't change at all here.  Any thoughts on the implementation?

forki:
thanks

Issue #3992, opened 2021-02-26
DependencyManager Win32Exception on non windows


vilinski:
### Description

Didn't found how to add private nuget sources in F# scripts. So tried to install paket as a package manager.
Followed instructions at https://fsprojects.github.io/Paket/fsi-integration.html 
But this throws `Win32Exception` on Macos

### Repro steps

1. downloaded the `FSharp.DependencyManager.Paket` nuget sending the line to F# interactive

```fsharp
#r "nuget: FSharp.DependencyManager.Paket"
```
2. Copy the manager dll to the folder where current sdk is installed

```
cp ~/.nuget/packages/fsharp.dependencymanager.paket/6.0.0-beta8/lib/netstandard2.0/FSharp.DependencyManager.Paket.dll \
    /usr/local/share/dotnet/sdk/5.0.103/FSharp/
```

3. Created a script, referencing a package using paket

```fsharp
#r "paket: nuget FParsec"

open FParsec
```

4. runned the script

```
$ dotnet fsi paketdeps.fsx
```

### Expected behavior

Expected it would add a package to the script

### Actual behavior

```
❯ dotnet fsi paketdeps.fsx
:paket> using /Users/username/.nuget/packages/paket/5.257.0/tools/paket.exe
exception while resolving dependencies: System.ComponentModel.Win32Exception (8): Exec format error
   at System.Diagnostics.Process.ForkAndExecProcess(String filename, String[] argv, String[] envp, String cwd, Boolean redirectStdin, Boolean redirectStdout, Boolean redirectStderr, Boolean setCredentials, UInt32 userId, UInt32 groupId, UInt32[] groups, Int32& stdinFd, Int32& stdoutFd, Int32& stderrFd, Boolean usesTerminal, Boolean throwOnNoExec)
   at System.Diagnostics.Process.StartCore(ProcessStartInfo startInfo)
   at System.Diagnostics.Process.Start()
   at ReferenceLoading.PaketHandler.ResolveDependenciesForLanguage$cont@176(String targetFramework, String scriptDir, IEnumerable`1 prioritizedSearchPaths, String fileType, String workingDir, FileInfo workingDirSpecFile, String rootDir, FSharpList`1 packageManagerTextLines, String loadScript, Unit unitVar)
[...]

PR #4000, opened 2021-04-11
DependencyManager: fix lookup of paket binary and load-script

iwinux:
This pull request tries to fix two issues mentioned in #3992:

1. `FSharp.DependencyManager.Paket` attempts to launch`paket.exe` directly but on macOS `dotnet tool install -g paket` installs Paket at `$HOME/.dotnet/tools/paket` without the `.exe` extension.

2. the path generated by Paket is actually `<temp-dir>/.paket/load/net50/main.group.fsx`, not `<temp-dir>/.paket/load/net5.0/main.group.fsx` (`net50` v.s. `net5.0`)

### DIFF
--- a/src/FSharp.DependencyManager.Paket/ReferenceLoading.PaketHandler.fs
+++ b/src/FSharp.DependencyManager.Paket/ReferenceLoading.PaketHandler.fs
@@ -14,6 +14,7 @@ let userProfile =
 
 let tweakTargetFramework =
     function
+        | "net5.0" -> "net50"
         | "netcoreapp5.0" -> "net50"
         | targetFramework -> targetFramework
 
--- a/src/FSharp.DependencyManager.Paket/ReferenceLoading.PaketHandler.fs
+++ b/src/FSharp.DependencyManager.Paket/ReferenceLoading.PaketHandler.fs
@@ -3,7 +3,7 @@ module internal ReferenceLoading.PaketHandler
 
 open System
 open System.IO
-let PM_EXE = "paket.exe"
+
 let PM_DIR = ".paket"
 
 let userProfile =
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
@iwinux I'm going to try the branch, although, could you give a shot at minimizing the diff (if there are opportunities, I think the change to pick up both `paket.exe` and `paket` don't warrant the amount of changes) and importantly, use a formatting putting `|>` at the beginning of the lines.

We can have refactoring in separate commits to make it easier to review.

Disregarding, I'm getting the stuff setup to make sure it works on windows.

iwinux:
@smoothdeveloper You're right about that. I've reverted the refactoring to keep changes local and minimal.

smoothdeveloper:
@iwinux thanks a lot! I'm still in middle of upgrades on my windows machine on slow connection but will get to try your changes as soon as possible.

smoothdeveloper:
Ok, I've given it a shot under `dotnet fsi`, `fsi` (net framework) and VS2019, all is clear.

There is only a tiny bit of duplicated call in the `Seq.allPairs` but I think we are fine with the code as is, it solves using it on the other platforms and doesn't regress anything on windows.

Thanks for your patience and working on the fix @iwinux!

smoothdeveloper (approved):

iwinux:
@smoothdeveloper should we be concerned about the failed CI builds?

smoothdeveloper:
I am concerned but they are unrelated to the changes, @forki may know better about the reason (it fails at downloading some roslyn bits).

forki:
@smoothdeveloper https://github.com/fsprojects/Paket/issues/4002

forki:
@iwinux could you please rebase on master? thanks

iwinux:
@forki done

smoothdeveloper:
Same outcome as #4003 so good for merge to me!

forki:
this is released as 6.0.0-beta9 - thanks!

Issue #3994, opened 2021-03-03
paket pack not functioning with net5.0 and mac


runefs:
### Description

When runnin `<?xml version="1.0" encoding="utf-8"?>
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <RootNamespace>Small</RootNamespace>
    <OutputType>Library</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="Environment.fs" />
  </ItemGroup>
  <Import Project="..\.paket\Paket.Restore.targets" />
</Project>

### Repro steps

using the following msbuild file

`<?xml version="1.0" encoding="utf-8"?>
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <RootNamespace>Small</RootNamespace>
    <OutputType>Library</OutputType>
    <TargetFramework>net5.0</TargetFramework>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="Environment.fs" />
  </ItemGroup>
  <Import Project="..\.paket\Paket.Restore.targets" />
</Project>`

1. run `dotnet paket pack --version 1.2.0 .`

### Expected behavior

I would expect it to create the package

### Actual behavior

The result is an error message `Paket failed with
-> Unable to find Release output path node in file <project_file_path>`

### Known workarounds

none

dotnet version 5.0.100
OS: MacOS Big Sur 11.2.2
paket version 5.257

Issue #3980, opened 2021-01-27
Brand new .NET 5.0 project fails to pack


kroppt:
### Description

By default, Paket won't pack a barebones .NET 5.0 project, created through dotnet CLI.

Using Paket version 5.257.0.

### Repro steps

```bash
#!/bin/bash
mkdir TestPaketPack
cd TestPaketPack
dotnet new classlib
paket init
echo > paket.dependencies
touch paket.references
cat <<EOF > paket.template
type project
id TestPaketPack
authors Author
description
  Tests Paket's pack functionality.
version 1.2.3
EOF
paket install
dotnet build -c Release
paket pack nugets
```

If possible then please create a git repository with a repro sample or attach a zip to the issue.

### Expected behavior

```text
Packaging: C:\absolute\path\to\TestPaketPack\paket.template
Wrote: C:\absolute\path\to\TestPaketPack\nugets\TestPaketPack.1.2.3.nupkg
```

### Actual behavior

```text
Paket failed with
-> Unable to find Release output path node in file C:\absolute\path\to\TestPaketPack\TestPaketPack.csproj for any known platforms
```

### Known workarounds

Replace `net5.0` with `net50` in the `TargetFramework` node of the project file.

kroppt:
I tested this again with Paket version 6.2.1+c82f25d6c324024cbd231df154a447a117c2546e and sure enough it was fixed.

PR #4007, opened 2021-04-22
Fix dotnet pack on net5.0

thinkbeforecoding:
This is a fix for dotnet paket pack on net5.0
This should solve #3980  and #3994 
The short version string for net5.0 is no 50 but 5.0

The rest of the tests seems to work and this is conform to net5.0 (inconsistent) versioning so it seems legit.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -8,7 +8,6 @@ if errorlevel 1 (
 
 setlocal
 
-set MSBuild=%~dp0packages\build\RoslynTools.MSBuild\tools\msbuild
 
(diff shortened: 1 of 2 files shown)


thinkbeforecoding:
On the way I removed the set MSBUILD from build.cmd that was pointing to a packages/ path that has been removed in a previous commit but has not been fixed for windows

thinkbeforecoding:
The appveyor built failed but it seems unrelated to the actual change

forki:
thanks!

Issue #4017, opened 2021-05-12
"copy_content_to_output_dir: preserve_newest" breaks paket.lock


andrewrondeau-optirtc:
### Description

When I put "copy_content_to_output_dir: preserve_newest" into my paket.dependencies file, an incorrect paket.lock is generated. paket.lock is incorrectly generated with "COPY-CONTENT-TO-OUTPUT-DIR: PRESERVE**-**NEWEST". I can no longer run ".paket/paket install" or build any projects until I update paket.lock to have "PRESERVE**_**NEWEST"

### Repro steps

Please provide the steps required to reproduce the problem

1. Add "copy_content_to_output_dir: preserve_newest" near the beginning of your paket.dependencies file
2. Run ".paket/paket install"
3. Re-run ".paket/paket install". Observe that it fails
4. Try to build any of the .csproj files in your solution. Observe that : The command ""...\.paket\paket.exe" restore --references-file "..."" exited with code 1. (Edited to remove personal information)

### Expected behavior

* paket.lock should have "COPY-CONTENT-TO-OUTPUT-DIR: PRESERVE_NEWEST"
* You should be able to re-run ".paket/paket install"
* You should be able to build your solution

### Actual behavior

Re-running ".paket/paket install" fails with:

`Paket version 5.257.0
Performance:
 - Runtime: 491 milliseconds
Paket failed with
-> Error during parsing of '...\opti\paket.lock'.
-> Unknown copy_content_to_output_dir settings: "preserve-newest"
`

Projects in my solution fail with:
`1>  Paket failed with
1>  -> Error during parsing of '...\paket.lock'.
1>  -> Unknown copy_content_to_output_dir settings: "preserve-newest"
`

### Known workarounds

Manually edit paket.lock to have "COPY-CONTENT-TO-OUTPUT-DIR: PRESERVE_NEWEST"

andrewrondeau-optirtc:
Fix: https://github.com/fsprojects/Paket/pull/4018

(I verified this in our build environment)

PR #4018, opened 2021-05-12
Fixed issue when paket.dependencies has "copy_content_to_output_dir: preserve_newest"

andrewrondeau-optirtc:
Original bug: https://github.com/fsprojects/Paket/issues/4017

> When I put "copy_content_to_output_dir: preserve_newest" into my paket.dependencies file, an incorrect paket.lock is generated. paket.lock is incorrectly generated with "COPY-CONTENT-TO-OUTPUT-DIR: PRESERVE-NEWEST".


forki:
thanks

Issue #4031, opened 2021-06-13
Support PackageReadmeFile in nuget pack task


baronfel:
### Description

SDK-style projects can set `PackageReadmeFile` to the path of a file in the generated package and nuget.org will show that readme inline.  We should add this to the paket invocation of the nuget pack task.  This is supported since .net sdk 5.0.300 and details can be seen [here](https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#packagereadmefile). The msbuild version for this is 16.10, so we could make a new condition in the current paket.local.targets to determine this, and only add this property if we're at least that version. this is very similar to the existing pattern.

PR #4032, opened 2021-06-13
add support for PackageReadmeFile

baronfel:
This adds support for the PackageReadmeFile msbuild property when msbuild greater than 16.10 is used.  This should fix https://github.com/fsprojects/Paket/issues/4031.


baronfel:
I tweaked the Paket.Restore.targets and used them locally on fsautocomplete [here](https://github.com/fsharp/FsAutoComplete/pull/792) and the readme was linked through to the created package as expected.

forki:
thx

Issue #4028, opened 2021-06-11
Packages are extracted with incorrect permissions


baronfel:
### Description

I was testing Microsoft.Playwright for browser testing, and found that when I installed the package in my project and restored it with Paket, Paket extracted the package with read-only permissions on the entire directory tree (`r--r--r--`), which makes the package unusuable because it attempts to invoke files from the package that are copied to your test projects' output directory.

After trying the package with Nuget, I found that Nuget marks the extracted directories with `rwxr-xr-x`, and the extracted binaries/files with `rwxr--r--`, and those permissions enable the package to work.

### Repro steps

1. Make a new project according to the instructions [here](https://playwright.dev/dotnet/docs/intro/#first-project), but use Paket to install the Microsoft.Playwright dependency instead of nuget.

2. The `playwright install` step should fail due to permissions of script files copied to the output directory

### Expected behavior

Playwright should work

### Actual behavior

Playwright encounters permissions errors so it can't launch the supporting binaries.

### Known workarounds

You can do a `chmod -R <path to output dir>/.playwright` to fix the permissions of the files.

baronfel:
Issue originally raised over at https://github.com/fsprojects/Paket/issues/4028 but closed after some investigation.

PR #4033, opened 2021-06-14
add nuget dependency and use it to extract package archives

baronfel:
I took at stab at solving #4028 in a more holistic fashion.

We now use the NuGet.Packaging APIs to actually extract the downloaded packages. That's all we use them for.  This should ensure that we behave exactly the same as they do for a non-core function (especially in terms of file permissions and such), and are more free to devote our time to where we do things differently.

I only used this on the main `ExtractPackageToUserFolder` in NuGetCache.fs, since all other entry points I saw led to that.

I was able to test this locally using that same playwright package that errored before by
* deleting that folder from my global cache
* building this change
* deleting my Paket.Restore.cached from a project's paket-files directory
* running `dotnet path/to/build/paket.dll restore --verbose`
* inspecting the logged output to verify the package was downloaded
* using `ls` to inspect the file permissions

Happy to say that this all works!


integrationtests/Paket.IntegrationTests/App.config
```diff
@@ -9,15 +9,10 @@
     <assemblyIdentity name="FSharp.Core" publicKeyToken="b03f5f7f11d50a3a" culture="neutral" />
     <bindingRedirect oldVersion="0.0.0.0-65535.65535.65535.65535" newVersion="5.0.0.0" />
   </dependentAssembly>
-  <dependentAssembly>
```
baronfel, line 12: some of these app.configs got changed a lot after adding the new nuget.packaging package....

forki:
I assume we need to ILMerge this one in build.fsx!?

baronfel:
ah yeah, you're right. I hadn't looked at the packaging story at all.  I'll do that today.  Do you have any ideas about other areas I might want to look into?  I suppose I need to add a spec for this package as well...

baronfel:
@forki do you know if there's a reason we don't just glob all the dlls + the paket.exe from the bin/net461 directory as inputs to ilrepack instead of the manual accounting that happens now?  I'm guessing the original reason was something like:

> the build copies .net framework assemblies to the publish directory, but ILRepack doesn't actually need all of the framework assemblies in that directory (because when executing in .net framework that are implicitly provided), so we exclude those from the list of repacked assemblies

baronfel:
I did a diff between the output of the `bin/net461` directory on master and this branch using the `diff` command to get the following list of necessary assemblies:

```
➜  Paket git:(extract-like-nuget) ✗ diff bin/net461 bak/net461 | grep 'Only in' | cut -d' ' -f 4
NuGet.Common.dll
NuGet.Configuration.dll
NuGet.Frameworks.dll
NuGet.Packaging.dll
NuGet.Versioning.dll
System.Configuration.ConfigurationManager.dll
System.Security.Cryptography.Cng.dll
System.Security.Cryptography.Pkcs.dll
System.Threading.Tasks.Extensions.dll
```

I added those to the ilrepack list, but when I tried to test the net461 merged executable with mono (using `mono paket.exe`) I got the following error:

```
➜  merged git:(extract-like-nuget) ✗ mono paket.exe
Assembly '/Users/chethusk/oss/Paket/bin/merged/paket.exe' doesn't have an entry point.
```

which does worry me somewhat.

baronfel:
Here's the graphical paket.lock diff after this change

paket-diff

baronfel:
test failure is a v3 download check: `#2700-1 v3 works properly`, not sure that I changed that here.

baronfel:
Ok, I've added a test that convinces me this is working.  To get the test to run I had to fix the invocation of the netcore tests, because the integration test project isn't netcoreapp3.0, it's 3.1. So that have been false-positive green for an unknown amount of time.

paket.dependencies
```diff
@@ -12,11 +15,15 @@ nuget System.Security.Cryptography.ProtectedData >= 4.4
 nuget NETStandard.Library ~> 2.0
 nuget Microsoft.NETCore.App ~> 2.1
 nuget System.Net.Http.WinHttpHandler >= 4.5
+nuget Microsoft.NETFramework.ReferenceAssemblies.net461
+nuget Nuget.Packaging
```
baronfel, line 19: I had to add these explicitly to make the netcore3.1 build of paket atually invoke correctly. otherwise the integration tests would blow up at runtime saying these were missing from the runtime.deps.json.

src/Paket.Core/Dependencies/NuGetCache.fs
```diff
@@ -454,13 +455,27 @@ let getDetailsFromCacheOr force nugetURL (packageName:PackageName) (version:SemV
...
+                                    override this.SaveToDisk(): unit = ()
+                                    [<CLIEvent>]
+                                    override this.SettingsChanged: IEvent<EventHandler,EventArgs> = Event<EventHandler,EventArgs>().Publish
+                                    }
+let private signingContext = Signing.ClientPolicyContext.GetClientPolicy(nugetSettings, NuGet.Common.NullLogger.Instance)
+let private extractionContext = NuGet.Packaging.PackageExtractionContext(PackageSaveMode.Defaultv3, XmlDocFileSaveMode.Compress, signingContext, NuGet.Common.NullLogger.Instance)
+
 /// Extracts the given package to the user folder
-let rec ExtractPackageToUserFolder(fileName:string, packageName:PackageName, version:SemVerInfo, kind:PackageResolver.ResolvedPackageKind) =
+let rec ExtractPackageToUserFolder(source: PackageSource, fileName:string, packageName:PackageName, version:SemVerInfo, kind:PackageResolver.ResolvedPackageKind) =
```
baronfel, line 473: the packagesource had to be added here so that we could get the url from it, because the nuget.packaging APIs expect that as the key that's used for signing verification.

note that we don't do signing verification here, but now we have the scaffolding to do so using nuget's own APIs.  if we modify the `signingContext` above and preload it with hashes from our lockfile for example, then we'd start getting signing validation for free.

baronfel:
The new test seems to fail when run on the .net framework version of paket, due to some kind of double-stomping on the .nuget files in the directories? I am unsure what that does.

baronfel:
Alright! My test passes now. The failing test is the same one form my other PR, where the v3 test is failing. I think it's happening on every branch right now.

baronfel:
@forki thoughts on this one?

baronfel:
Should be unblocked by #4042

forki:
very cool. Let's give it a shot!

forki:
image

just tried to release it. but it's unhappy that I run it on windows. Ideas?

baronfel:
None right off the top of my head; is this something that I could replicate on my machine easily, or does it really on secrets that only you have to get to this step?

baronfel:
Translation of the error for me in the future:

> The specified executable is not a valid application for this operating system platform

baronfel:
So the built paket.exe (non-merged) runs fine for me, so the problem is likely in the ilrepack step I'd say.

baronfel:
Figured it out: I messed up the order of the ilrepack arguments. ilrepack expects the 'primary' assembly to be the first positional parameter and during my refactor to make the list of input assemblies a multiline list I messed up that order. PR with fix + comments coming soon.

Issue #4039, opened 2021-06-30
[Question] Garbage files imported from dependency


MiloszKrajewski:
### Description

I don't think it is actual bug but maybe you know how to address it.

When importing dependency with `Content` (I guess) files, these files are shown in your assembly. Argghh... see screenshot:
![image](https://user-images.githubusercontent.com/4333436/123936394-0c01e900-d98d-11eb-9910-96260b2ec5a1.png)

all those `libmongocrypt.dll` and `snappy64.dll` are actually from `MongoDB.Driver` but they are "invading" my assembly.
They cannot be excluded (I tried to "hide" them somehow).
Same thing happens in VS.

### Repro steps

import `MongoDB.Driver 2.11.6` into your solution.

### Expected behavior

"Foreign" files are not visible in solution, as they are not shown when using `<PackageReference Include="..." Version="..." />`.

### Actual behavior

"Foreign" files are visible in solution.

### Known workarounds

Using `<PackageReference Include="..." Version="..." />` for one assembly.

baronfel:
paket has a mechanism to ignore content from dependencies (`content: none`), but that mechanism doesn't seemed to be hooked up to the PackageReference mechanism that paket uses to interop with the .net SDK.

l3m:
Which in turn means that every .Net 5 project using Paket and FSharp.Core will add FSharp.Core.xml as un-removable file to the project.

baronfel:
Specifically, the files are coming from from the [Mongo.Libmongocrypt](https://www.nuget.org/packages/MongoDB.Libmongocrypt/) package, which has both the `content` and `contentFiles` node in its nuget package:

```
tree ~/Downloads/mongodb.libmongocrypt.1.2.1.nupkg 
/Users/chethusk/Downloads/mongodb.libmongocrypt.1.2.1.nupkg
├── License.txt
├── MongoDB.Libmongocrypt.nuspec
├── [Content_Types].xml
├── _rels
├── build
│   └── MongoDB.Libmongocrypt.targets
├── content
│   ├── libmongocrypt.dylib
│   ├── libmongocrypt.so
│   └── mongocrypt.dll
├── contentFiles
│   └── any
│       ├── net452
│       │   ├── libmongocrypt.dylib
│       │   ├── libmongocrypt.so
│       │   └── mongocrypt.dll
│       ├── netstandard1.5
│       │   ├── libmongocrypt.dylib
│       │   ├── libmongocrypt.so
[...]

PR #4040, opened 2021-06-30
thread through existing omitcontent setting into ExcludeAssets=contentFiles in the SDK

baronfel:
This should allow for closing #4039 by letting users specify `content: none` on a group, dependency, or reference-level.

We follow the existing patterns for the copy_local setting, and we alter the generated PackageReference nodes in both places they are generated:

* paket.props file
* Paket.Restore.targets

In addition, there is a usability tweak for new users: when we create the fsharp.core dependency in the converted paket.dependencies file we automatically add `content: none` to it, because FSharp.Core > 5.0 comes with the xml doc in the contentFiles folders, so users start getting FSharp.Core.xml listed as a content file in their apps.


baronfel:
@forki since I'm changing the format of the generated props files, do I need to update the version info on [this line](https://github.com/baronfel/Paket/blob/support-omitcontent-in-sdk/src/Paket.Core/Installation/RestoreProcess.fs#L348) and the associated restore.targets version?

baronfel:
The failing test is the same one from the other PRs: the v3 validation check...

baronfel:
Should be unblocked by #4042

forki:
thanks!

PR #4042, opened 2021-07-01
fix v3 validation test by pointing to a known-good v3 url

baronfel:
This fixes a failing test that has blocked green builds for several recent PRs.

The root cause of the failure was the pinned v3 nuget feed was no longer reachable, and the solution is providing a new url for one that is reachable (ie the main nuget feed).


baronfel:
Green! at last!

forki:
Thanks!

Issue #4005, opened 2021-04-17
import_targets: false has no effect (on Gprc.Tools in netcoreapp3.1 F# exe)


marklam:
### Description

Gprc.Tools stops the build from its targets file if the project being built is not a .csproj, but `import_targets: false` does not prevent those targets being loaded when used in the paket.references or paket.dependencies files.

(Found this because a package I added had Gprc.Tools as a dependency)

### Repro steps

```
mkdir \temp\whatever
cd \temp\whatever
dotnet new console -f netcoreapp3.1 -lang F#
dotnet new tool-manifest
dotnet tool install paket
dotnet paket convert-from-nuget
dotnet paket add Grpc.Tools --project whatever
edit the paket.references to include 'Grpc.Tools import_targets: false'
or edit the paket.dependencies to include 'nuget Grpc.Tools import_targets: false'
dotnet restore
dotnet build
```
Git repo containing the same: https://github.com/marklam/paket-import-targets-false

### Expected behavior

Project builds

### Actual behavior

Error from Gprc.Tools' targets file
`C:\Users\Mark\.nuget\packages\grpc.tools\2.37.0\build\_protobuf\Google.Protobuf.Tools.targets(51,5): error : Google.Protobuf.Tools proto compilation is only supported by default in a C# project (extension .csproj) [C:\temp\whatever.fsproj]`

### Known workarounds

None

mvkara:
I was having this issue when importing Grpc.AspNetCore into an F# project since it takes Grpc.Tools as a dependency. Although I'm not sure why, since the proto classes are generated in another project. Effectively it means you can't host Grpc in F# using Paket.

I found a workaround that allows the build to pass. See here: https://github.com/grpc/grpc/issues/20515#issuecomment-629415019. Since you aren't actually compiling any proto files anyway (for myself this is done in another C# project and referenced in) all this does is bypass the validation and allow the build to pass.

Ideally however target imports from transitive dependencies could also be disabled.

baronfel:
I think this is going to be similar to #4039: PackageReference nodes have [different values for `ExcludeAssets`](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) that can be set to control this behavior in msbuild, and paket isn't setting all of them just yet. Thankfully, #4040 shows this is a pretty simple process to add, so threading through the current `import_targets` behavior into including `ExcludeAssets=build;buildMultitargeting;buildTransitive` should be pretty simple to do.

PR #4044, opened 2021-07-02
thread through import_targets: false to msbuild exclude assets

baronfel:
This should fix https://github.com/fsprojects/Paket/issues/4005 in a manner similar to #4040, by setting the appropriate ExcludeAssets flags when the user requests targets to not be imported.

This also fixes a bug in the Paket.Restore.targets I introduced with my fix in #4040 where the comparisons were incorrect. Inspection of the generated PackageReference in a msbuild structured build log revealed this.

With this change, the sample project described in #4005 builds without errors (since the targets aren't imported):

```
mkdir \temp\whatever
cd \temp\whatever
dotnet new console -f netcoreapp3.1 -lang F#
dotnet new tool-manifest
dotnet tool install paket
dotnet paket convert-from-nuget
dotnet paket add Grpc.Tools --project whatever
# edit the paket.references to include 'Grpc.Tools import_targets: false'
dotnet restore
dotnet build
```

### DIFF
--- a/src/Paket.Core/Installation/RestoreProcess.fs
+++ b/src/Paket.Core/Installation/RestoreProcess.fs
@@ -12,6 +12,11 @@ open Chessie.ErrorHandling
 open System.Reflection
 open Requirements
 
+/// Combines the import targets settings from the lock file and a project's references file, so that the project settings take priority
+let private combineImportTargets (resolvedSettings: InstallSettings) (packageInstallSettings: PackageInstallSettings) =
+    packageInstallSettings.Settings.ImportTargets
+    |> Option.orElse resolvedSettings.ImportTargets
+
 /// Combines the content settings from the lock file and a project's references file, so that the project settings take priority
 let private combineOmitContent (resolvedSettings: InstallSettings) (packageInstallSettings: PackageInstallSettings) = 
     packageInstallSettings.Settings.OmitContent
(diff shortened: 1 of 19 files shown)


src/Paket.Core/Installation/RestoreProcess.fs
```diff
@@ -452,15 +459,21 @@ let createProjectReferencesFiles (lockFile:LockFile) (projectFile:ProjectFile) (
...
+                        | _ -> "true"
                     let line =
-                        packageName.ToString() + "," +
-                        package.Version.ToString() + "," +
-                        (if direct then "Direct" else "Transitive") + "," +
-                        kv.Key.ToString() + "," +
-                        privateAssetsAll  + "," +
-                        copyLocal + "," +
-                        omitContent
+                        [ packageName.ToString()
```
baronfel, line 468: kept the order here, but refactored for easier diffs

src/Paket.Core/embedded/Paket.Restore.targets
```diff
@@ -238,13 +238,15 @@
...
       </PaketReferencesFileLinesInfo>
       <PackageReference Include="%(PaketReferencesFileLinesInfo.PackageName)">
         <Version>%(PaketReferencesFileLinesInfo.PackageVersion)</Version>
         <PrivateAssets Condition=" ('%(PaketReferencesFileLinesInfo.AllPrivateAssets)' == 'true') Or ('$(PackAsTool)' == 'true') ">All</PrivateAssets>
         <ExcludeAssets Condition=" '%(PaketReferencesFileLinesInfo.Splits)' == '6' And %(PaketReferencesFileLinesInfo.CopyLocal) == 'false'">runtime</ExcludeAssets>
         <ExcludeAssets Condition=" '%(PaketReferencesFileLinesInfo.Splits)' != '6' And %(PaketReferencesFileLinesInfo.AllPrivateAssets) == 'exclude'">runtime</ExcludeAssets>
-        <ExcludeAssets Condition=" '%(PaketReferencesFileLinesInfo.Splits)' != '7' And %(PaketReferencesFileLinesInfo.OmitContent) == 'true'">$(ExcludeAssets);contentFiles</ExcludeAssets>
+        <ExcludeAssets Condition=" '%(PaketReferencesFileLinesInfo.Splits)' == '7' And %(PaketReferencesFileLinesInfo.OmitContent) == 'true'">$(ExcludeAssets);contentFiles</ExcludeAssets>
```
baronfel, line 248: I'm thinking these conditionals are wrong: the split seems to be between versions of the resolved file that have 6 segments (some old Paket version), and versions of the resolved file that have more than 6 segments.

baronfel: Fixed them after testing locally.

baronfel:
I've added 3 new integration tests to cover the package reference enhancements I've made lately, and they're all green on my local host.

baronfel:
And they pass in CI too! The only failures that remain are the ILRepack args that I've fixed in #4043

baronfel:
Updated the existing documentation to call out how these new changes map to PackageReference usages.

forki:
thanks!!

PR #4046, opened 2021-07-15
CSX scripts: remove `namespace` declaration

andersstorhaug:
I use Roslyn C# scripting in several projects, and have discovered that `generate-load-scripts` looks like a nice way to prepare NuGet dependencies for deployment artifacts.

However, there is one issue: AFAIK C# scripts [don't support the `namespace` keyword](https://github.com/dotnet/roslyn/issues/4478).

This is a minor change that will simply omit the `namespace PaketLoadScripts` line in CSX scripts.


baronfel:
This is a good change, and thanks for adding tests!

PR #4050, opened 2021-07-18
tweak paket.resolved split handling to be correct

### LINKED ISSUE #4049: Parsing of PaketReferencesFileLinesInfo from PROJECT.paket.resolved files is broken
### Description

The generated PackageReference ExcludeAssets metadata is generated incorrectly, due to logic errors in the properties set in the `PaketReferencesFileLinesInfo` block. This results in no metadata except `ImportTargets` being respected.

### Repro steps

Let's take a line from fsautocomplete's net5.0.paket.resolved file as an example (this is for a dependency with `copy_local: false`) set:

```
Microsoft.Build.Framework,16.10,Direct,Main,false,false,false,true
```

Based on this, we expect the following `PaketReferencesFileLinesInfo`:
* Splits: 8
* PackageName: Microsoft.Build.Framework
* PackageVersion: 16.10
* AllPrivateAssets: false
* CopyLocal: false
* OmitContent: false
* ImportTargets: true

Instead, we get the following:

```
Microsoft.Build.Framework,16.10,Direct,Main,false,false,false,true
    PackageName = Microsoft.Build.Framework
    Splits = 8
    AllPrivateAssets = false
    PackageVersion = 16.10
    ImportTargets = true
```

Note that CopyLocal and OmitContent aren't set at all.

As a result, the PackageReference generated is

```
Microsoft.Build.Framework
    Version = 16.10
    AllowExplicitVersion = true
    GeneratePathProperty = true
```

Note the lack of the `ExcludeAssets=runtime` property. This triggers a build error in the targets for these packages.

### Expected behavior

`ExcludeAssets=runtime` should be set in this case, and CopyLocal, OmitContent nodes are correctly assigned.

### Actual behavior

ExcludeAssets is not set, resulting in a build error.

### Proposed solution

* The Splits comparison on all optional properties needs to be a greater-than comparison
* In the generation of the ExcludeAssets conditions, the check for == 6 on the CopyLocal check needs to be != 6 liek the others.
  *  better yet, the check on splits may not be needed at all in the ExcludeAssets conditions, because we also check for the existence of the property we care about.

baronfel:
split comparison needs to be >= in order to assign the correct values.
parameter checking in the ExcludeAssets doesn't need to care about splits.

Fixes #4049

### DIFF
--- a/src/Paket.Core/embedded/Paket.Restore.targets
+++ b/src/Paket.Core/embedded/Paket.Restore.targets
@@ -236,17 +236,16 @@
         <PackageName>$([System.String]::Copy('%(PaketReferencesFileLines.Identity)').Split(',')[0])</PackageName>
         <PackageVersion>$([System.String]::Copy('%(PaketReferencesFileLines.Identity)').Split(',')[1])</PackageVersion>
         <AllPrivateAssets>$([System.String]::Copy('%(PaketReferencesFileLines.Identity)').Split(',')[4])</AllPrivateAssets>
-        <CopyLocal Condition="'%(PaketReferencesFileLinesInfo.Splits)' == '6'">$([System.String]::Copy('%(PaketReferencesFileLines.Identity)').Split(',')[5])</CopyLocal>
-        <OmitContent Condition="'%(PaketReferencesFileLinesInfo.Splits)' == '7'">$([System.String]::Copy('%(PaketReferencesFileLines.Identity)').Split(',')[6])</OmitContent>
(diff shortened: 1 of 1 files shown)


baronfel:
Given the same example from the linked issue, we now get the following PaketReferencesFileLinesInfo for Microsoft.Build.Framework:

```
Microsoft.Build.Framework,16.10,Direct,Main,false,false,false,true
    PackageName = Microsoft.Build.Framework
    OmitContent = false
    Splits = 8
    AllPrivateAssets = false
    PackageVersion = 16.10
    ImportTargets = true
    CopyLocal = false
```

Note that all properties are correctly parsed.

And the generated PackageReference:

```
Microsoft.Build.Framework
    Version = 16.10
    ExcludeAssets = runtime
    AllowExplicitVersion = true
    GeneratePathProperty = true
```

Note the included ExcludeAssets property.

This is the intended shape of things, IMO.

PR #4055, opened 2021-07-30
Fixed parsing of net6 targets

### LINKED ISSUE #4054: Paket does not recognize net6.0-android30.0
### Description

Recently Xamarin added new platform to its packages. New platfrom breaks Paket:

```Could not detect any platforms from 'net6.0-android30.0' in Xamarin.AndroidX.Collection 1.1.0.8, please tell the package authors```

Since today I cannot use Paket with Xamarin.Android without pinning packages to the oldest version.

vshapenko:
Closes #4054


forki:
thanks!

PR #4071, opened 2021-08-13
Fix new Paket.Restore.targets 'PackageReadmeFile'

beauvankirk:
The new added support for PackageReadmeFile in MSBuild>=16.10 appears to use incorrect parameter name (caused failures for me). 
It should be 'Readme' rather than 'PackageReadmeFile'.
Extra space in NoDefaultExcludes also caused failure.


baronfel:
Yep, this is indeed the correct Pack Task Property Name, see [the targets](https://github.com/NuGet/NuGet.Client/blob/dc22f0ba7909a47194e70d12275424b165df1e0f/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L278) and [the task definition](https://github.com/NuGet/NuGet.Client/blob/dc22f0ba7909a47194e70d12275424b165df1e0f/src/NuGet.Core/NuGet.Build.Tasks.Pack/PackTask.cs#L75).

PR #4076, opened 2021-08-20
Package metadata

markpattison:
I hope this gives consistent project and repo URLs on https://www.nuget.org/

- Paket
- Paket.Core
- FSharp.DependencyManager.Paket

### DIFF
--- a/src/FSharp.DependencyManager.Paket/FSharp.DependencyManager.Paket.fsproj
+++ b/src/FSharp.DependencyManager.Paket/FSharp.DependencyManager.Paket.fsproj
@@ -2,6 +2,13 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <!-- package metadata -->
+    <Description>A dependency manager for .NET with support for NuGet packages and git repositories.</Description>
+    <PackageProjectUrl>http://fsprojects.github.io/Paket</PackageProjectUrl>
+    <PackageLicenseUrl>http://fsprojects.github.io/Paket/license.html</PackageLicenseUrl>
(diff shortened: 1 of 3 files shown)


baronfel:
I think adding the README would be useful too, since the nuget beta design puts readmes front and center:

image


[Here's](https://github.com/fsharp/FsAutoComplete/pull/792/files#diff-fbfdf221e61d9ee5995990c90109910b3a1578733a6635662df3f6185b343384R10-R18) how I did this in FSAC, you've got to add the readme to the package and specify the relative path to it from the package root as a property.

markpattison:
@baronfel I've added the readme to the same 3 packages and also replaced the deprecated `packagelicenseUrl` fields.

I can't work out why, but the repo url doesn't make it into the package metadata.

markpattison:
Repo URLs (and type) are now appearing correctly in the `nuspec` files.  Should be good to go.

forki:
image

forki:
It's now breaking the release process

baronfel:
I think this is a bug with paket's template-file packaging. the two packages that don't use paket templates seem to be well-formed (precisely how fsautocomplete does it).

@forki do you remember how/why paket.core isn't using `dotnet pack` to package? it might be worth moving to that.

forki:
I do not remember. sorry

markpattison:
Looks like the Paket.Core `nupkg` has put `README.md` inside a folder named `_` within the zip, rather than in the root directory.

I put the following in the `paket.template` file:

```
files
    ../../README.md ==> /
```

I'll see if I can get it to work.

forki:
image

yet another one

forki:
yay thanks for the help. Paket 6.0.13 is online

PR #4079, opened 2021-08-31
Fix Paket.Core package metadata

markpattison:


markpattison:
Hopefully fixes remaining NuGet upload issue in #4076

PR #4083, opened 2021-09-15
unify the setting of version + package metadata in the 461 and .net core builds of Paket

### LINKED ISSUE #4082: Wrong version number?
### Description

The latest paket version number reporting seems odd

grafik

Is this intentional?

baronfel:
Closes #4082 

The merged .Net Framework 4.6.1 build of paket.exe used the 'raw' published binaries build from the `publish` step of the build script, which themselves used the raw outputs of the `build` step of the build script. This `build` step didn't assign version numbers to the version msbuild properties, and so that now-required metadata was no longer present.

This MR therefore
* ensures that versioning information is set at build-time, (and implicitly publish-time), not just pack-time
* tells ILRepack to migrate over attributes as well when doing the merge, so that the assemblyVersion is maintained on the merged binary

Verification of this can be seen by using ilspy or another decompiler to look at the AssemblyInformationalVersion attributes of the `tools/paket.exe` and `tools/netcoreapp2.1/paket.dll` files, which both should be 6.1.0 + a build hash after this change.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -134,11 +134,36 @@ Target "CleanDocs" (fun _ ->
 // --------------------------------------------------------------------------------------
 // Build library & test project
 
+let releaseNotesProp releaseNotesLines =
+    let xn name = XName.Get(name)
+    let text = releaseNotesLines |> String.concat Environment.NewLine
+    let doc =
+        XDocument(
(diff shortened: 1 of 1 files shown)


forki:
image


this is what I get locally

PR #4104, opened 2021-12-13
In NuGetCache change the XmlDocFileSaveMode to None

### LINKED ISSUE #4103: Paket not extracting XML docs correctly, breaks XML doc intellisense
Version 6.2.1

### Description

When paket extracts `.xml` files from nuget packages, it extracts them as .zip files. eg. `Microsoft.CSharp.xml.zip`.

### Repro steps

```
# from a solution folder that uses paket with `storage: none`

dotnet nuget locals -c all
rm -rf paket-files # make sure paket restores files
dotnet paket restore

cd ~/.nuget/packages
find . -name "*.xml*"

# observe hundreds of .xml.zip files
```
### Known workarounds

Run a script to unzip the `.xml` doc files in the nuget global cache

seanamos:
Resolves #4103

### DIFF
--- a/integrationtests/scenarios/i004002-pack-global-tools/before/tool1.csproj
+++ b/integrationtests/scenarios/i004002-pack-global-tools/before/tool1.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp2.1</TargetFramework>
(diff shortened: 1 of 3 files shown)


baronfel:
The test failures are due to a new flag in [the SDK](https://docs.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/duplicate-files-in-output) that checks for duplicate files being published to the publish output directory. We should set the `ErrorOnDuplicatePublishOutputFiles` property to `false` for these two tests, or fix the underlying publish issue, but since these are tests I think the minimal effort is better.

PR #4109, opened 2022-01-10
added `--release-notes-file` option to pack

krauthaufen:
Hey, I added an option to parse a standard-markdown release notes file (via Fake.Core.ReleaseNotes) which could simplify packaging a lot when not using FAKE scripts.

If there are any concerns/questions/suggestions please let me know.

Cheers


forki:
thanks for contribution. It's released in 7.0.0-alpha005

PR #4135, opened 2022-03-29
NuGet Cache: avoid duplicate nupkg file

cdrnet:
Currently Paket ends up putting each nupkg file twice into the NuGet cache: once in the folder of the package, and once in the subfolder for the specific version. NuGet itself puts it only into the version subfolder - as did Paket before v6.

The reason for this seems to be that we first download the nupkg to the expected folder. However, because NuGet's PackageExtractor.InstallFromSourceAsync doesn't support inplace extraction where the nupkg is already there in the correct place, we first move it to the parent folder and then extract from there.

The proposed change would delete that file in the parent folder after the extraction.

Advantage of deleting:
 - Closer to NuGet behavior
 - Better compatibility with tools which get confused and break by the extra file (e.g. NUKE).
 - No redundant file, less disk usage


cdrnet:
Thank you very much for the super quick release! 🚀

PR #4138, opened 2022-04-08
Fix exception when searching for config file

### LINKED ISSUE #4137: UnauthorizedAccessException when directory without read access exists
### Description

If **any** subdirectory **anywhere** in the project directory exists that doesn't have read access, a paket install fails with an `UnauthorizedAccessException`.

The reason I have a directory without read access is because I run a docker container with a data volume mounted. There is a directory in the volume that removes access for other users.

### Repro steps

Repro steps are for linux

1. Start with any project that uses paket and has a `paket.references` file
2. Inside the project, add a new directory `mkdir data_volume`
3. Change owner `sudo chown root:root data_volume`
4. Change permissions `chmod go-r data_volume`
5. `dotnet paket install`

### Expected behavior

Paket installs dependencies as normal.

### Actual behavior

```
Paket version 7.1.3+c3f7fc17b8e07eeb633f61549fd0259920f742d5
Skipping resolver for group Main since it is already up-to-date
Installing into projects:
Created dependency graph (2 packages in total)
Total time taken: 0 milliseconds
Paket failed with
-> UnauthorizedAccessException: Access to the path '/home/oliver/paket_repro/data_volume' is denied.
-> IOException: Permission denied
```

### Known workarounds

Relocate the data volume outside the project folder if possible

### Cause

Using a debugger I've traced the cause to https://github.com/fsprojects/Paket/blob/c11f2580967ec4d67522b875038e9111826f3c21/src/Paket.Core/Installation/BindingRedirects.fs#L93-L94

This attempts to find a `app.config` or `web.config` file by first getting an array of all `*.config` files in the entire directory tree, which is where it runs into the error.

This is probably also potential for a bug if it accidentally finds an unrelated config file, and is also a potential performance hit where scanning the entire project directory tree might take a considerable ammount of time.

olivercoad:
Fixes #4137

### DIFF
--- a/src/Paket.Core/Installation/BindingRedirects.fs
+++ b/src/Paket.Core/Installation/BindingRedirects.fs
@@ -83,17 +83,6 @@ let internal indentAssemblyBindings config =
         if not parent.HasElements then
             parent.Remove()
 
-let private configFiles = [ "app"; "web" ] |> Set.ofList
-let private projectFiles = [ ".csproj"; ".vbproj"; ".fsproj"; ".wixproj"; ".nproj"; ".vcxproj"; ".pyproj"; ".sfproj"; ".sqlproj" ] |> Set.ofList
-let private toLower (s:string) = s.ToLower()
-let private isAppOrWebConfig =
-    configFiles.Contains << (fun (x : string) -> Path.GetFileNameWithoutExtension(x) |> toLower)
-
-let internal getConfig getFiles directory  =
-    getFiles(directory, "*.config", SearchOption.AllDirectories)
-    |> Seq.tryFind isAppOrWebConfig
-   
-
 let private baseConfig = """<?xml version="1.0" encoding="utf-8"?>
 <configuration>
 </configuration>
@@ -181,8 +170,8 @@ let applyBindingRedirectsToFolder isFirstGroup createNewBindingFiles cleanBindin
     let applyBindingRedirects projectFile =
         let bindingRedirects = bindingRedirects projectFile |> Seq.toList
         let path = Path.GetDirectoryName projectFile.FileName
-        match getConfig Directory.GetFiles path with
-        | Some c -> Some c
+        match projectFile.TryFindConfigFile() with
+        | Some c -> Some c.FullName
         | None -> 
             match createNewBindingFiles, List.isEmpty bindingRedirects with
             | true, false ->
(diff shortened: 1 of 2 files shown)


src/Paket.Core/PaketConfigFiles/ProjectFile.fs
```diff
@@ -1934,6 +1934,10 @@ type ProjectFile with
             search true (DirectoryInfo folder)
 
         findAllFiles folder
+
+    /// Finds all project files
+    static member FindAllProjectFiles folder : FileInfo [] =
+        ProjectFile.SearchAllProjectRelatedFiles "*proj*" folder
```
forki, line 1940: that's not the same pattern as we did above, right? This would include "MyprojectFile.fs" or stuff like that

src/Paket.Core/Installation/BindingRedirects.fs
```diff
@@ -83,17 +83,6 @@ let internal indentAssemblyBindings config =
         if not parent.HasElements then
             parent.Remove()
 
-let private configFiles = [ "app"; "web" ] |> Set.ofList
-let private projectFiles = [ ".csproj"; ".vbproj"; ".fsproj"; ".wixproj"; ".nproj"; ".vcxproj"; ".pyproj"; ".sfproj"; ".sqlproj" ] |> Set.ofList
```
forki, line 87: I think we need to keep this ugly explicit list

olivercoad: What for? It's private, not used anywhere, and is just a duplicating the list in ProjectFile.fs

src/Paket.Core/PaketConfigFiles/ProjectFile.fs
olivercoad: This is the same as what it was originally (see L1902). Yes it would, but it gets filtered out anyway in the next line.

Do you think I should change it to `"*.*proj"`?

src/Paket.Core/Installation/BindingRedirects.fs
forki: wasn't this was filtered things to  "*.*proj"?

olivercoad: No, not since https://github.com/fsprojects/Paket/commit/aac9770c6bc5e11261703685a5f15a6fec937b21

olivercoad:
Thanks for the quick merge and release!

Issue #4157, opened 2022-06-15
Paket pack cannot ignore framework subdirectory


legrab:
### Description

Running paket pack on an SDK style project that builds without the _TargetFramework_ subdirectory will not find the assembly.

Using `AppendTargetFrameworkToOutputPath` set to `false` or specifying `OutDir` instead of `OutputPath` results in building the assemblies into the top `${OutputPath}\` instead of the framework subdirectory as `${OutputPath}\${TargetFramework}\`.

Either setting is ignored by paket and it tries looking up the assembly in the `${OutputPath}\${TargetFramework}\` folder where it is not.

### Repro steps

1. Create a project with any of the two csproj properties (set `AppendTargetFrameworkToOutputPath` to `false` or specify the `OutDir`).

2. Try running paket pack on the project.

### Expected behavior

Paket finds the assembly in the directory as per where it is configured to build.

### Actual behavior

Paked looks for the assembly in the directory the assemlby would be if neither of the two property would be configured.

### Known workarounds

Dotnet pack can be used instead on the project but that leads to a lot of extra effort and new issues when other projects are packed with paket in the same solution.

### Duplication

Please note that a narrower range of this issue (mentioning only the `AppendTargetFrameworkToOutputPath` property) has been raised 2 years ago on issue #3971.

Flohack74:
@forki could you take a look on this, thats really a showstopper for Paket and net 6.0 for us, since our CI pipeline requires to put all output assemblies into the same out folder.

forki:
could you please prepare the repro so that I just have to run that paket pack call? Sorry I'm very low on time at the moment, but I would then try to look at it this evening

Koala06:
this should do it: https://github.com/Koala06/paket_4157

As per my understanding of the issue, it is in L1642 where the only place where the referred 'AppendTargetFrameworkToOutputPath' property is handled, but that works with the constant 'bin' pre-directory, so that clearly is not the generic case when the BuildOutputFolder is different.
This leads us to this line:
https://github.com/fsprojects/Paket/blob/80d1c2cd598d6e891a958801533ae7fc6a3a6309/src/Paket.Core/PaketConfigFiles/ProjectFile.fs#L1659
Here we can see that 'framework' sub-dir is added to the path disregarding the property.

Please also note that OutDir is also problematic as leading to the same error.

legrab:
The suggested Pull Requests have been merges, so this issue can be closed

PR #4159, opened 2022-06-22
Do not append TargetFramework subdirectory to OutputPath

Koala06:
when AppendTargetFrameworkToOutputPath property is false

The change has not been tested

This commit is meant resolve issue  #4157


Koala06:
I have tested the 7.2.0-alpha001 with and without the AppendTargetFrameworkToOutputPath on this sample repo: https://github.com/Koala06/paket_4157

Both cases work as expected; the package is created whether the assembly is built into the subdir or not.

Koala06:
I will add a test case to check the newly supported configuration to OutputSpecs.fs, hopefully in a few days.

legrab:
I have created an additional PR for testing the change: https://github.com/fsprojects/Paket/pull/4162

legrab:
@forki Is there any chance that we could have a proper release containing the changes? Or a more extensive test may be required?

PR #4163, opened 2022-06-26
un-ignore .paket/ directory

jcmrva:
Per [recommendation](https://github.com/fsprojects/Paket/issues/2840#issuecomment-488618766)


forki:
I don't think this applies here.

Issue #4164, opened 2022-07-06
Dependency manager looks up older tool version of paket


aggieben:
### Description

When I run FSI using the Dependency Manager extension for paket, it resolves a path to `paket.exe` through an older version of a dotnet CLI tool package.

### Repro steps

Please provide the steps required to reproduce the problem

1. Launch FSI with `dotnet fsi --compilertool:"D:\Users\ben\proj\mbrace\MBrace.Core\packages\fsi\FSharp.DependencyManager.Paket\lib\netstandard2.0\"`

2. Evaluate this: `#r "packet: nuget FSharp.Data" // or any package name, this is just the one I used.

### Expected behavior

The dependency manager resolves to latest available version of the paket cli tool and successfully resolves requested package.

### Actual behavior

```
λ dotnet fsi --compilertool:"D:\Users\ben\proj\mbrace\MBrace.Core\packages\fsi\FSharp.DependencyManager.Paket\lib\netstandard2.0\"

Microsoft (R) F# Interactive version 12.0.4.0 for F# 6.0
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> #r "paket: nuget FSharp.Data";;
:paket> using C:\Users\Ben\.nuget\packages\paket\6.0.0-beta2\tools\paket.exe
:paket> Paket version 6.0.0-beta2
:paket> Total time taken: 0 milliseconds
exception while resolving dependencies: System.Exception: Package resolution using 'C:\Users\Ben\.nuget\packages\paket\6.0.0-beta2\tools\paket.exe' failed, see directory 'C:\Users\Ben\AppData\Local\Temp\script-packages\1601146677'.
Paket failed with
-> Error in paket.dependencies line 1
     Could not parse framework 'net6.0'. Try to update or install again or report a paket bug.
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1448.Invoke(String message)
   at ReferenceLoading.PaketHandler.ResolveDependenciesForLanguage$cont@168(String targetFramework, String scriptDir, IEnumerable`1 prioritizedSearchPaths, String fileType, String workingDir, FileInfo workingDirSpecFile, String rootDir, FSharpList`1 packageManagerTextLines, String loadScript, Unit unitVar) in D:\code\Paket\src\FSharp.DependencyManager.Paket\ReferenceLoading.PaketHandler.fs:line 253
[...]

PR #4168, opened 2022-07-20
fix to make issue #4164 less likely to hit

smoothdeveloper:
related to how extension picks latest paket from $home/.nuget/packages (#4164, cc @aggieben)


smoothdeveloper:
getting the expected outcome in the added test now, and tried the extension locally to check it picks up latest (so long the invariants about version numbering don't change, always 3 numbers, and optional string that is alpha sorted).

forki:
thanks for the contribution

forki:
thanks for the contribution

PR #4174, opened 2022-09-02
Update readme to reflect update to NuGet

bbrandt:
... took them long enough. :-)


bbrandt:
Just changing phrasing from does not to did not so it can remain truthy.

forki:
thx

PR #4175, opened 2022-09-06
Remove recommendation to git-ignore .paket/

mattgallagher92:
It seems that it is not recommended to ignore files it contains. See https://github.com/fsprojects/Paket/issues/2840#issuecomment-488618766.

The inconsistent documentation caused confusion in https://github.com/SAFE-Stack/SAFE-template/pull/526


forki:
thx

Issue #3756, opened 2019-12-17
How do I reference a Git dependency in my project?


njlr:
### Description

I would like to use a Git dependency in my project. The Git repo contains `.fsproj` files, so I would expect referencing to be possible. 

### Repro steps

Please provide the steps required to reproduce the problem

1. Add the Git dependency to `paket.dependencies` at the "solution" level. 

```
git https://github.com/org/proj.git
```

2. Install the packages. 

```bash
dotnet paket install
```

Paket fetches the files and puts them into `paket-files` :+1: 

3. Add the Git package to `paket.references` of my project. 

How?

### Expected behavior

I should be able to reference the Git dependency somehow. 

### Actual behavior

I cannot find a way to reference the Git dependency. The [docs](https://fsprojects.github.io/Paket/git-dependencies.html) do not explain what to do. 

### Known workarounds

Create a NuGet package instead?

kaashyapan:
It does appear that you may only reference git repositories that contain Nuget packages and supply a build command

> If you have NuGet packages inside a git repository you can easily use the repository as a NuGet source from the

```
git https://github.com/forki/nupkgtest.git master build: "build.cmd", Packages: /source/, OS: windows
git https://github.com/forki/nupkgtest.git master build: "build.sh", Packages: /source/, OS: mono

```

Issue #3064, opened 2018-02-22
paket.reference to recognizing github packages in paket.lock file 


gdennie:
### Description

The paket.dependencies file is:
```
source https://www.nuget.org/api/v2

github fsprojects/Argu
```

### Repro steps

1. Create a F# Windows console application using VS 2015/17 Community Edition (OS: Windows 2008 R2)
2. Perform `paket init`
3. Create paket.dependencies file with the following contents:
```
source https://www.nuget.org/api/v2

github fsprojects/Argu
```
4. Create paket.references file in project with contents:
`Argu`
5. Perform `paket update` (or `paket install`) from VS menu or commandline

### Expected behavior

`Argu` would now exist as a reference within the project containing the aforementioned `paket.reference` file.

### Actual behavior

The following error message is display (and is not completely represented in VS Error List when install or update is performed from VS' paket menu):
```
Paket failed with
-> Installation Errors :
    - <...full path to project...>\paket.references uses NuGet package Argu, but it was not found in the paket.lock file in group Main.
```

### Known workarounds

clone the projects and manually link them into your solutions...ignoring paket.

panesofglass:
I think this is related to the GitHub changes. Even after working around the paket download issue, my builds were still breaking on AppVeyor with an error that the file was not found. At least, I think this is related. If not, I'm happy to open a new issue.

forki:
@panesofglass with the new bootstrapper?

panesofglass:
@forki yes.

forki:
@gdennie I looked at this. What makes you think this would work? you are referencing a github project - not a nuget package. Is this a feature request?

gdennie:
@forki I reviewed the documentation again, [GitHub dependencies](https://fsprojects.github.io/Paket/github-dependencies.html). At the time I somehow thought that `paket` allowed the import of `github` hosted libraries similar to how it treated `nuget`. Upon review, that is not the case. The documentation in this area could use some work.

Thanks for your attention.
George.

forki:
Did you look at https://fsprojects.github.io/Paket/git-dependencies.html#Using-Git-repositories-as-NuGet-source - with git dependencies you can let paket build the nupkgs for you.

Also please send improvements to the docs that clarify the problem

gdennie:
Either I did not notice that or did not recognize it as the defacto solution that it is. The fact that github packages must be built is an obvious but understated distinction from nuget that could use a bit more fanfare.

[...]

PR #4190, opened 2022-12-31
Update git-dependencies.md

hesxenon:
with #3756 and #3064 it is quite clear, that the documentation around adding git repositories as a (usable) dependency isn't all that clear. Which is a pity, because the actual process isn't all that hard and also quite sensible-ish.

This change aims to slightly improve the docs and shed some light on this topic because - basically - all that is missing is some basic knowledge about how namespace/package resolution works in fsharp.

### DIFF
--- a/docs/content/git-dependencies.md
+++ b/docs/content/git-dependencies.md
@@ -71,12 +71,14 @@ git https://uri/to/repo.git master build: "powershell build.ps1", OS: windows
 
 ## Using Git repositories as NuGet source
 
+### With pre-built, checked-in nupgks
+
 If you have NuGet packages inside a git repository you can easily use the
 repository as a NuGet source from the [`paket.dependencies`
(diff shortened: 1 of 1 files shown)


docs/content/git-dependencies.md
```diff
@@ -103,4 +117,8 @@ git https://github.com/forki/nupkgtest.git master build: "build.sh", Packages: /
 nuget Argu
 ```
 
-In this sample we have different build scripts for Mono and Windows.
+In this sample we have different build scripts for Mono and Windows. Both of the specified scripts simply cody the prebuilt nupkgs from `/bin` to `/source`
```
smoothdeveloper, line 120: there is a typo: cody -> copy?

hesxenon: absolutely right, thanks for reviewing

PR #4196, opened 2023-02-06
Update the blacklist terminology to be more inclusive

ruhullahshah:
This commit is an attempt to remove the usage of the terminology "blacklist" within this project and rather use an inclusive terminology that is acceptable to a wider audience (devs/users) without running into the danger of hurting sentiments and thereby doing away with colonial-era terminology.

As an example, https://thenewstack.io/words-matter-finally-tech-looks-at-removing-exclusionary-language/


forki:
thank you!

ruhullahshah:
@forki You are welcome. I do not know the reason for the failing tests though. Running tests locally works.

Greifenkralle:
A step into the right direction. 👍

ruhullahshah:
@forki Is there something I can/have to do to make the failing checks pass?

PR #4203, opened 2023-03-03
Proposal: increase process lock trials from 100 to 500 to support dotnet restore on larger solutions

cdrnet:
Otherwise, given how the mechanism works, I don't think a process lock max trial count of 100 could reliably support `dotnet restore` on solutions with more than 100 projects. The downside is potentially longer time until it aborts when there is a left-over lock, which should be rare though.

Alternatively maybe this could be made configurable.

Any thoughts?


cdrnet:
Thank you very much for the release!

PR #4219, opened 2023-07-13
support for dotnet 8

smoothdeveloper:
Like dotnet 8, this is not ready, but it is cooking.

according to https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md older versions of windows are getting dropped, I'll need to adjust some of the tests I copy pasted.

I also believe we should consider a split "pre dotnet 5" and replace most of the stuff we keep duplicating, maybe by time of dotnet 10 would be a good time for this?


smoothdeveloper:
@forki I think this is ready, unless the remaining CI failures would be related to my changes.

Issue #4186, opened 2022-12-08
Default dependencies refer to net5.0 framework


Recktenwald:
### Description

Initializing a new F# project and using paket breaks the project.

### Repro steps
Make sure that .net6 or .net7 are installed.

1. create a new project with `dotnet new console -lang F#`
2. install paket
3. add any dependency, e.g. `dotnet paket add FSharp.Core`
4. `dotnet run`

### Expected behavior

We see the `Hello from F#!` output

### Actual behavior

```
Unhandled exception. System.IO.FileNotFoundException: Could not load file or assembly 'FSharp.Core, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'FSharp.Core, Version=7.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
```

### Known workarounds

Make sure that the `.fsproj` and `paket.dependencies` files agree on the framework.
For example use `framework: auto-detect` in the `paket.dependencies` file.

I would suggest changing the default template that is crated with `paket init` to `auto-detect`. This seems to me to be the easiest, and most beginner-friendly way.

I located, what I think, are the relevant lines of code. In the `Environment.fs` file at the very bottom 
```fs
let init (directory : DirectoryInfo) =
        let sources = [PackageSources.DefaultNuGetV3Source]
        let additionalLines = [
            "storage: none"
            "framework: net5.0, netstandard2.0, netstandard2.1"
        ]
        initWithContent sources additionalLines directory
```
But maybe there is a good reason, which I don't know, for not using `auto-detect` as the default.

ChrSteinert:
I guess this can be closed with the latest version?

PR #4228, opened 2023-09-25
paket init: Default framework restriction

ChrSteinert:
In response to #4186 , changes the default framework restriction in the paket.dependencies file when running paket init to be `auto-detect, and adds a newline at the end of the file.

While it is obviously maintainers choice to take or leave this, I also quite often "stumble" over the `net5` default restriction.
I would personally might pick something like the latest LTS, but `auto-detect` is the least … "getting in the way of the user".

Any advice/optinion is appreciated!

### DIFF
--- a/src/Paket.Core/PackageAnalysis/Environment.fs
+++ b/src/Paket.Core/PackageAnalysis/Environment.fs
@@ -118,6 +118,7 @@ module PaketEnv =
         let sources = [PackageSources.DefaultNuGetV3Source]
         let additionalLines = [
             "storage: none"
-            "framework: net5.0, netstandard2.0, netstandard2.1"
+            "framework: auto-detect"
+            ""
         ]
         initWithContent sources additionalLines directory
--- a/src/Paket.Core/PackageAnalysis/Environment.fs
+++ b/src/Paket.Core/PackageAnalysis/Environment.fs
@@ -118,7 +118,6 @@ module PaketEnv =
         let sources = [PackageSources.DefaultNuGetV3Source]
         let additionalLines = [
             "storage: none"
-            "framework: auto-detect"
             ""
         ]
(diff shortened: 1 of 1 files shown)


smoothdeveloper:
@ChrSteinert from my subjective experience, leaving the line totally out of the paket.dependencies gives me the best experience, and advanced users can "pin" it, if they don't like the behaviour.

I think having it omitted doesn't impact the .fsproj, causes no churn in terms of version control of those files, I assume it may make the paket install take a bit longer.

My suggestion unless we get better feedback would be to remove the line from the template / `paket init`.

ChrSteinert:
Huh – I was not aware, that omitting it is an option.

I found the default being a good pointer to „you can configure this that way“, without necessarily ready the documentation.

What do you think about adding a framework restriction as a comment, like many other default config files do?
Von: Gauthier Segay ***@***.***>
Gesendet: Monday, September 25, 2023 5:15:09 PM
An: fsprojects/Paket ***@***.***>
Cc: Christian Steinert ***@***.***>; Mention ***@***.***>
Betreff: Re: [fsprojects/Paket] paket init: Default framework restriction (PR #4228)


@ChrSteinert<https://github.com/ChrSteinert> from my subjective experience, leaving the line totally out of the paket.dependencies gives me the best experience, and advanced users can "pin" it, if they don't like the behaviour.

I think having it omitted doesn't impact the .fsproj, causes no churn in terms of version control of those files, I assume it may make the paket install take a bit longer.

My suggestion unless we get better feedback would be to remove the line from the template / paket init.

ChrSteinert:
Did I just miss it, or is the (top level/group level) framework restriction not even mentioned in the docs?

ChrSteinert:
Never mind… I just cant read…

Issue #4241, opened 2024-01-05
Visual studio not restoring packages for projects that target net8.0-windows


DanJensen:
### Description

I'm working on upgrading a net6 product suite (which uses paket) to net8.  Some of the projects are Windows UI applications, and as such they target net6.0-windows.  Everything works fine targeting net6.  However, when upgrading to net8, something is causing VS 2022 to no longer restore any packages in projects that target net8.0-windows.  The net8.0 projects work fine.  And ALL projects work if I run a build from the command line using dotnet.exe.  So the problem is specific to VS when combined with net8.0-windows projects.  

### Repro steps

I have multiple projects that target net8.0, and multiple that target net8.0-windows.  The behavior described above is consistent in all of them, and it doesn't matter which nuget packages I'm referencing.  In VS, net8.0 projects work, but net8.0-windows projects do not.  And everything works from the command line.  I should also add that "dotnet paket restore" does bring everything down into the packages folder as expected.

Just to be clear, my paket.dependencies file does have groups defined.  One has a restriction of net8.0 and one has a restriction of net8.0-windows.  I know I'm using the correct groups in the various paket.references files because: 1) The same groups worked fine for net6; and 2) It works fine for net8 if I build it from the command line with dotnet.exe.

Also, I'm using paket v8.0.0.  I'm using that same paket version with the net6 codebase and it's fine.

### Expected behavior

Packages should be restored to the projects in VS Solution Explorer and builds should work within VS.

### Actual behavior

There is no "Packages" node listed under the Dependencies section for the net8.0-windows projects in VS Solution Explorer.  And of course, builds fail because it doesn't recognize the referenced objects.

### Known workarounds

I'd love to find one.  :(

PR #4242, opened 2024-01-12
net8.0-windows target issue

Jooseppi12:
This PR is for addressing the issue in #4241 which was caused by mismatching conditions in the `paket.props` file in the `obj` folder of a project.

When net8.0 support was added the `net8.0-windows` and os specific `net8.0` targets were left on TargetFrameworkVersion `v7.0` accidentally, which caused mismatch in the above mentioned file.

Also `net8.0-windows` seems to be defaulting to `net8.0-windows7.0` instead of `net8.0-windows10.0.17763.0`, therefore the TargetPlatformVersion is also mismatching, so the `PackageReference` is not added during `DesignTime`

Also attached a simple project that reproduces the problem.

[repro4241.zip](https://github.com/fsprojects/Paket/files/13917774/repro4241.zip)


forki:
thanks! released in 8.0.3

PR #4244, opened 2024-01-31
Aliases should not be inherited.

vha-schleupen:
If aliases are configured, these should not be set for transitive dependencies.


forki:
thanks. released in 8.1.0-alpha001 - but unfortunately I can't sign the assemblies right now. Maybe the cert ist expired. not sure

PR #4249, opened 2024-03-05
process dependencytree from top to bottom

vha-schleupen:
The dependency tree should be processed from top to bottom. Up to now, the dependencies are processed in no particular order.
If package 'A' is both directly and transitively dependent and an alias is configured for it, it can happen that the transitive dependency is processed first and therefore the alias setting is not taken into account. 

The direct dependencies should be processed first.


vha-schleupen:
> I think it's important to make sure that we do not add doublicates to the list. Could you please add some kind of "distinct" function

fixed. @forki

PR #4253, opened 2024-03-25
use nuget CPM to enable strict package reference on direct dependencies

antymon4o:
This PR fixes issue [references: strict either does not work, or does not work as expected](https://github.com/fsprojects/Paket/issues/2257#top)
#2257

In Paket.Restore.targets based on ManagePackageVersionsCentrally property:
 - add condition on Package reference to include direct only packages
 - add condition on Version element to be used only when ManagePackageVersionsCentrally not is enabled
 - add package version items for all the packages in ref file

In RestoreProcess.fs - the changes are similiar to those in Paket.Restore.targets


antymon4o:
@forki @agross
Is it possible for you to have a loot at this PR and tell you opinion on the approach?
Thank you very much! 😊

antymon4o:
👋 @forki, can I draw your attention to this PR, please? Or if someone else can review this PR, would you forward it or tag him/her? I would greatly appreciate any attention. Thanks again.

forki:
Thanks

PR #4271, opened 2024-09-03
Allow sources to be specified as Environment Variables

### LINKED ISSUE #1796: Variables in paket.dependencies
It would be useful to allow some kind of variables or references in `paket.dependencies`.

Something like:

```
source ...

variable mylib.version >= 1.3.0

nuget mylib.dll1 ${mylib.version}
nuget mylib.dll2 ${mylib.version}
```

This would allow defining a common version (range) for a set of co-released libs (i.e. part of one solution).

msarilar:
resolves #1796


forki:
this should be mentioned in the docs

dsyme:
/repo-assist Please make a new PR with the same contents as this PR but also adding docs.

Issue #4226, opened 2023-09-23
References to delisted versions of Nuget.Packaging


Numpsy:
### Description

An observation whilst looking at https://github.com/fsprojects/FAKE/issues/2760 -

Both the released version of Paket and the v8 alphas have references to Nuget.Packaging 5.9.1, but that version has been delisted from nuget.org:

![image](https://github.com/fsprojects/Paket/assets/1178570/ef9eda37-97e4-4055-b38b-62619b2b5eb8)

There are also multiple versions in this version range that are listed with security issues, so - perhaps the minimum version should be updated to a fixed version?

PR #4275, opened 2024-11-12
Update the NuGet.* packages from 5.9.1 to 5.11.6

Numpsy:
refs https://github.com/fsprojects/Paket/issues/4226

The currently referenced 5.9.1 version has been delisted from nuget.org as well as being marked as suffering from security vulnerabilities.

5.11.6 is the only 5.x version which is both listed and which has no known security issues, so attempt to update to that version.


forki:
thanks!

Issue #4278, opened 2024-11-18
Using paket restore on project with targetframework net9.0-windows fails


davikor:
### Description

When I try to run `paket restore` on a project with TFM `net9.0-windows` the command fails with error message `Paket failed with -> MatchFailureException: The match cases were incomplete`

### Repro steps

```
dotnet new wpf -n PaketTest -o PaketTest
cd PaketTest
//check if tfm in PaketTest.csproj is net9.0-windows
dotnet tool install paket --create-manifest-if-needed
dotnet paket add CsvHelper
dotnet paket restore -f -v
```

### Expected behavior

`paket restore` should finish without error.

### Remarks

The stacktrace references the exception to occur here in [FrameworkHandling.fs](https://github.com/fsprojects/Paket/blob/master/src/Paket.Core/Versioning/FrameworkHandling.fs#L887). [Here](https://github.com/fsprojects/Paket/blob/906529e0b239e8dfaef5ad3359ea343c3f225da6/src/Paket.Core/Versioning/FrameworkHandling.fs#L989) the switch has a comment mentioning, that "for now windows version of 9.0 is alias to 8.0". Maybe this is related to this issue?

smoothdeveloper:
@davikor thanks, I'll look into reproducing / adding a test case and fix the code you are pointing out, it indeed requires more cases there despite windows version supported are intended to be the same for now.

smoothdeveloper:
@davikor, can you confirm you are all set with latest release?

davikor:
@smoothdeveloper thanks a lot. 9.0.2 fixes this issue for me.

PR #4279, opened 2024-11-18
Fix for issue #4278

smoothdeveloper:
There was a warning on this incomplete match expression.

Will add coverage and probably other tweaks related to dotnet sdk in the solution in a separate PR.

With those changes, I tried manually running the binary on steps similar to what @davikor reported in #4278 and paket didn't fail.

A new prerelease version can be pushed with the fix.


forki:
Thanks!

Jamesdogma (approved):

PR #4284, opened 2025-01-09
Apply framework restriction

fpellet:
As the `apply framework restriction` test shows, the framework restriction is not applied. This downloads a version that only supports .net 9 when I request the .net 8 framework.

My solution works, except in force mode.
I have the impression that I have to download the package for it to work.
I can do it, unless you have something better.


forki:
unfortunately this not working properly and user report errors like https://github.com/fsprojects/Paket/issues/4316 - I habe to revert that for now.

PR #4287, opened 2025-01-27
Fix script generation if package contains FSharp.Core name

fpellet:
It's normal to ignore FSharp.Core, but not Company.FSharp.Core, CustomFSharp.Core ou CustomFSharp.CoreCustom


smoothdeveloper:
Seems like overlook when initially implementing load script generation, thanks for adjusting the code!

smoothdeveloper:
Maybe we should make sure FSharp.Coresomething & FSharp.Core.something also behaves as your change intends to adjust?

fpellet:
Done 🙂

PR #4281, opened 2024-12-07
add a test for #4278 and adjustments for enabling warning as errors

smoothdeveloper:
Giving it a try on CI, locally, for the added test, I face some MSbuild nonsense:

> C:\dev\src\github\fsprojects\Paket\integrationtests\scenarios\i004278-net90-windows-fails\before\before.sln.metaproj : error MSB4126: The specified solution configuration "Debug|x64" is invalid. Please specify a valid solution configuration using the Configuration and Platform properties (e.g. MSBuild.exe Solution.sln /p:Configuration=Debug /p:Platform="Any CPU") or leave those properties blank to use the default solution configuration. [C:\dev\src\github\fsprojects\Paket\integrationtests\scenarios\i004278-net90-windows-fails\before\before.sln]

I did minor tweaks to have warning as errors, the change in [src/Paket.Core/Dependencies/DependenciesFileParser.fs](https://github.com/fsprojects/Paket/compare/master...smoothdeveloper:fix-4278?expand=1#diff-72dc78df0df0be9d06b144a35c5e09e2ba716cf53f5e340fe74254889781dc1a) is worth some attention, but the rest is straight forward.

PR #4290, opened 2025-01-30
Fixed a Few Compiler Warnings

1eyewonder:
Here are a few fixes to some compiler warnings we are getting on builds. I'm wanting to try and go through and attempt to do a few rounds of cleanup to make real issues more obvious when we encounter them.

I'll be making separate PRs for:
1. InstallModel.CreateFromLibs -> InstallModel.CreateFromContent
  1. There are a lot of warnings we can clean up here
2. build.fsx
  1. I think we can eliminate a lot of these warnings by updating our FAKE dependencies and referencing the modern core namespaces instead


smoothdeveloper:
Can you base your PRs on top of #4281?

It already enables warning as errors, while leaving those that are a bit invasive to fix in the codebase.

PR #4292, opened 2025-03-04
Add "generate_path_property" option

sin-ack:
This replicates the behavior of `<PackageReference GeneratePathProperty="true" />` in Paket. This is required in certain edge cases, such as when creating a source generator project that needs to reference third-party assemblies.

Tests have been added for verifying the option in the lockfile, paket.dependencies/paket.references files and to verify the output in the generated `paket.props` files.

### DIFF
--- a/src/Paket.Core/Dependencies/DependenciesFileParser.fs
+++ b/src/Paket.Core/Dependencies/DependenciesFileParser.fs
@@ -216,6 +216,7 @@ module DependenciesFileParser =
     | SpecificVersion of bool
     | CopyContentToOutputDir of CopyToOutputDirectorySettings
     | GenerateLoadScripts of bool option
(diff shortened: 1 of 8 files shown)


sin-ack:
Hmm, I learned after opening this that apparently it is possible to enable this without Paket integration using [ItemDefinitionGroup](https://learn.microsoft.com/en-us/visualstudio/msbuild/item-definitions?view=vs-2022):

```xml
<ItemDefinitionGroup>
  <PackageReference>
    <GeneratePathProperty>true</GeneratePathProperty>
  </PackageReference>
</ItemDefinitionGroup>
```

If this PR is unnecessary after this, feel free to close it.

dsyme:
Closing per comment

Issue #4296, opened 2025-05-21
.net 10 support


Larocceau:
.net 10 is coming out later this year, so it would be good to get started on support on it. I would be interested in having a go at it, but that would probably be in a few weeks; given the [dev guide](https://github.com/fsprojects/Paket/blob/master/DEV_GUIDE.md#updates-to-support-new-major-sdk-releases) it should not be too hard

As for now the issue on supporting .net 9 is still open #4237, is there actually anything that needs to be done there still?

smoothdeveloper:
Thanks @Larocceau, I closed the .net 9 issue as the PR for support was merged.

If you want to give a shot to extend for .net 10 support, feel free to make a PR and reach out for assistance.

Based on previous experience, minor adjustments to few files, and adding a simple test case is good enough for things to work.

Numpsy:
Something like https://github.com/fsprojects/Paket/compare/master...Numpsy:Paket:net10 ?

(Doesn't quite work at the moment because the ```tryNormalizeVersion``` function turns the '10' in 'net10-windows' into '1.0' and fails the tests)

granicz:
Hi @smoothdeveloper , any plans on fixing+merging the above PR?

forki:
It does already in alpha channel. We just need a couple of more people to
confirm that everything is OK.

Casper ***@***.***> schrieb am Sa., 15. Nov. 2025, 18:36:

yaakov-h:
I've tried 10.0.0-alpha011 and there is some weird interaction going on with test projects and paket.

With NuGet, if I add my test packages (nunit3, nunit3testadapter, microsoft.net.test.sdk) I get a project.deps.json that includes microsoft.net.test.sdk as a dependent package, same as on previous runtimes.

With Paket, that is missing from my project.deps.json, so our in-house test environment isn't detecting it as a test project. It still works with VS IDE and dotnet-test CLI. There's nothing obvious in MSBuild XML or binlogs as to how this reference is getting omitted during the build.

This is after needing to set `<RestoreEnablePackagePruning>false</RestoreEnablePackagePruning>` to disable the pruning as per above, though I wonder if the test package is somehow getting pruned regardless.

Helevn:
Paket Restore Fails with Exit Code 1 After Successful Manual Restore
Problem
MSBuild Paket restore target fails with exit code 1 even when dotnet paket restore succeeds manually.

Steps to Reproduce
Create .NET project using Paket

[...]

PR #4297, opened 2025-06-04
Try adding some .NET 10 support

Numpsy:
refs https://github.com/fsprojects/Paket/issues/4296

Basically cloned from the changes to add .NET 9 support, as I'm learning this as I go along.

As it stands the new tests don't quite work because the ```tryNormalizeVersion``` function is parsing the ```10``` in ```net10-windows``` into ```1.0``` instead of ```10.0```  - I haven't had time to look at that further.

### DIFF
--- a/src/Paket.Core/Versioning/FrameworkHandling.fs
+++ b/src/Paket.Core/Versioning/FrameworkHandling.fs
@@ -152,6 +152,7 @@ type Net8WindowsVersion =
         | _ -> None
 
 type Net9WindowsVersion = Net8WindowsVersion
+type Net10WindowsVersion = Net9WindowsVersion
 
 [<RequireQualifiedAccess>]
 type Net5Os =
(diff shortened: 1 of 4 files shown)


Numpsy:
> parsing the 10 in net10-windows into 1.0

Saying that, the [MS Docs](https://learn.microsoft.com/en-us/dotnet/standard/frameworks) list ```net11``` as being .NET Framework 1.1, so I suppose you can't try treating 10/11 as 10.0/11.o in that case

Jand42:
Hi @Numpsy, @gdziadkiewicz, see my commit https://github.com/Jand42/Paket/commit/76e0c027fa4b18354c8eaa66a6637c8fb22381de on top of this PR which makes parses `".netXX.0"` into version `"XX.0"`, also handling OS specific versions, while keeping UAP parsing intact. All tests are passing.

Issue #4209, opened 2023-04-20
Paket does not use global nuget cache directory set in environment variable


milbrandt:
### Description

Using nuget config it's possible to select an alternative nuget cache location with the environment variable `NUGET_PACKAGES`, see global packages at https://learn.microsoft.com/en-us/nuget/consume-packages/managing-the-global-packages-and-cache-folders.

### Repro steps

Useing Paket version 7.2.0+bb14ab674b2748070a624f394cd6796e11aaffa8

Please provide the steps required to reproduce the problem

1. Clear current NuGet caches `dotnet nuget locals all --clear`

2. Move Nuget Cache to another drive, e.g. set `NUGET_PACKAGES`to `E:\nuget` in the Windows system environment variables.

3. Restore a solution using Paket

4. Build will fail, in the folder `%USERPROFILE%\AppData\Local\NuGet\Cache\` projects are restored and not all files are in the corresponding subfolders, the environment variable is not respected

If possible then please create a git repository with a repro sample or attach a zip to the issue.

May be releated with https://github.com/fsprojects/Paket/issues/3853

### Expected behavior

Please provide a description of the behavior you expect.

### Actual behavior

Please provide a description of the actual behavior you observe. 

### Known workarounds

Don't use the environment variable :(, replace the directory `%USERPROFILE%\AppData\Local\NuGet` with a link to another drive.

Issue #3853, opened 2020-05-25
Paket does not use global nuget cache directory set in configuration file


audun:
### Description

Using nuget config it's possible to select an alternative nuget cache location: https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior

However, Paket does not properly use the configured value, causing some strange and confusing error messages, for example:

`Package C:\src\project\packages\Microsoft.AspNet.Razor\microsoft.aspnet.razor.4.0.0-beta7.nupkg couldn't be extracted to "C:\src\project\packages\Microsoft.AspNet.Razor". System.IO.InvalidDataException: End of Central Directory record could not be found.. Trying to extract files individually.`

### Repro steps

1. Set an alternate cache directory: `nuget config -set globalPackagesFolder=e:\nuget\packages`

2. Verify the configuration: `nuget config globalPackagesFolder`, should return `e:\nuget\packages`

3. Use paket restore

### Expected behavior

Downloaded packages should be put in `e:\nuget\packages`

### Actual behavior

Downloaded packages end up in `C:\Users\user\.nuget\packages`

### Known workarounds

Setting the environment variable `set NUGET_PACKAGES=e:\nuget\packages` works.

audun:
A bit more complete error log:

```
  ==> Trying again
Downloading Microsoft.AspNet.Razor 4.0.0-beta7
Not moving as 'C:\Users\user\.nuget\packages\microsoft.aspnet.razor\4.0.0-beta7\microsoft.aspnet.razor.4.0.0-beta7.nupkg' already exists (maybe some other instance downloaded it as well)
Download of Microsoft.AspNet.Razor 4.0.0-beta7 done in 3 seconds. (878 kbit/s, 0 MB)
Package C:\src\project\packages\Microsoft.AspNet.Razor\microsoft.aspnet.razor.4.0.0-beta7.nupkg couldn't be extracted to "C:\src\project\packages\Microsoft.AspNet.Razor". System.IO.InvalidDataException: End of Central Directory record could not be found.. Trying to extract files individually.
Something went wrong while downloading Microsoft.AspNet.Razor 4.0.0-beta7
Message: Error during extraction of C:\src\project\packages\Microsoft.AspNet.Razor\microsoft.aspnet.razor.4.0.0-beta7.nupkg.
```

PR #4301, opened 2025-06-25
Use NuGet.Configuration to get the path to the NuGet global…

Numpsy:
… packages folder

refs
https://github.com/fsprojects/Paket/issues/3853
https://github.com/fsprojects/Paket/issues/4209

I don't know the history of things in this area, I just hit https://github.com/fsprojects/Paket/issues/3853 when building some things that use paket on a system that has the ```globalPackagesFolder``` setting set to a custom location (on a different drive than the default user profile directory) and wondered if it would be possible to fix it by using the 'where is the cache folder' functionality in ```NuGet.Configuration``` rather than doing it locally (it contains the logic to look for  ```NUGET_PACKAGES```, then ```globalPackagesFolder```, then the regular defaults)

### DIFF
--- a/src/Paket.Core/Common/Constants.fs
+++ b/src/Paket.Core/Common/Constants.fs
@@ -114,12 +114,9 @@ let GitRepoCacheFolder = Path.Combine(LocalRootForTempData,".paket","git","db")
 let [<Literal>] GlobalPackagesFolderEnvironmentKey = "NUGET_PACKAGES"
 
 let UserNuGetPackagesFolder =
-    getEnVar GlobalPackagesFolderEnvironmentKey
-    |> Option.map (fun path ->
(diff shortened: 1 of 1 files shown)


src/Paket.Core/Common/Constants.fs
```diff
@@ -114,12 +114,9 @@ let GitRepoCacheFolder = Path.Combine(LocalRootForTempData,".paket","git","db")
 let [<Literal>] GlobalPackagesFolderEnvironmentKey = "NUGET_PACKAGES"
```
Numpsy, line 114: This gets referenced from https://github.com/fsprojects/Paket/blob/eed370a1ac41eab5096faae8b0621777c816b9e5/integrationtests/Paket.IntegrationTests/NugetDownloadSpecs.fs#L24 so I suppose removing that would be a breaking change

forki:
thanks! this will go into 10.0.0-alpha006

PR #4305, opened 2025-11-11
Version parsing fixes for .NET 10

Jand42:
Versioning fixes to keep `netXX.0` intact to not get parser confused between .NET 10 and 1.0.
The `tryNormalizeVersion` function now takes `isNet` argument to decide to keep `XX.0` intact, so UAP is not affected.
Simplified pattern matching helper for `"net..."`.

All unit tests passing.


forki:
thanks. this is released in 10.0.0-alpha004

PR #4314, opened 2026-01-16
reduce memory allocations in Requirements.simplify

chinwobble:
Context:
I am converting a project from nuget to paket and there are many conflicts. I noticed that the `dotnet paket install` command can take several minutes just to tell me there is a conflict and paket cannot generate a lock file.

I have used https://github.com/microsoft/perfview to find where the bottlenecks.

I have identified a few bottlenecks:
- my paket.dependencies has 86 lines in it and it is allocating ~2.7 GB of memory
- the api calls to nuget are quite slow. 300ms just to return the list versions of available for each package. Based on my paket.dependencies file, there are 563 api calls nuget.org. These are async and run concurrently but In some cases there are large gaps between api calls of 200ms+. 

I have run perf view with the following settings
```cmd
build.cmd publish 

PerfView.exe  "/DataFile:PerfViewData-baseline.etl" /BufferSizeMB:256 /StackCompression /KernelEvents:VirtualAlloc,VAMap,Default /Process:"paket" /ClrEvents:GC,Binder,Security,AppDomainResourceManagement,Contention,Exception,Threading,JITSymbols,Type,GCHeapSurvivalAndMovement,GCHeapAndTypeNames,Stack,ThreadTransfer,Codesymbols,Compilation /NoGui /NoNGenRundown /DotNetAllocSampled run C:\dev\paket\bin\net8\paket.exe install
```
Equivalent to the below settings:

image

Using the tool `GC Heap Alloc Ignore Free tool` and setting include pattern to be:

```
Process% paket (29684);paket!<StartupCode$paket>.$Paket.Program.main@();paket!Paket.Program.main()
```
By default events are captured machine wide, we are only interested in memory allocation events that from stack traces within the `paket` process' main method. `perfview` captures diagnostic system wide.

Drilling into the call tree shows that `Requirements.Simplify'` allocates 13% of the memory of the lifetime of `paket install`.  I use the `.NET SampAlloc` which samples allocations as I ran out of disk space running with ` .NET Alloc CheckBox`. It consumed 70 GB+.

image

## GC Stats
After collecting the trace before and after the code change the stats have changed accordingly:

  | Before | After
-- | -- | --
Total CPU Time: | 16.841 ms | 15.399 ms
Total GC CPU Time: | 732 ms | 653 ms
Total Allocs : | 2,759.402 MB | 2222.850 MB
Total GC Pause: | 765 ms | 653 ms
% Time paused for Garbage Collection: | 1.40% | 2.00%
% CPU Time spent Garbage Collecting: | 4.30% | 4.20%

### DIFF
--- a/tests/Paket.Tests/Versioning/FrameworkRestrictionTests.fs
+++ b/tests/Paket.Tests/Versioning/FrameworkRestrictionTests.fs
@@ -333,3 +333,14 @@ let ``combineOr needs to consider partly disjunct sets``() =
                         FrameworkRestriction.AtLeast (DotNetStandard DotNetStandardVersion.V1_3)])
 
 
+[<Test>]
+[<TestCase("|| (&& (>= net45) (< netstandard2.0)) (&& (< net45) (>= netstandard2.0) (< netstandard2.1)) (>= net47) (&& (< net5.0) (>= netstandard2.1))")>]
+[<TestCase("|| (&& (>= net45) (< netstandard2.0)) (&& (< net45) (>= netstandard2.0) (< netstandard2.1)) (>= net47)")>]
+[<TestCase("&& (>= net45) (< net47) (< netstandard2.0)")>]
+[<TestCase("|| (&& (< net45) (>= netstandard2.0) (< netstandard2.1)) (>= net47)")>]
+[<TestCase("|| (&& (< net20) (>= netstandard1.0) (< netstandard1.3)) (&& (< net20) (>= netstandard1.3) (< netstandard2.0))")>]
+let ``parseRestrictions returns correct result`` frameworkRestrictionString =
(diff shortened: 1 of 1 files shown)


src/Paket.Core/Versioning/Requirements.fs
```diff
@@ -57,7 +57,7 @@ type FrameworkRestrictionP =
     // NOTE: All critical paths test only if this set is empty, so we use lazy seq here
     member x.RepresentedFrameworks =
         match x with
-        | FrameworkRestrictionP.ExactlyP r -> [ r ] |> Set.ofList
+        | FrameworkRestrictionP.ExactlyP r -> Set.singleton r
```
chinwobble: this line reduces allocations by 72 bytes per operation.

PR #4337, opened 2026-07-23
Pull in upstream

dAdil:
Brings in changes from upstream's master. 

Also, since the last release, 1 transient dependency has been marked as having a low severity vulnerability. I've marked low-sev vulnerabilities as "warnings not as errors". This is consistent with the medium and high levels which are already excluded. This would probably be simple enough to get merged upstream, but it got me unblocked for now.


dAdil:
Apologies, this was meant to go in a fork!

PR #4415, opened 2026-08-30
build: upgrade to .NET 10

fpellet:
Upgrades the build to the .NET 10 SDK and retargets the produced binaries from `net9` to `net10.0`, following the same shape as the .NET 9 upgrade in d3b6d43c8 that `DEV_GUIDE.md` points to as the reference procedure.

## Prerequisite: the local paket tool

`.config/dotnet-tools.json` pinned paket **9.0.2**, which predates net10.0 support (added in 10.0.1). Since `build.sh` runs `dotnet paket restore` with that tool, it had to be bumped to 10.3.1 before the projects could retarget. The tool regenerates `.paket/Paket.Restore.targets`, which picks up the group conditions plumbing from 10.1.1 — the file now depends on the `show-conditions` command and so requires paket >= 10.1.1.

Note this file still lags `src/Paket.Core/embedded/Paket.Restore.targets`, which carries unreleased fixes. That divergence is pre-existing and inherent to self-hosting: any restore with a released tool rewrites the file with the released content.

## Two regressions surfaced by the SDK

`TreatWarningsAsErrors=true` turns every new diagnostic into a build failure.

- **NU1510** (x35): NuGet 10 flags framework-provided `PackageReference`s as prunable. They come from paket's transitive resolution of `NETStandard.Library`, not from hand-written references, so they cannot simply be removed — added to `WarningsNotAsErrors` alongside the existing exemptions.
- **FS0049** in `SemVer.fs`: the F# 10 compiler rejects the `Int M` pattern binding. Renamed to `ma`; behaviour unchanged.

## Impact on consumers

The `paket` .NET tool moves to `net10.0`, so its minimum runtime goes from .NET 9 to .NET 10. The ILRepack-merged `paket.exe` still targets `net461`, so the bootstrapper / .NET Framework path is unaffected. Released here as `11.0.0-alpha001`.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "paket": {
-      "version": "9.0.2",
+      "version": "10.3.1",
       "commands": [
         "paket"
       ],
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Install build SDK and integration test runtime
         run: |
           curl --fail --silent --show-error --location https://dot.net/v1/dotnet-install.sh --output /tmp/dotnet-install.sh
-          bash /tmp/dotnet-install.sh --version 9.0.309 --install-dir "$HOME/.local/share/dotnetcore"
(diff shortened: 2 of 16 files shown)


dsyme:
Magnificent 🎉

Issue #3411, opened 2018-11-08
Paket hangs forever if using symlinks


incca-konstantin-savchenko:
### Description
I have following folder structure:

	├── Project1
	│    └── ...
	│    └── symlink1
	│    └── symlink2
	│    └── P1.csproj
	│    └── paket.references
	├── Project2
	│    └── ...
	│    └── symlink1
	│    └── symlink2
	│    └── symlink3
	│    └── P2.csproj
	│    └── paket.references
	└──  (...)
	└──  MySolution.sln
	└──  paket.dependencies

Our solution has multiple projects and there are symlinks inside project folders. The symlinks are mapped to local network storage. They are not included in any project and only provide access to shared resources. Some of them have considerable amount of data (20-50GB).

Executing paket (install/update/restore) in this configuration takes eternity.
After:
`.paket/paket restore -v`
Is get: "Parsing (...)\paket.dependencies" and nothing happens for hours. :/

### Known workarounds
After removing all symlinks paket runs as expected and takes only 10-20 seconds.

Is there any possibility to exclude certain paths from processing?
(Paket version 5.186.0)
Thanks

forki:
Oh wow.

Are 5he symlinks only in folders that have a references file in it? If so
we could possibly come up with something in the references file.

Am Do., 8. Nov. 2018, 08:51 hat incca-konstantin-savchenko <
notifications@github.com> geschrieben:

incca-konstantin-savchenko:
Yes, more or less. There are some more symlinks but it seems only those "heavy-weight" linked folders cause the problem.

forki:
Currently I won't be able to do this myself, but I think we would accept a PR that creates a solution to ignore folders and files from paket's csproj / references file search. I think we should create a `.paketignore` file with similar features like `.gitignore`. For the start it's probably enough to just allow ignore patterns.

PR #4421, opened 2026-08-31
Fix infinite recursion in project discovery on symlink cycles

michaelglass:
## What went wrong

`ProjectFile.FindAllProjectFiles` recurses through every subdirectory, following symlinks. A cyclic directory symlink (for example the macOS SDK ncurses links inside a Nix `.devenv` profile) sends that walk round the loop until the path outgrows PATH_MAX, re-finding every project on each lap. `Utils.FindAllFiles` (used by `convert-from-nuget`) has the same shape and the same exposure.

## Fix

Both walks keep following symlinks and key a visited set on a canonical path, so each physical directory is entered once. Discovered files keep the path they were reached by.

Canonicalisation: Paket.Core compiles for net461 and netstandard2.0, and `FileSystemInfo.ResolveLinkTarget` is .NET 6+, so it is bound by reflection once at startup and used when the runtime has it (the `paket` tool on net10.0, and any consumer on .NET 6+). A Unix runtime without it (Mono) falls back to libc `realpath`. Windows on .NET Framework keeps the lexical path, as today.

## Tests

- `sub/loop -> ..` under a temp root: discovery terminates and reports the project once. Without the visited set it reports it 32 times on macOS.
- A symlink to a directory outside the search root: the project is found, under the link path. Pins that links are still followed.

Both pass with the reflection path and with the `realpath` fallback forced.


src/Paket.Core/Common/Utils.fs
```diff
@@ -298,6 +298,47 @@ let isWindows =
...
+module internal NativePath =
+    open System.Runtime.InteropServices
+
+    // POSIX realpath: on Linux/macOS libc resolves `path`, following every symbolic
+    // link, into the caller-provided `resolved` buffer (which must be at least
+    // PATH_MAX bytes), returning that buffer on success or NULL on failure. We pass
+    // UTF-8 bytes so non-ASCII paths round-trip, and a 4096-byte buffer (>= PATH_MAX
+    // on Linux (4096) and macOS (1024)), so no manual free is required.
+    [<DllImport("libc", EntryPoint = "realpath", SetLastError = true)>]
+    extern nativeint realpath(byte[] path, byte[] resolved)
```
dsyme: Is there no way to do this fix using dotnet primitives?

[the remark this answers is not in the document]
michaelglass: Added `[<Out>]` in e42914b54. For the record: `byte[]` is blittable, so the marshaller pins the array and passes its address, and the write was already visible; the attribute documents direction rather than changing behaviour.

tests/Paket.Tests/ProjectFile/SymlinkLoopSpecs.fs
```diff
@@ -0,0 +1,50 @@
...
+
+// Regression test: ProjectFile.FindAllProjectFiles walks a directory tree to find
+// every *proj* file. A self-referential or cyclic directory symlink (e.g. the macOS
+// SDK ncurses symlink loops inside a Nix `.devenv` profile) used to drive that walk
+// into unbounded recursion. The walk now resolves directories to their canonical
+// (symlink-followed) path and prunes ones it has already visited, so it terminates
+// and reports each physical project exactly once.
+[<Test>]
+[<Timeout(120000)>]
+let ``FindAllProjectFiles terminates on symlink cycles and finds each project once`` () =
```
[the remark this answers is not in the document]
michaelglass: Kept the unconditional `[<Timeout>]`. The `NO_UNIT_TIMEOUTATTRIBUTE` gating predates NUnit's netstandard2.0 build: the pinned NUnit 3.12 implements `TimeoutAttribute` on .NET Core (runs the test on a Task and fails it after the deadline). Verified here on net10.0: a probe test sleeping 4 s under `[<Timeout(1000)>]` fails after 1 s, and the Linux CI job ran this test (Passed, 25 ms). Gating it with `Ignore` would skip the regression on the only platform where the scenario exists; Windows already `Assert.Ignore`s it.

Also checked the failure mode with the cycle check disabled: the test fails in ~1 s (project found 32 times; the walk stops once the path exceeds PATH_MAX) rather than hanging, so the timeout is only a safety net.

src/Paket.Core/Common/Utils.fs
michaelglass: Not on Paket.Core's targets: `FileSystemInfo.ResolveLinkTarget`/`LinkTarget` are .NET 6+, and Paket.Core builds net461 + netstandard2.0, so there is no managed way to canonicalise a path there.

What netstandard2.0 does have is the reparse-point attribute, and a cycle can only arise through a link. So 89237a173 drops the P/Invoke entirely and just never recurses into symlinked directories (via the existing `SymlinkUtils.isDirectoryLink`). Net diff to `src` is now +2 lines. The trade is that a project reachable only through a directory symlink is no longer discovered, which matches git and most build tools.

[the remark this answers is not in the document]
michaelglass: Moot as of 89237a173: the P/Invoke is gone, the walk now skips symlinked directories via the reparse-point attribute instead.

michaelglass: hmm. I think dropping symlinks altogether is probably a regression ...

michaelglass: Reworked in 00b174c9c. Skipping links altogether was the wrong trade, so links are followed again and cycles are cut by a visited set keyed on a canonical path. The primitive is `FileSystemInfo.ResolveLinkTarget(returnFinalTarget = true)`, bound by reflection since it is .NET 6+ and Paket.Core compiles for netstandard2.0 (adding a net10.0 target trips 17 unrelated overload-ambiguity errors). Only a runtime without it, i.e. Mono, falls back to libc `realpath`.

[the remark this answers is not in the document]
michaelglass: Back as of 00b174c9c, as the Mono-only fallback, with `[<Out>]`.

michaelglass: ... sorry I've gone back and forth on this and I think it just needs an executive decision. What do you think?

either we detect cycles in symlinks (retain current behavior but fix a bug, implies keeping libc fallback for linux hosts? or dropping net461?)

or we stop following symlinks altogether for package discovery (this changes paket behavior, but also incidentally fixes #3411 (no reliance on modern dotnet features)

Issue #4350, opened 2026-08-25
Replace vendored source files and legacy build dependencies


dsyme:
## Motivation

`paket.dependencies` still contains several unusual dependencies: individual source files and a binary downloaded from GitHub, a private/legacy package feed, a 2017 prerelease documentation package, and a pre-SDK reference-assembly package.

These entries were reasonable compatibility measures when introduced, and some deliberately dogfood Paket's GitHub-file support, but they now make restore depend on old repositories/feeds and make ownership, updates, licensing and security review harder. We should migrate them to maintained packages, standard SDK tooling, or small repository-owned implementations where necessary.

This is an umbrella issue. The entries have different risk levels and should normally be changed in separate PRs.

### Production source dependencies

- `github fsharp/FAKE:0341a2e... src/app/FakeLib/Globbing/Globbing.fs`
  - Compiled directly into `Paket.Core` and used by `NupkgWriter`, `TemplateFile` and `PublicAPI` for `isMatch`/`search` behavior.
  - Added in 2015 for wildcard support in #690.
  - Replace with a maintained globbing package compatible with all Paket targets, or a small repository-owned implementation of the required semantics. Preserve existing wildcard behavior with focused tests before removing the remote source file. The current `Microsoft.Extensions.FileSystemGlobbing` package is a candidate, but compatibility with `net461` and FAKE's exact glob semantics must be verified.

- `github fsprojects/FSharp.TypeProviders.SDK:dc5ac01... src/AssemblyReader.fs`
  - Compiled directly into `Paket.Core`. It supplies assembly references, custom-attribute decoding and SHA1 helpers used by install, package metadata, dependency cache and platform matching code.
  - Replace the metadata-reading portions with `System.Reflection.Metadata`/`PEReader` or another maintained API, and replace the hashing helpers with `System.Security.Cryptography`. This is the highest-risk item and needs regression coverage for binding redirects, assembly metadata and package metadata.

### Test source dependency

- `github forki/FsUnit FsUnit.fs`
  - Compiled directly into both test projects and used throughout the unit and integration suites.
  - It was introduced in 2014 to dogfood Paket's GitHub-file feature, not because a source-file dependency is intrinsically required.
  - Prefer a normal test package if one supports both `net461` and `net9`; otherwise migrate assertions to NUnit or keep a minimal repository-owned compatibility module while the `net461` test target remains. The current FsUnit release targets .NET 6+, so this cannot be a blind package upgrade.

### Documentation dependencies

[...]

Issue #4348, opened 2026-08-25
Investigate removing FAKE from the build


dsyme:
## Motivation

The repository still bootstraps FAKE 4.64.17 to run `build.fsx`, via `build.cmd` and `build.sh`. Most of the build now delegates to the .NET CLI, but starting FAKE still brings in legacy Mono/MSBuild discovery and old build-only dependencies.

This is already causing maintenance friction. In #4002, `GenerateHelp` fails because FAKE eagerly resolves an obsolete `RoslynTools.MSBuild` path even though the docs script does not appear to need it. The Linux build also installs Mono primarily to execute the FAKE 4 runner.

We should decide whether FAKE can be removed from Paket's own build, and agree on a replacement path before making piecemeal changes.

## Current scope

FAKE currently owns:

- target orchestration and conditional `Skip*` parameters in `build.fsx`;
- clean, restore, build, publish, test, ILRepack, checksum, icon and pack steps;
- documentation generation, including a nested FAKE invocation of `docs/tools/generate.fsx`;
- NuGet publishing and the older GitHub/docs release automation;
- the `build.cmd <Target>` / `build.sh <Target>` contributor interface.

There are also independent uses which should be assessed separately:

- `FakeLib.dll` is referenced by the integration-test project;
- a pinned FAKE globbing source file is compiled into `Paket.Core` and used by packaging/template APIs;
- `Fake.Core.ReleaseNotes` is a product dependency used by the CLI;
- many FAKE package names in tests and docs are compatibility fixtures/examples and are not build-system dependencies.

Removing the build runner therefore does not necessarily mean removing every FAKE-related reference in one change.

## Proposed replacement

Prefer standard .NET and repository-native mechanisms:

1. Add a small checked-in .NET build project (F# or C#) for cross-platform orchestration, process execution, release-note parsing, file operations, ILRepack and checksums. Avoid duplicating this logic in bash and batch/PowerShell.
2. Keep thin `build.sh` and `build.cmd` wrappers so existing target-oriented contributor commands remain stable during migration.
3. Move version/package metadata that naturally belongs to the projects into MSBuild props/targets, and call `dotnet restore/build/publish/test/pack` directly.
4. Move CI matrix and artifact orchestration into GitHub Actions. Move releases to a dedicated workflow using GitHub Actions/`gh` and `dotnet nuget push`, rather than interactive credentials and the pinned Octokit script.
[...]

Issue #4346, opened 2026-08-25
Recover and re-enable the full Linux integration test suites


dsyme:
### Context
* PR https://github.com/fsprojects/Paket/pull/4345 restored GitHub Actions CI and exposed integration-suite rot after prerequisite/restore/build failures were fixed.
* Diagnostic run https://github.com/fsprojects/Paket/actions/runs/32868698103 ran with .NET SDK 9.0.309, .NET 8 runtime, Mono, EnableWindowsTargeting, and a temporary 120-minute test timeout.
* **net9 result:** 13 failed, 203 passed, 10 skipped, 226 total, 10m26s.
* **net461 result:** 3 failed, 55 passed, 0 skipped, 58 total, 1h34m.
* The full suites were removed from PR 4345's required matrix so CI establishment could proceed. The required Linux restore/build/publish/unit tests and Windows remain green. However, the integration suites need proper restoration and re-enablement rather than silent abandonment.

---

#### **net9**
1. `#2394 mix dependencies from multiple groups with conditions`
2. `fcs can type check`
3. `run fsi integration tests that have deterministic output`
4. `#2694 paket fixnuspec should not remove project references`
5. `#3317 pack multitarget with p2p`
6. `#3558 pack multitarget with p2p by tfm`
7. `#3983-dont-remove-packed-projects`
8. `#4183 writes ranges for floating deps`
9. `apply framework restriction for force true/false and NuGet v2/v3` (resolved Microsoft.AspNetCore.WebUtilities 10.0.11 while test expects <9)
10. `apply framework restriction for force true/false and NuGet v2/v3` (resolved Microsoft.AspNetCore.WebUtilities 10.0.11 while test expects <9)
11. `apply framework restriction for force true/false and NuGet v2/v3` (resolved Microsoft.AspNetCore.WebUtilities 10.0.11 while test expects <9)
12. `apply framework restriction for force true/false and NuGet v2/v3` (resolved Microsoft.AspNetCore.WebUtilities 10.0.11 while test expects <9)
13. `#2684 Paket should not be called the second time in msbuild`

#### **net461**
* `#1783 generates binding redirect when assembly with different version of main group`
* `#2394 mix dependencies from multiple groups with conditions`
* `#2528 sdk-2: install should not warn with FSharp.Core`

---

### Likely Drift Clusters
[...]

Issue #4183, opened 2022-11-09
dotnet publish multi-targeting inferring incorrect versioning


TheAngryByrd:
### Description

I update [FsToolkit.ErrorHandling ](https://github.com/demystifyfp/FsToolkit.ErrorHandling) from paket 6.2.1 to 7.1.5 (i've tried the alpha and as far down as 7.0) and there seems to be change with how multi-targeting works and having different versions for each.

Previously with `6.2.1` with a project a project with two TFMs `<TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>`

and two groups with different `conditions`

```md
group NetStandard2
source https://api.nuget.org/v3/index.json
nuget FSharp.Core 4.7.2
nuget Ply
nuget Hopac
nuget FSharp.Control.AsyncSeq
nuget Expecto
nuget Expecto.Hopac
nuget Microsoft.NET.Test.Sdk
nuget YoloDev.Expecto.TestSdk
nuget Fable.Mocha 2.12.0
framework: netstandard2.0, net5.0
storage: none
condition: netstandard2_0


group NetStandard2_1
source https://api.nuget.org/v3/index.json
nuget FSharp.Core 6.0.3
nuget Hopac
nuget FSharp.Control.AsyncSeq
nuget Expecto
nuget Expecto.Hopac
nuget Microsoft.NET.Test.Sdk
nuget YoloDev.Expecto.TestSdk
nuget Fable.Mocha 2.12.0
nuget IcedTasks
storage: none
condition: netstandard2_1
```

I would get a nuspec file like:

```xml
    <dependencies>
      <group targetFramework=".NETStandard2.0">
        <dependency id="FSharp.Core" version="4.7.2" exclude="Build,Analyzers" />
      </group>
      <group targetFramework=".NETStandard2.1">
        <dependency id="FSharp.Core" version="6.0.3" exclude="Build,Analyzers" />
      </group>
    </dependencies>
```

However after updating to 7.1.5 my nuspec file looks like:

```md
group NetStandard2
source https://api.nuget.org/v3/index.json
nuget FSharp.Core >= 4.7.2
nuget Ply
nuget Hopac
nuget FSharp.Control.AsyncSeq
nuget Expecto
nuget Expecto.Hopac
nuget Microsoft.NET.Test.Sdk
nuget YoloDev.Expecto.TestSdk
nuget Fable.Mocha
nuget Fable.Core >= 4.0.0-theta-006
[...]

PR #4423, opened 2026-09-01
build: write build.fsx against FAKE 6 and merge with dotnet-ilrepack

fpellet:
Part of #4348 — the "upgrade to fake 6" step of the plan, after #4422 removed the FAKE runner from the entry points.

## What changes

**FAKE 6 as a library.** `build.fsx` was written against `FakeLib.dll`, i.e. FAKE 4.64.17 — a .NETFramework 4.5 assembly from 2017 that only happens to load under .NET 10. It is now written against the FAKE 6 modules. Those target net6.0/netstandard2.0, so they cannot live in the `Build` group, which is pinned to `framework >= net461` for FSharp.Formatting and the reference assemblies; they get a new `BuildScript` group restricted to net10.0, the framework `dotnet fsi` runs on. `build.sh` / `build.cmd` generate its load scripts after the restore, so running `dotnet fsi build.fsx <Target>` directly assumes one of the entry points has run at least once.

**`dotnet-ilrepack` instead of `ILRepack.exe`.** `ILRepack.exe` is a .NET Framework binary, so on Linux FAKE ran it through Mono. The `dotnet-ilrepack` tool ships the same repacker as a .NET tool. Mono also supplied the .NET Framework reference assemblies implicitly; under `dotnet` they have to be passed explicitly, so the repacker is pointed at the same directory MSBuild already resolves net461 against (`TargetFrameworkRootPath` from `Directory.Build.props`).

`MergePaketTool` therefore no longer needs Mono. It is still needed for the `net461` test passes and for `PublishNuGet`, which pushes with the merged `net461` `paket.exe`.

## Behaviour differences worth reviewing

- **`isMono` guards became `Environment.isWindows`.** FAKE 4 derived `isMono` from `Environment.OSVersion.Platform`, so it was true on Linux and `not isMono` really meant "on Windows". FAKE 6 tests for the Mono runtime, which is false under .NET 10 everywhere, so a literal translation would have silently enabled those targets on Linux.

- **`DotNet.test` defaults to Debug**, unlike build/publish/pack. With `--no-build` that would look for `bin/Debug`, so the configuration is now explicit on every test invocation.

- **`BuildServer.isLocalBuild` changes meaning on CI.** FAKE 4 reported `isLocalBuild = true` even on GitHub Actions; FAKE 6 detects the build server and reports `false`. The doc targets are guarded by it, so `GenerateDocs` / `GenerateReferenceDocs` / `GenerateHelp` now no longer run on the Windows runner. That looks desirable — `docs/tools/generate.fsx` is still bound to FSharp.Formatting 3 and its Razor templates, and the docs targets still shell out to the FAKE 4 runner for it — but it is a side effect of the library swap rather than a deliberate decision, so say the word if it should be made explicit instead.

- **`github fsharp/FAKE modules/Octokit/Octokit.fsx` was re-resolved** in `paket.lock` (`13eee5a7…` → `e8e1cae7…`) as a side effect of the restore; it has no SHA in `paket.dependencies`. `build.fsx` `#load`s it unconditionally, so it affects every target, not just `ReleaseGitHub`. The new revision still exposes the `createClient` / `createDraft` / `uploadFile` API the script uses, so nothing is broken — but it can be reverted or pinned the way `enricosada/add_icon_to_exe` already is, if preferred. #4350 proposes replacing it outright.

- **The SDK install directory is pinned** to `LocalApplicationData/dotnetcore` through `CustomInstallDir`. FAKE 6's `DotNet.install` defaults to `~/.dotnet` and can also settle on a `dotnet` found through `PATH`; the CI installs the SDK and the .NET 8 runtime the integration tests need into the FAKE 4 location.

## Fixes found along the way

Targets that were silently doing the wrong thing, each in its own commit:

- `QuickTest` built the test assembly without the version property, so `Loading assembly metadata works` compared `1.0.0` against `RELEASE_NOTES.md` and failed. Only visible locally — CI runs `BuildPackage`.
- The netcore unit-test pass ran on the wrong framework, and the bootstrapper unit tests were not run at all.
- Empty `Skip*` variables now count as unset, as they did under FAKE 4.
- `PublishNuGet` resolves the merged `paket.exe` absolutely, since `ToolType.CreateFullFramework()` only prefixes Mono for `.exe` paths.
- Without a runner to parse it, the command line is translated by the script itself; extra target arguments and empty parameter names are now rejected instead of silently changing which target runs.

## Verification

`dotnet fsi build.fsx <Target>` type-checks and runs; the generated `dotnet build/test/pack/publish/restore/exec` command lines were compared against the FAKE 4 originals, including `--filter`, `--logger`, `--no-build`, `--output` and the `PackageReleaseNotesFile` quoting.

## Not in scope

The remaining steps of #4348 — replacing the doc generator, dropping the last Mono uses, and moving further onto standard commands — are left for follow-ups.


fpellet:
Previous build on fake 4 crash on GenerateHelp, and not execute:
  - GenerateReferenceDocs
  - GenerateDocs
  - All
  - MergePaketTool
  - AddIconToExe
  - RunIntegrationTestsNet
  - RunIntegrationTestsNetCore
  - SignAssemblies
  - CalculateDownloadHash
  - NuGet
  - BuildPackage

fpellet:
## Update — the Windows job now runs the integration suite, for the first time

Green on `c6bb869e9`. The extra commits since the description are all consequences of one finding, so here is the chain.

### The Windows job used to pass by accident

On `master`, run [33528762012](https://github.com/fsprojects/Paket/actions/runs/33528762012), job `99926228529`: GitHub reports **success**, while the log reports `Status: Failure` — `GenerateHelp` fails with `System.Exception: generating documentation failed` (#4002). `build.cmd` did not propagate `dotnet fsi`'s exit code, so everything after `GenerateHelp` — `MergePaketTool`, `AddIconToExe`, the integration tests, `NuGet`, `BuildPackage` — had never actually run on Windows.

That settles the `BuildServer.isLocalBuild` point raised in the description: on this branch the doc targets are skipped, the build gets past that point, and `RunIntegrationTestsNet` executes. It is worth knowing that the full suite runs nowhere else either — the Linux `build` job passes `SkipIntegrationTests=true`, and the `flaky` job filters to `TestCategory=Flaky`, i.e. a single integration test. **This branch is the first time the integration suite has run in CI at all**, which is why it turned up so much.

### Two real bugs it uncovered

**Missing `System.Buffers` binding redirect** (`99a265b7c`, `89b84776d`). `System.Net.Http.WinHttpHandler` references `System.Buffers 4.0.3.0`, but `net461` resolves the `netstandard2.0` asset of `System.Buffers 4.6.1`, which is stamped `4.0.2.0`. MSBuild only unifies references upwards, so `AutoGenerateBindingRedirects` reports MSB3277 and emits nothing at all — and every HTTP response died in `WinHttpResponseParser` with a `FileLoadException`. The redirect is added to `src/Paket/App.config` and to both `net461` test assemblies: tests that drive Paket.Core in-process bind against `<testassembly>.dll.config`, not `paket.exe.config`. Only `bin/net461/paket.exe` was affected; the shipped `bin/merged/paket.exe` repacks both assemblies and never resolves the reference, so there is no release note.

**The suite could hang forever** (`9b50f5137`). `ExecProcessWithLambdas` (FakeLib 4.64.17) ends with an unbounded `WaitForExit()` *after* its timeout logic; it returns only once the redirected streams reach EOF. With node reuse on, `dotnet build/run/pack` leaves MSBuild worker nodes behind holding the inherited stdout/stderr handles, so EOF never comes — which is why `PROCESS TIMED OUT` never appeared in the log despite the 7-minute setting. Reproducible on Linux too. Setting `MSBUILDDISABLENODEREUSE=1` in `TestHelper` took `ConditionSpecs` from an infinite hang to 2m28s.

Separately, `7bf9a000e`: `i001195-broken-appconfig` had stopped being broken. Its `app.config` deliberately contained a `<<<<<<< HEAD` marker to make it invalid XML, and `dd0b881e6` stripped it as if it were merge residue — so paket had nothing to report and the test failed with "paket should fail". Restored. It will probably be stripped again by the same tooling; happy to make the fixture malformed in a less conflict-marker-shaped way if you prefer.

### What I parked, and why (`3f35b7b29`, `c6bb869e9`)

The first complete run finished 226 tests with 12 failures. None of them comes from this PR — they are the debt of a suite that had never run. Rather than grow this PR into an open-ended investigation, each is marked with what it actually does and tracked by #4346:

| test(s) | marked | reason |
|---|---|---|
| `apply framework restriction` (×4) | `Ignore` | framework restriction not applied to the resolution |
| `#2694`, `#3558`, `#3983` | `Ignore` | the expected `.nupkg` is never produced |
| `#3317` | `Ignore` | `dotnet pack` exits 1 |
| `#4183` | `Ignore` | `Many groups detected map []`, floating deps dropped |
| `#2684` | `Ignore` | second restore still reports mismatching hashes |
| `run fsi …deterministic output` | `Ignore` | 2 of 8 scripts no longer produce their expected output |
| `#3014`, `#1635` | `Flaky` | network-dependent, see below |

One deserves a closer look before merge: **`apply framework restriction` is not a stale bound.** It resolves `Microsoft.AspNetCore.WebUtilities 10.0.11` under `framework: =net8.0`, although that version ships only `lib/net10.0` and `8.0.21` is the newest one that fits. Raising the assertion to `< 11` would have disarmed a test that is catching a genuine resolution bug, so it is ignored with that diagnosis written above it instead.

`#3014` and `#1635` are network flakes, not failures: they hit `api.github.com` (60 unauthenticated requests an hour per IP, shared across hosted runners) and `myget.org`, they passed in one Windows run and failed in the next with both endpoints healthy, and they pass locally. `[<Flaky>]` is the category this repo already keeps for that — with the consequence that the Linux flaky job now inherits them.

### Result

Windows on `c6bb869e9`: `Status: Ok`. net461 integration 224 discovered / **211 passed** / 13 skipped in 16 minutes; net10.0 227 / 207 / 20. Unit tests unchanged.

And the green now means something: with the pwsh default the job reported success over `Status: Failure`, whereas with `shell: cmd` (`0a26cdf3d`, added to close stdin) the failing run correctly reported `Process completed with exit code 1`. Supporting changes: normal verbosity on the integration run so that a hang names the test that never returned (`a721b1346`), and the timeout cut from 60 to 30 minutes now that the suite takes ~16 (`be223af75`).

### Still open for you to call

- The `Octokit.fsx` re-resolution noted in the description is unchanged — leave it, revert it, or pin it the way `enricosada/add_icon_to_exe` is.
- Skipping the doc targets on CI is currently a side effect of `isLocalBuild` rather than a deliberate guard. Given that `GenerateHelp` fails on the Windows runner anyway (#4002), skipping looks right, but say the word if you would rather it were explicit.

